PR #54, opened 2024-09-07
Add src-rust-lib and clean src-tauri

maidh91:
# Summary
- Clean code only


maidh91:
hi @afadil 

- this is also another PR to refactor the code only, no business logic change
- `src-rust-lib` contains all business logic of the backend that can be reused
- `src-tauri` contains tauri commands only as a wrapper calling to the business code in `src-rust-lib`

this helps the code easy to extend/switch/migrate stack later, 
e.g. if we need to try another desktop framework rather than Tauri, we can easily add another thing like `src-electron` calling to `src-rust-lib`

PR #55, opened 2024-09-07
Update import icon

alexanderameye:
Change import icon to signify an import/download instead of an upload/export.

Also change height of import button so it matches the height of the button next to it.

**before**
image

**after**
image


alexanderameye:
Combined minor visual fixes in

https://github.com/afadil/wealthfolio/pull/58

PR #57, opened 2024-09-07
Fix/nan income sources if income is empty

alexanderameye:
When totalIncome is 0, then the percentages show up as NaN. This sets the percentages to 0 if they could not be calculated.

before
image

after
image

Could alternatively do something like this, I see this happens in other places in the code base.
image


alexanderameye:
Put changes in https://github.com/afadil/wealthfolio/pull/58

PR #58, opened 2024-09-07
Minor visual fixes

alexanderameye:
1. do not allow negative values for share/price/fee input fields
image

2. add missing border for ticker search results

before
image

after
image

3. Change import icon to signify an import/download instead of an upload/export.

Also change height of import button so it matches the height of the button next to it.

**before**
image

**after**
image

4. When totalIncome is 0, then the percentages show up as NaN. This sets the percentages to 0 if they could not be calculated.

before
image

after
image

Could alternatively do something like this, I see this happens in other places in the code base.
image


afadil:
Merged. Thank you @alexanderameye !

PR #66, opened 2024-09-07
fix activity form

afadil:


alexanderameye:
I'm not very familiar with zod for validation, but currently you can still enter negative numbers here (the form validation blocks it though). Wouldn't it be more user friendly to also not just allow negative values to be in the input field if they were going to be blocked anyways?

image

PR #79, opened 2024-09-09
feat: make currencies selection popover scrollable

CanglongCl:
Without this we are not able to select currency easily.


afadil:
Thank you @CanglongCl

PR #94, opened 2024-09-10
Clean diesel

maidh91:
- Remove duplicate diesel.toml
- Use src-core/diesel.toml instead of the one at root


afadil:
diesel.toml removed in another commit

PR #105, opened 2024-09-13
Setup adapters for commands

maidh91:
# Summary
- Previously, in frontend code, we have centralized all tauri commands into a single dir at `src/commands`. Now we continue to improve this to centralize all tauri commands into a single file at `src/adapters/tauri.ts`
- We also introduce the `RUN_ENV` enum, and `getRunEnv` helper to determine the running env, e.g. desktop, browser, mobile
- This helps cleaning the code with these things
  - preparing for new running env
  - use generic type from `invoke` function instead of manual type casting

### DIFF
--- a/src/adapters/index.ts
+++ b/src/adapters/index.ts
@@ -5,7 +5,7 @@ export enum RUN_ENV {
   UNSUPPORTED = 'unsupported',
 };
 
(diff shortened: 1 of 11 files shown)


afadil:
Merged! Thank you @maidh91. This is coming up nicely.

PR #112, opened 2024-09-16
Added Dockerfile and docker-compose for web-only debugging/environment

AnushthanS:
Added a Dockerfile and docker-compose. These are useful if anyone wants to run the web-app side of the application without installing rust, tauri or pnpm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22
+FROM node:latest
 
 RUN npm i -g pnpm
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM node:22
+
+RUN npm i -g pnpm
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+ENV PATH="/root/.cargo/bin:$PATH"
+
+WORKDIR /app
+
+COPY . .
+
+RUN pnpm install
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:22
```
maidh91: node:latest pls

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN npm i -g pnpm
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
```
maidh91, line 5: should the command be `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`

https://www.rust-lang.org/tools/install

AnushthanS: That command initiated a cli based interactive installer. I tried a bunch of flags and ended up with this as the simplest method.

Dockerfile
colemvnio: I would tend to favor the fixed version as well to promote predictability, backwards compatibility and security. Using the latest tag would promote builds to non-lts versions which could introduce unwanted/unknown issues.

When updating versions, then the Dockerfile should simply be updated as well.

leomeneguzzi:
up!
This will be great, I will try to run this on my self host!

PR #125, opened 2024-09-28
Fix average monthly income calculation

jeromecornet:
The number of months for monthly average income calculation only takes into account the months where there is an actual income. So if there is a month with no income (for example last year), the 'monthly average' gets artificially bumped up.

This simplifies the calculation to use the actual number of months in the given year (current month number for the current year, 12 for previous years) rather than only the months where there was some income

### DIFF
--- a/src-core/src/models.rs
+++ b/src-core/src/models.rs
@@ -1,6 +1,7 @@
 use diesel::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+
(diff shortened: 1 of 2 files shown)


jeromecornet:
To fix the overall calculation as well as to take into account history starting less than 2 years ago, I ended up adding a query to figure out the date of the oldest transactions so that the correct number of months is taken into account (i.e. less than 12 for previous year/2 years ago if the history doesn't go that far)

afadil:
Merci @jeromecornet

PR #128, opened 2024-09-29
Improve currency inputs

justinwaltrip:
Improve support for currency inputs, such as prefixes (ex. "$12.34") and group separators (ex. "1,234"). This should make it easier to paste currency values into the application when copied from a variety of sources.

Ex. If user pastes string "$1,234.56", then the input field should display value "1234.56"


afadil:
Thanks @justinwaltrip !

PR #144, opened 2024-10-17
WIP Import Profiles

Yiroh:
Managed to get the software to compile and run, can't test it since I have no CSVs to test with. Submitting pull request for that purpose.


src-core/wealthfolio.db
afadil, line 1: remove this db.file

src-core/migrations/2024-10-15-173026_create_csv_import_profiles/up.sql
afadil, line 1: I suggest we use only one table with mapping as map stored in text columns. The app will read and parse the mapping from each column. Something like this:

CREATE TABLE IF NOT EXISTS import_mappings (
    account_id TEXT PRIMARY KEY,
    fields_mappings TEXT NOT NULL,
    activity_type_mappings TEXT NOT NULL,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);

src-core/src/db_functions.rs
afadil, line 1: move this code to a account_service. with only 2 functions, get_account_import_mapping(account_id) and save_account_import_mapping(account_id, import_mapping)

src-core/src/models.rs
afadil, line 1: I suggest we keep the model simple:
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct ImportMapping {
    pub account_id: String,
    pub fields_mappings: HashMap<String, String>,
    pub activity_type_mappings: HashMap<String, String>,
    pub created_at: DateTime<Utc>,
    pub updated_at: DateTime<Utc>,
}

src/commands/activity.ts
afadil, line 1: can you please remove all the frontend code, because I completely changed the UX

afadil (changes requested):
Hi,
I left you some comment so the PR can match the new UX for import I'm working on. If you can please make the adjustments.

Thank you so much.

Yiroh:
Made the changes you suggested and simplified an area where it felt more complex than it needed to be (my creation of csv_profile.rs). I wrote an excerpt about it in the paragraph of my commit changes.

PR #153, opened 2024-10-28
support for private assets

geordie:
Add support for private assets with the following changes:
* Update the Add Activity UX to have a checkbox that when unchecked enables entry of manual symbols that don't get looked up by a public API
* Enable addition of activities with symbols that are not public
* Add the ability to sync quotes for non-public assets by using the quoted values input by the user when they create activities


afadil:
Hi, thanks for the contribution. 
Can you please check why the custom assets do not impact the whole portfolio value displayed on the home page.

geordie:
Yes. Will take a look.

geordie:
I fixed what looked like a pre-existing presentation layer issue on the home page with account summaries. Otherwise, have page value calculation seems to work for me.

If you're still seeing issues could you provide a quick set of steps to repro and I can resolve?

afadil:
Some feedback and issues noticed after testing:
- Private Asset Source: Private assets shouldn’t be pulling data from Yahoo. Try using XXXX as the source instead.
- Value Calculation: The current value calculation for private assets looks off. When I add multiple BUY or Transfer In activities for the same asset, the total value isn’t adding up right. Might need a look at how those values are being combined.
- Edit Activity Checkbox: When editing activities for private assets, the Use Symbol Lookup checkbox should default to unchecked. Right now, it’s auto-checked, which feels unnecessary.
Also, I think we need to automatically insert a default quote for the symbol and allow the user to manually update it in the symbol page.

geordie:
Last commit resolves the Edit Activity Checkbox issue.

Looking at other issues.

geordie:
> * Private Asset Source: Private assets shouldn’t be pulling data from Yahoo. Try using XXXX as the source instead.

For this, the thinking is that when creating a new activity, unchecking "Use Symbol Lookup" is not the same as saying that the asset is private. Only that the user does not want to lookup the symbol in the form. So when the user submits the form, the activity_service first tries to find the symbol via the public service in case it is a public symbol. If no symbol is found, then the activity service falls back to creating a private asset.

When the user updates an activity for a private asset, no lookup will be done as long as the user does not check the "Use Symbol Lookup" option and type, or manually change the symbol. This is the case because on submit, the activity service first tries to look the asset up in the database. If found, no lookup is done and no quote lookups happen for private assets.

Does that cover your concern or do you have different behaviour in mind?

geordie:
> * Also, I think we need to automatically insert a default quote for the symbol and allow the user to manually update it in the symbol page.

For this, can you say more about what you mean? Currently, when a private asset is added, quotes also get added for each day from the activity day to today.

There is an issue if the user specifies zero amounts for the price or number of shares when adding the activity. The issue is that this causes the Asset Profile page not to load because all holdings with a 0 shares get filtered out of the returned holding set in this line of code:

https://github.com/afadil/wealthfolio/blob/eb4886f219c64e3f45c2f8a8a4217b3c3c369f2b/src-core/src/portfolio/holdings_service.rs#L415

Then when the asset profile page tries to load it gives up because there is no holding. That happens here:

https://github.com/afadil/wealthfolio/blob/eb4886f219c64e3f45c2f8a8a4217b3c3c369f2b/src/pages/asset/asset-profile-page.tsx#L87

A few ways to remedy this:

1. Remove the threshold filter in the holding service and show the asset with zero shares
2. Show a more graceful no shares of this holding message on the asset filter page
3. Set defaults for number of shares and values when adding and updating activities that prevent zero values

3 is probably a poor option because you want to be able represent that you bought a thing then sold all of that thing. This is arguably a bug right now that if you buy 100 shares of an asset, then sell 100 shares of an asset, then try to link to that asset from the activity list, you get a blank asset page. You would rather see the history of that asset I think.

That suggests 1 is a good option. If you agree let me know and I can remove the filter.

2 might also be a good idea but not specific to private assets.

Or maybe I'm missing your original point :)

afadil:
Hi

> > * Private Asset Source: Private assets shouldn’t be pulling data from Yahoo. Try using XXXX as the source instead.
> 
> For this, the thinking is that when creating a new activity, unchecking "Use Symbol Lookup" is not the same as saying that the asset is private. Only that the user does not want to lookup the symbol in the form. So when the user submits the form, the activity_service first tries to find the symbol via the public service in case it is a public symbol. If no symbol is found, then the activity service falls back to creating a private asset.
> 
> When the user updates an activity for a private asset, no lookup will be done as long as the user does not check the "Use Symbol Lookup" option and type, or manually change the symbol. This is the case because on submit, the activity service first tries to look the asset up in the database. If found, no lookup is done and no quote lookups happen for private assets.
> 
> Does that cover your concern or do you have different behaviour in mind?

I think it's less confusing if "Use Symbol Lookup" is applied everywhere and not only in the form. I don't see a use case where the user want to not lookup the symbol only in form and let the app do it in the backend.

I think we should rename Use Symbol Lookup to "Custom Symbol" to reflect this general Behaviour.

What do you think?

afadil:
> > * Also, I think we need to automatically insert a default quote for the symbol and allow the user to manually update it in the symbol page.
> 
> For this, can you say more about what you mean? Currently, when a private asset is added, quotes also get added for each day from the activity day to today.
> 
> There is an issue if the user specifies zero amounts for the price or number of shares when adding the activity. The issue is that this causes the Asset Profile page not to load because all holdings with a 0 shares get filtered out of the returned holding set in this line of code:
> 
> https://github.com/afadil/wealthfolio/blob/eb4886f219c64e3f45c2f8a8a4217b3c3c369f2b/src-core/src/portfolio/holdings_service.rs#L415
> 
> Then when the asset profile page tries to load it gives up because there is no holding. That happens here:
> 
> https://github.com/afadil/wealthfolio/blob/eb4886f219c64e3f45c2f8a8a4217b3c3c369f2b/src/pages/asset/asset-profile-page.tsx#L87
> 
> A few ways to remedy this:
> 
> 1. Remove the threshold filter in the holding service and show the asset with zero shares
> 2. Show a more graceful no shares of this holding message on the asset filter page
> 3. Set defaults for number of shares and values when adding and updating activities that prevent zero values
> 
> 3 is probably a poor option because you want to be able represent that you bought a thing then sold all of that thing. This is arguably a bug right now that if you buy 100 shares of an asset, then sell 100 shares of an asset, then try to link to that asset from the activity list, you get a blank asset page. You would rather see the history of that asset I think.
> 
> That suggests 1 is a good option. If you agree let me know and I can remove the filter.
> 
> 2 might also be a good idea but not specific to private assets.
> 
> Or maybe I'm missing your original point :)

For this, Sorry, I didn't see that quotes are added when I create the activity. This explains the data I have because I was entering random values for the price.
In this case we need to update the quotes on activity update and delete. 

For the other point, You can't log an activity with 0 quantity and 0 price so I dont see the issue.

geordie:
> I think it's less confusing if "Use Symbol Lookup" is applied everywhere and not only in the form. I don't see a use case where the user want to not lookup the symbol only in form and let the app do it in the backend.
> 
> I think we should rename Use Symbol Lookup to "Custom Symbol" to reflect this general Behaviour.
> 
> What do you think?

The reason it may make sense to only have the "Use Symbol Lookup" work in the form is that it helps prevent ambiguous symbols. So for example, let's say someone unchecked the checkbox in the form, whatever it's called, and then input "AMZN." If the app then assumes the symbol is private and therefore doesn't use the public symbol lookup in the backend, then the user now has the potential for ambiguous assets. They would create a private asset ID'd by AMZN that conflicts with a public asset ID'd by AMZN that represents Amazon public stock - so something other than the private asset.

You could then see this ambiguity being a problem for things like holdings aggregation, portfolio value calculations, and when a user tries to import CSVs with records ID's by a symbol that has multiple assets. In this last case, app logic to determine if the CSV records are meant accrue the updates to the private or public asset would be tough to figure out.

By always trying to lookup symbols in the public domain first, the app can prevent, or at least greatly reduce the chances of these ambiguously ID's assets. This is achieved by always attempting to see if input symbols are in the public domain first, and only allowing private asset IDs that are not in the public domain.

geordie:
> For the other point, You can't log an activity with 0 quantity and 0 price so I dont see the issue.

I'm able to add 0 shares of for $0. Steps:

* On the Activity Page, click on "Add Manually" to launch the Add Activity Form
* Select an account
* Choose BUY as the Activity Type
* Leave today's date
* Choose a public symbol (ex. AMZN)
* Leave shares, price and fee with the defaults of 0
* Click the Add Activity  button

Result: The Activity now shows up in the Activity list. This seems fine.

If you then go to the Holdings Page and select the Holdings tab, the newly added holding with zero value does not show up in the list.

I would expect it to show up.

geordie:
> In this case we need to update the quotes on activity update and delete.

I believe update_activity does synch quotes. Happens by virtue of the last param in this call being set to true:

https://github.com/afadil/wealthfolio/blob/eb4886f219c64e3f45c2f8a8a4217b3c3c369f2b/src-core/src/activity/activity_service.rs#L135C1-L136C69

I don't understand how deleting activities should update quotes. Can you elaborate on what you mean?

afadil:
All right. I read your comments and I think it better to keep "Use Symbol Lookup".
Going to add other assets type in the future (Real Estate, commodities...) so Use Symbol Lookup make sense for this use case.  
It seems the PR is ready to merge. Do you have any other upcoming commits, or can I merge now?
Thank you for your help with this one.

geordie:
I'm good to merge.

PR #203, opened 2025-01-14
fix: asset-profile-page overwriting formData from displayed data.

Zuricos:
I have discovered a bug in the "asset-profile-page" component. 
Currently, the displayed data in the input fields of sector:weight and country:weight are used when new entries are added.
I.e. this will cause that the weigth will be 18 instead of 0.179 after adding a new entry.

With this change, this should no longer happen.

### DIFF
--- a/src/pages/asset/asset-profile-page.tsx
+++ b/src/pages/asset/asset-profile-page.tsx
@@ -193,13 +193,30 @@ export const AssetProfilePage = () => {
               )}
               placeholder="sector:weight"
               onChange={(values) =>
-                setFormData((prev) => ({
-                  ...prev,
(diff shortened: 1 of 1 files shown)


afadil:
Hi @Zuricos, thank you for this contribution. I Think that with the proposed update, it's no more possible to delete a sector.

Zuricos:
Hi @afadil , thanks for your feedback, I will look into it, probably this week, and will come back with a fix.

Zuricos:
Hi @afadil,

This issue should now be resolved. However, I am unable to verify it on my end since I couldn't get the project to run locally, even when using the dev container.

STUFFIN5454 (approved):

PR #209, opened 2025-01-27
Simplify private asset profile management

geordie:
Private asset building previously used a full MarketDataProvider when most of the methods on that trait are not applicable to private assets (ex. get_stock_history, get_latest_quote). To simplify the code, added an AssetProfiler trait that simply gets the profile for an asset but does not have all of the MarketDataService methods that are necessary for public assets.

Also renamed get_symbol_profile for the MarketDataService to get_asset_profile to reflect that some assets are equities with public symbols.

Also removed a react component related to manual (private) assets that is no longer used.


afadil:
Merged. Thank you @geordie

PR #212, opened 2025-01-31
Issue fix: private asset portfolio history

geordie:
Issue: After a creating a private asset and adding a quote, the portfolio history value can be incorrect. This is caused by the history value calculation which assumes there is a quote at least every 30 days. Private assets often won't have quotes at least every 30 days.

Steps to recreate

* Create a cash account
* Add a buy activity on that account for a private (manual) asset with the activity date set to one year ago. As an example, set the asset value to $900K
* Add a quote for 11 months ago with a value greater than the orginal value (ex. $1000K).
* Recalculate the portfolio history
EXPECTED: The portfolio value for the account will be $1000K and the total gain will be $100K for every day subsequent to the added quote.
ACTUAL: The portfolio value shows value of 0 for the from one year ago to 11 months ago, then the expected value for 30 days, then a value of 0 again after the 30 days of correct values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ out
 # Testing
 coverage
 .nyc_output
+src-core/tests/output/
 
 # Environment files
 .env
--- a/src-core/Cargo.lock
+++ b/src-core/Cargo.lock
@@ -54,6 +54,28 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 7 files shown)


geordie:
To run the included integration test, from the repo root dir:

```
> cd src-core
> cargo test
```

PR #217, opened 2025-02-24
Make column visibility toggling on activity-table.tsx more intuitive

geordie:
Currently, the Columns drop down menu on activity-table is not intuitive. Columns can't be shown/hidden by clicking on column items in the drop down, and there are columns shown in the drop down that aren't displayed or displayable.

This fix adjusts column state and initialState of the component so that only columns that can be shown/hidden are displayed in the dropdown and that when these are checked, the columns get displayed or hidden.

I decided it made sense to be able to show/hide the Price, Fee, Value, and Account columns. This choice can be easily modified by changing the enableHiding value for each respective columns.


afadil:
Thank you @geordie

PR #228, opened 2025-04-16
Minor UI improvements to holdings page

SvenHockers:
## Minor Improvements to holdings page:

1. The "By Sector" bar chart now uses the entire vertical space allowing for more sectors to be displayed
2. Modified the `CustomTooltip` so that the value is displayed in the users base currency
3. Made adjustments to `custom-pie-chart.tsx` enabeling sizing directly when invoking the `CustomPieChart`
4. Made adjustments to  `ustom-pie-chart.tsx` which ensures the value is displayed in the users base currency.

With the adjusted pie chart implementation I resized the pie charts on the holdings page to fit a bit better into the cards


afadil:
the charts design and behavior changed in v1.1. Not sure this PR still relevant?

PR #265, opened 2025-06-03
Fix duplicate portfolio deposits

afadil:
## Summary
- prevent reapplying past activities when snapshots exist by starting from the next day

## Testing
- `npm test`

------
https://chatgpt.com/codex/tasks/task_e_683f5107d4e88324944e4728151e2388


eldossjogy:
Hey,

I looked into this issue further after posting about it. I'm not very experienced with Rust, but I came up with a similar workaround.

I also noticed that updating an activity causes it to duplicate. I believe this might be due to the snapshot and daily valuation not updating properly.

If you'd like you can take a look at some of the stuff I was working on in my [fork](https://github.com/eldossjogy/wealthfolio/tree/test)

Issue #252, opened 2025-05-21
Pre-market / after-hours total on main page is not accurate


jcheong-1:
This appears to only happen pre-market/after-hours outside of regular trading hours. The total on the main page is not accurate, while the total on the holdings page is accurate.

Is it possible for the totals on the main page to grab the total balance from the holdings page?

Thanks.

jeromecornet:
I believe I've got a fix in https://github.com/afadil/wealthfolio/pull/270.

Are you seeing "No quotes for date {} (account '{}'). Skipping day." in the logs ?

jcheong-1:
No I think this issue is different.

I'm in Canada on the PST time zone. If I check after hours or on the weekend when the stock market is not open, the total value on the main account page upon launch the app, is often different than the total value on the performance / holdings page, which display the correct value.

I believe the total on the main page should derive the total from the holdings page and thus they should be the same all the time.

jeromecornet:
So I didn't write any of this but from reading the code here is my understanding.

The computation of the valuation on the dashboard is done on a daily basis, and the date calculation is timezone dependent. I suspect that what is after market for you in PST is actually the next day in UTC.

In the holdings tab, the computation is based on the last available quote for each symbol. So if there is no quote for a given day, the valuation doesn't match the holdings screen (because the asset is valued at 0). So my PR pretends that there is always a quote for today that matches the last known quote when computing the valuation, ensuring that both screens line up.


In any case if you can test the code in the linked PR, it might fix the issue for you. If it doesn't then there may be another issue.

jcheong-1:
Hmm, thanks for the information, unfortunately I'm not tech savvy enough to test the code in the PR. I'll just wait for the next release and see.

About the total on the dashboard page, when I launch up the app every few hours during market hours during the day the total does update in real time and are always the same as the holdings tab.

I only observe the discrepancy outside regular trading hours, therefore the holdings tab total is the one I use for an accurate total.

PR #270, opened 2025-06-09
Use the last known quote for a symbol that may not have a quote for that day

jeromecornet:
Fixes https://github.com/afadil/wealthfolio/issues/252 (I think)

I believe that I am hitting the same issue as https://github.com/afadil/wealthfolio/issues/252 . In my case it's a mutual fund for which the quotes are only updated once a day after trading hours, and the dashboard values it at $0 instead of the last known value.

I'm seeing this called in the logs: https://github.com/afadil/wealthfolio/blob/be21a6b1fca377b19e057f6796d5d1abfc183580/src-core/src/portfolio/valuation/valuation_service.rs#L263-L266 

So I tweaked the valuation service to fallback to the last known quote instead of returning None. 
I know very little about rust, so I tried to make the code change legible, but feel free to change any of it.

It's not efficient at all, because if the same symbol is held in multiple accounts, it goes through the same process for each account. But I figured a slow correct result is better than a bad result.

### DIFF
--- a/src-core/src/portfolio/valuation/valuation_service.rs
+++ b/src-core/src/portfolio/valuation/valuation_service.rs
@@ -158,6 +158,45 @@ impl ValuationService {
         }
         quotes_by_date
     }
+
+    /// Fills in missing quotes for the current date by using the last known quote for each symbol
+    fn fill_missing_quotes_with_last_known(
+        &self,
+        quotes_by_date: &HashMap<NaiveDate, HashMap<String, Quote>>,
+        current_date: NaiveDate,
+        required_symbols: &HashSet<String>,
+        start_date: NaiveDate,
+    ) -> HashMap<String, Quote> {
+        let mut quotes_for_current_date = quotes_by_date
+            .get(&current_date)
+            .cloned()
+            .unwrap_or_default();
+
+        // For each required symbol that's missing a quote on the current date,
+        // look backward to find the most recent quote
+        for symbol in required_symbols {
+            if !quotes_for_current_date.contains_key(symbol) {
+                // Look backward from current_date to start_date to find the last known quote
+                let mut search_date = current_date - Duration::days(1);
+                while search_date >= start_date {
+                    if let Some(daily_quotes) = quotes_by_date.get(&search_date) {
+                        if let Some(quote) = daily_quotes.get(symbol) {
+                            // Copy the quote for search_date into current_date
+                            warn!(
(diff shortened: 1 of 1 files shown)


jeromecornet:
OK, I rebased the PR after #271 was merged. It still fixes the issue I have

jeromecornet:
@afadil After more testing, I think this (or the equivalent functionality) is even more needed after #271 for assets which don't have quotes every day (or more specifically a mix of assets which receive quotes on different schedules) and the removal of the `-filled` quotes.

afadil:
@jeromecornet Are you still experiencing the issue with the code in the Main branch? Can you share some mutual fund tickers so I can test with them?

jeromecornet:
Yes I'm running off main and still having the issue that this PR fixes.

You can try using `0P0000UF8O.TO` as a mutual fund that publishes prices once a day to reproduce; you may have to mix any regular stock (say `RY.TO`) that has real-time quotes.

jeromecornet:
Sorry I didn't see you just merged #280 . Let me test again after this

jeromecornet:
From the code based off 25592e76147b267721978f902970f52ea907859d (right after you merged #271), I wanted to clear all the stuff my code added so I ran this to re-execute the migration that #271 introduced.
```sql
delete from holdings_snapshots;
delete from daily_account_valuations;
```

The dashboard is sort of OK, but only shows data for the days that have all available quotes. In my case, right now if I select 1W it only shows quotes for June 4, 5, 6 and 9th.  I had also backfilled historical data up to 2019, with manually imported historical quotes (1 per month) and it.

With this PR merged (and after re-running the migration to the historical valuations) the dashboard shows the graph with 1 day per calendar day. 

Actually the quickest way to reproduce the issue is to 'skip symbol lookup' (which is a quick way to create an asset with a quote that doesn't update)
Start with the code on `main`:
- Activities->Add Manually
- Add Asset , Skip Symbol Lookup, Symbol: TEST_VALUATION, Shares: 1000000, Average Cost: 1 , Account: some account, Date: 2025-01-01
- Add Activity

Look at the dashboard, '1Y': you don't see a $1M spike in valuation, even though you had just added an asset worth 1M on Jan 1st.

Now checkout my branch, and look at the dashboard.
You now see a $1M spike in valuation for January 1st

This is because under the cover, adding an asset creates a MANUAL quote for that day for that symbol, but then the quote isn't updated ever again.

I hope this makes sense

afadil:
Great! will check this. Thanks @jeromecornet

PR #271, opened 2025-06-10
1 1 4 fix

afadil:
This pull request introduces several updates across the codebase, focusing on database schema changes, feature enhancements, and code cleanup. The highlights include a database migration to add a new column, improvements to the `HoldingsCalculator` for handling base currency conversions, and simplifications in the `MarketDataService` by removing unnecessary methods.

### Database Schema Changes:
* Added a new column `net_contribution_base` to the `holdings_snapshots` table with a default value of `'0'`, and created indices to optimize queries on `daily_account_valuation`. Corresponding changes were made to the migration files (`up.sql` and `down.sql`). [[1]](diffhunk://#diff-bc4a7d2e3d22893f6e7d2c0ed7f4f1b85a77c5b656678003556c13e2e0433decR1-R13) [[2]](diffhunk://#diff-690418198fcc46e2b31b97a21622843ab502e7b5c10481d0700f42dbb2159022R1-R4)

### Feature Enhancements:
* Updated `HoldingsCalculator` to handle base currency conversions for deposits, withdrawals, and holdings adjustments. Introduced a new `base_currency` field using `RwLock` to enable thread-safe updates. [[1]](diffhunk://#diff-935d42394319e3cc0960642f3ea963a2f393c755890e666a8f6ab60d26f17696L11-R26) [[2]](diffhunk://#diff-935d42394319e3cc0960642f3ea963a2f393c755890e666a8f6ab60d26f17696R48) [[3]](diffhunk://#diff-935d42394319e3cc0960642f3ea963a2f393c755890e666a8f6ab60d26f17696L167-R173) [[4]](diffhunk://#diff-935d42394319e3cc0960642f3ea963a2f393c755890e666a8f6ab60d26f17696R280-R344) [[5]](diffhunk://#diff-935d42394319e3cc0960642f3ea963a2f393c755890e666a8f6ab60d26f17696R530-R548)
* Selections in Performance page are now persisted to local storage
* Added another popular index to the benchmark selection on the performance page. Users can also search for a new index or ETF to benchmark against.
* Added a small chart to the Top 10 dividends source on the Income Page.
* Fixed simple return calculation for accounts with a different currency

### Code Simplifications:
[...]

PR #273, opened 2025-06-10
Update methods signature and tests (build fix)

jeromecornet:
@afadil 
running `pnpm tauri build` fails on main right now because #271 introduced a couple of method signature changes.
This PR updates the failing files. 
- fixed gainLossAmount calculation call
- introduced a helper method to convert a `SimplePerformanceMetrics[]` into `AccountValuations[]` for the `Goal` tests.


afadil:
Thanks @jeromecornet

PR #288, opened 2025-06-13
Allow bulk transactions import into multiple accounts

jeromecornet:
It's pretty common for a single broker to hold multiple kinds of accounts, for example in Canada you can have a RRSP, TFSA, RESP and a taxable investment account. Unfortunately, the current CSV import workflow only allows importing transactions one account at a time, even though the broker may be able to export all transactions at once in a single CSV file, which makes it a pain if you manage a lot of accounts.

This PR adds the ability to optionally override the account ID for every row of the CSV. It does not fundamentally change the flow (where the CSV import mapping is stored per account), but allows combining imports for multiple accounts in a single shot.  Ideally the CSV import mapping would be stored per broker, not per account, but this seems to big of a change and I'm not sure it brings anything. The downside of this approach is that you have to select the same account every time you want to import from the same broker.

This flow is very much modelled after the symbols mapping import: if the import is set to use an account column, and an account mapping is not found for a given value of the account column, the row is added to the mapping table to select an account to map to.

A migration was added to add the account mapping to the activity import profile (defaulting to empty).

Just for completeness sake, the account name is displayed in the import screen (even when not overriding it) for every transaction in the preview.

### DIFF
--- a/src-core/migrations/2025-06-11-133126_account_import_mapping/down.sql
+++ b/src-core/migrations/2025-06-11-133126_account_import_mapping/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE activity_import_profiles DROP COLUMN account_mappings;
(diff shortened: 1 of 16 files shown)


jeromecornet:
I addressed the review comments and rebased to the latest main

afadil:
Thanks @jeromecornet. This is very useful one.

PR #292, opened 2025-06-14
Feat/update-balance-for-cash

Ripwords:
- Update ActivityForm to fetch the latest account performance when handling UPDATE_BALANCE activity type.
- Adjust logic to calculate the current balance based on the latest account data.
- Ensure activity creation is prevented if there is no change in balance after calculation.

image

This pull request introduces the new `UPDATE_BALANCE` activity type, which allows users to adjust account balances directly. It includes updates across multiple files to integrate this new functionality, including schema changes, UI enhancements, and logic for handling balance updates.

When submitting a new balance, if it's more than the account's previous balance, it automatically creates a deposit activity, and if less, then a withdraw activity

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       '@fontsource/ibm-plex-mono':
-        specifier: ^5.2.5
-        version: 5.2.5
+        specifier: ^5.2.6
+        version: 5.2.6
       '@hookform/resolvers':
         specifier: ^3.9.1
         version: 3.10.0(react-hook-form@7.55.0(react@18.3.1))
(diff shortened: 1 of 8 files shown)


afadil:
Thank you for this PR.  
I believe we should not use an new Activity type (Transaction) but rather as a regular frontend action that generates a deposit or withdrawal activities. This will keep the schema and model more aligned with transactions.  Also Not sure we need Fee input and use different currency checkbox here.

It would be nice to add an edit icon on the cash balance header in the account page.
CleanShot 2025-06-15 at 10 26 55@2x

Ripwords:
Yea i agree, to be honest i should've spent more time studying the codebase, but i forked the repo and made these changes as a POC and it actually worked pretty well for me. I did wanted to make the changes as minimal as possible so adding the transaction type definitely wasn't ideal.

Ripwords:
@afadil Before we implement the direct edit in the account page, in the meantime I've added a checkbox instead, and removed the *Update Balance* as an ActivityType.

image

afadil:
Closing this because update cash balance was added in [v1.1.6](https://github.com/afadil/wealthfolio/releases/tag/v1.1.6)

Issue #296, opened 2025-06-16
[Bug] 2 About buttons on the task bar - both don't do anything


fallenczar:
V 1.1.3
Currently there are 2 about buttons: One under Wealthinfo, Second under Help - both do not show the software version. I would suggest removing one of these buttons and making the remaining functional.

afadil:
On which operating system are you encountering this issue?

jcheong-1:
I'm on Windows 10 64 bit, clicking on Help then About, nothing happens.

fallenczar:
Windows 11

PR #314, opened 2025-07-02
Retain original activity date when updating instead of resetting to 4 PM

eldossjogy:
Changelog:
- Retained date when updating an activity
- Added version to the nav sidebar

<details>
  <summary>Screenshot of changes</summary>

  <br>

  **Before:**

  image  
  image

  **After:**

  image  
  image

</details>

### DIFF
--- a/src/pages/activity/components/activity-form.tsx
+++ b/src/pages/activity/components/activity-form.tsx
@@ -72,9 +72,7 @@ export function ActivityForm({ accounts, activity, open, onClose }: ActivityForm
     comment: activity?.comment || null,
     assetId: activity?.assetId,
     activityDate: activity?.date ? (() => {
-      const date = new Date(activity.date);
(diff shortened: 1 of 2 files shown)


afadil:
Any reason we want the version in the sidebar?
Version is already in about and help menu.

eldossjogy:
I wasn’t aware that the version info was already in the menu. However, I just tried clicking on About under both Help and Wealthfolio on Windows 11, and nothing happens. Is this a bug?

afadil:
Should be a bug. I'll have a look

jcheong-1:
> Should be a bug. I'll have a look

I believe the About dialog box not working is reported from #296 as well

eldossjogy:
Apologies @afadil for the oversight. I've reverted the changes related to the version number in this PR. It now only includes updates relevant to the activity date logic.

afadil:
Thanks @eldossjogy. I'll fix #296 in the next version. About seems broken in Windows indeed

PR #321, opened 2025-07-09
fix : add manual activity overflow issue

krish4747:
Fixed a scrolling issue in the "Add Manual Activities" third pane. Content was overflowing without scroll, breaking UX for longer input forms.

### Before vs After

| Before | After |
|--------|-------|
|Before | After|


afadil:
Thanks @krish4747

PR #342, opened 2025-08-06
Adding withdrawal to contribution limits

### LINKED ISSUE #338: The limits doesn't include withdrawals
I setup limits and marked the accounts to be included for the limits. It looks like deposits are being counted but withdrawals are not counted.

vaishsriram:
This PR calculates limits not only from deposits but from withdrawals also.

The solution is tested locally.

Closes https://github.com/afadil/wealthfolio/issues/338

### DIFF
--- a/src-core/src/activities/activities_repository.rs
+++ b/src-core/src/activities/activities_repository.rs
@@ -419,6 +419,53 @@ impl ActivityRepositoryTrait for ActivityRepository {
         Ok(converted_results)
     }
 
+    /// Retrieves withdrawal activities for specified accounts within a year as raw data
+    fn get_withdrawal_activities(
+        &self,
+        account_ids: &[String],
+        start_date: NaiveDateTime,
+        end_date: NaiveDateTime,
+    ) -> Result<Vec<(String, Decimal, Decimal, String, Option<Decimal>)>> {
+        let mut conn = get_connection(&self.pool)?;
+
+        // Use a proper join with explicit ON condition
+        let results = activities::table
+            .inner_join(accounts::table.on(activities::account_id.eq(accounts::id)))
+            .filter(accounts::id.eq_any(account_ids))
+            .filter(accounts::is_active.eq(true))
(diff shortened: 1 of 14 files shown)


afadil:
@vaishsriram, thanks for the contribution! I'm not sure about this, though.  
Actually, contribution limits are used to track and enforce legal caps on how much users can deposit into regulated (tax-advantaged) accounts annually (e.g., RRSP, TFSA, FHSA in Canada or 401(k), IRA, HSA in the U.S.). These limits are defined by tax law and vary by account type.

-  Only **deposits** count against the limit  
-  **Withdrawals** do not affect current-year room  
-  For some accounts, TFSA withdrawals restore room in the *following* year, not the current year.

vaishsriram:
> @vaishsriram, thanks for the contribution! I'm not sure about this, though. Actually, contribution limits are used to track and enforce legal caps on how much users can deposit into regulated (tax-advantaged) accounts annually (e.g., RRSP, TFSA, FHSA in Canada or 401(k), IRA, HSA in the U.S.). These limits are defined by tax law and vary by account type.
> 
> * Only **deposits** count against the limit
> * **Withdrawals** do not affect current-year room
> * For some accounts, TFSA withdrawals restore room in the _following_ year, not the current year.

@afadil thanks for the detailed response, and apologies for the delay in getting back.

I see what you mean for regulated, tax-advantaged accounts (RRSP, TFSA, 401k, etc.) where withdrawals don’t affect contribution room. That makes sense.

That said, I was thinking more about a different use case: after-tax contribution accounts where a user might still want to set an annual limit for themselves, and both deposits and withdrawals would affect that limit.

From your suggestion, it sounds like modeling this as a goal with time constraints (e.g., yearly savings goals) could be a better fit. The one wrinkle I see is that the same account can’t be assigned to multiple goals today, so I’m curious if you think extending the goal system (or making contribution limits a first-class property of an account) would be a more flexible path forward.

Do you think goals with dates offer a more flexible alternative, or is the current setup still the better option?

PR #376, opened 2025-08-31
feat: add privacy toggle to holding page

andreaangiolillo:


afadil:
I prefer keeping the UI minimalist and not adding the privacy toggle everywhere. 
Version 2.0 introduced an app launcher (CMK+K) that can be triggered from anywhere and has a show/hide balance action.

PR #378, opened 2025-09-01
added function to import quotes

FilipHarald:


afadil:
Nice! This is a popular request. Could you adapt it to match the look and feel of the activities CSV import for visual consistency? Additionally, instead of having a separate section in the settings menu, I suggest replacing it with a button inside Market Data to display the import wizard.

FilipHarald:
@afadil what do you think about this?

Screenshot from 2025-09-28 09-40-29
Screenshot from 2025-09-28 09-40-40
Screenshot from 2025-09-28 09-40-56
Screenshot from 2025-09-28 09-41-20
Screenshot from 2025-09-28 09-41-27
Screenshot from 2025-09-28 09-41-36

Issue #402, opened 2025-10-22
Cannot edit Tax transactions


GiamBoscaro:
When I try to edit a tax transaction, for example because I have forgotten a description, I get the error:

`quantity: Number must be greater than 0`

Obviously the amount paid in tax is already greater than 0 so I don't know which non existing field is trying to validate.

PR #404, opened 2025-10-22
use nonnegative() for quantity for other type

vafada:
fixes: #402

Updating an activity that is of type "other" is currently erroring out due to `quantity` having a value of zero. 

Screenshot 2025-10-22 at 7 12 16 PM

The "other" activities (split/fee/tax) doesn't  have a "quantity" input field so upon creating a new one, "quantity" is defaulted to zero. On edit, zod fails validation because the `quantity` value is 0 and `positive() requires the value to be strictly greater than zero (greater than 0).`

This PR changes the `quantity` to be `.nonnegative() ensures the value is greater than or equal to zero (0 or more)` so zod doesn't error out on submit

### DIFF
--- a/src/pages/activity/components/forms/schemas.ts
+++ b/src/pages/activity/components/forms/schemas.ts
@@ -128,7 +128,7 @@ export const otherActivitySchema = baseActivitySchema.extend({
   ]),
   assetId: z.string().min(1, { message: 'Please select a security' }).optional(),
   amount: z.coerce.number().min(0).optional(),
-  quantity: z.coerce.number().positive().optional(),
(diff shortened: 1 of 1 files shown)


src/pages/activity/components/forms/schemas.ts
```diff
@@ -128,7 +128,7 @@ export const otherActivitySchema = baseActivitySchema.extend({
   ]),
   assetId: z.string().min(1, { message: 'Please select a security' }).optional(),
   amount: z.coerce.number().min(0).optional(),
```
vafada, line 130: this could probably be also `amount: z.coerce.number().nonnegative().optional(),`

PR #447, opened 2025-11-23
fix: server responds 404 on requests for valid SPA routes

### LINKED ISSUE #433: Navigating to any page except root/homepage returns 404
Hello,
I've installed self-hosted (Docker) version v2.0.0 of Wealthfolio, but when I access any other URL except for homepage, it gives me 404 error. This also happens when refresh/F5 is pressed.

Steps to reproduce:
1. Install Wealthfolio using instructions in https://github.com/afadil/wealthfolio?tab=readme-ov-file#running-the-container.
2. Login to the app.
3. Navigate to http://localhost:8088/dashboard, or press F5 when on /dashboard URL.

Actual behavior:
- It returns 404.

Expected behavior:
- It correctly loads dashboard page (200 OK), or at least redirects to root (/).

Thank you in advance for addressing this.

chrismilson:
While navigating around the web client, the URL path changes to match the page that the user is on. This happens client side, because the web client is a Single Page App (SPA). If the user tries to load the app directly on a non "/" path, (e.g. by clicking an external link) or tries to refresh their browser while at such a path, the server responds with an HTTP404 error.

This updates the server app router's fallback service to respond with the static index.html instead of falling through and responding 404.

The issue does not manifest when developing the application locally because the frontend is served directly by vite, not as static assets by the server.

Fixes #433

### DIFF
--- a/src-server/src/main.rs
+++ b/src-server/src/main.rs
@@ -10,14 +10,19 @@ mod secrets;
 use api::app_router;
 use config::Config;
 use main_lib::{build_state, init_tracing};
-use tower_http::services::ServeDir;
+use tower_http::services::{ServeDir, ServeFile};
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     let config = Config::from_env();
     init_tracing();
     let state = build_state(&config).await?;
-    let router = app_router(state, &config).fallback_service(ServeDir::new(&config.static_dir));
+    let static_dir_path = std::path::Path::new(&config.static_dir);
+    let router = app_router(state, &config).fallback_service(
+        ServeDir::new(static_dir_path)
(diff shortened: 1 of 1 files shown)


chrismilson:
I didn't see an obvious way to test since the fallback service is added to the app router in the main function, instead of in the api module.

chrismilson:
This is what the site looks like if the user tries to load a route that is not handled by the SPA:
image

 Will this require changes to the frondend code, and will those frontend changes need to be included in this PR?

chrismilson:
This also includes an unrelated change to explicitly include the container registry in the Dockerfile image references, please let me know if that should be in its own PR (and whether it would be appreciated or not).

afadil:
Sorry, I already fixed this in a previous commit.
also added a custom 404 page.
closing this PR

PR #448, opened 2025-11-23
Fix ctx.api.activities.getAll with accountId

### LINKED ISSUE #446: type error when using `activities.getAll()` with account id
Code:
```
        await ctx.api.activities.getAll("asd");
```

Error:
```
Uncaught (in promise) invalid args `accountIdFilter` for command `search_activities`: invalid type: string "asd", expected a sequence
```

andrzejressel:
Fixes #446

Issue was hidden due to type removal in `use-activity-search.ts`. I've also updated the names in struct to represent that they contain arrays now.

I wonder if maybe this argument should be required alltogether and empty array will mean the same as undefined now.

### DIFF
--- a/src/commands/activity.ts
+++ b/src/commands/activity.ts
@@ -10,8 +10,8 @@ import {
 import { getRunEnv, RUN_ENV, invokeTauri, invokeWeb, logger } from "@/adapters";
 
 interface Filters {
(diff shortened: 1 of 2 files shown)


andrzejressel:
I guess docker build does not like pull requests from external directories?

PR #469, opened 2025-12-03
feat: use local time as default for new activities

marcosdsanchez:
# PR Description

## Title
feat: use local time as default for new activities

## Description
This PR updates the default time for new activities to use the user's current local time instead of a fixed 4:00 PM (16:00).

Changes applied to:
- `src/pages/activity/components/activity-form.tsx` (Desktop Activity Form)
- `src/pages/activity/activity-manager-page.tsx` (Activity Manager Page)
- `src/pages/activity/components/mobile-forms/mobile-activity-form.tsx` (Mobile Activity Form)

## Motivation
Previously, all new activities defaulted to 4:00 PM (market close time). While this might be suitable for some stock trades, it is often inaccurate for other types of activities, particularly **cash transactions** (deposits, withdrawals, transfers) which happen at specific times throughout the day.

By using the local time:
- Users get a more accurate default timestamp for their activities.
- It reduces the friction of having to manually adjust the time for every transaction that didn't happen exactly at market close.
- It improves the tracking of cash flows and intraday balances.

## Verification
- Verified that clicking "Add Activity" on desktop now presets the date/time picker to the current local time.
- Verified that the mobile activity form also defaults to the current local time.
- Verified that the standalone Activity Manager page respects the local time default.

### DIFF
--- a/src/pages/activity/activity-manager-page.tsx
+++ b/src/pages/activity/activity-manager-page.tsx
@@ -126,9 +126,7 @@ const ActivityManagerPage = () => {
     isDraft: initialActivity?.isDraft ?? false,
     comment: initialActivity?.comment ?? null,
     assetId: initialActivity?.assetId,
-    activityDate: initialActivity?.date
-      ? new Date(initialActivity.date)
-      : new Date(),
+    activityDate: initialActivity?.date ? new Date(initialActivity.date) : new Date(),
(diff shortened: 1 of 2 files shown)


marcosdsanchez:
@afadil let me know if you are interested in the changes so that I either merge with latest or discard the PR. Thanks!

PR #522, opened 2026-01-01
Allow INTEREST activities to preserve custom symbols

samsonluk:
- Remove INTEREST from $CASH symbol forcing in activity forms
- Update CSV import validation to support custom symbols for INTEREST
- Maintain backward compatibility: empty symbols still default to $CASH
- Enables tracking bond coupon payments with specific bond symbols


afadil:
closing this. added in v3.0.5 with the new v3 system

PR #532, opened 2026-01-04
Add copy symbol button

eblancof:
Simple feature to let the user copy the symbol of the opened asset. 

As text copying is disabled in the application (although it could be enabled in the web version), this is something I found lacking.


afadil:
added in v3.8

PR #536, opened 2026-01-08
Move previous period income to be tracked on same axis as income

### LINKED ISSUE #535: Income History Previous Period Income is plotted on the wrong axis
# Issue
When viewing the Income History chart, the Previous Period Income is plotted against the same axis as Cumulative Income, but it is being compared against Monthly Income, so the axis used to plot the two should be the same.

## UI Location
Insights > Income > Income History

t-miller:
Fixes #535


afadil:
already added in main

PR #538, opened 2026-01-08
Replace "primary" with "addons" for addon dynamic items

andrzejressel:
They are of course the same interfaces, but logically these "dynamic items" refer to addons


afadil:
done in v3

PR #539, opened 2026-01-09
Add new belarusian ruble currency

somedev:
This adds a BYN - currency code for Belarusian Ruble which is commonly used now instead of legacy BYR.


somedev:
@afadil could you please take a look?

PR #542, opened 2026-01-12
Increase default decimal precision

glowacki-dev:
Default harcoded decimal precision of 6 becomes problematic when tracking cryptocurrency transactions. Lots of them are divisible with a higher precision which causes small rounding errors when tracking values and holdings.
Most of the time it would be negligible (it would be sub-cent difference) but a difference of 0.000001 BTC is roughly $0.1 which can add up to significant numbers with multiple transactions.

Changes made:
- Increase default precision from 6 to 8 (enough to accurately track BTC holdings)
- Replaced hardcoded precision in UI components with constants, which would make it easier to change when needed and keep in sync with backend
- Updated asset lots table to properly format asset quantity

This PR isn't a real fix for the underlying issue - to do this the precision should probably be user-configurable or automatically determined based on tracked assets. It makes the issue less annoying though and lays some groundwork for allowing proper configuration for this value

### DIFF
--- a/src/pages/asset/asset-lots-table.tsx
+++ b/src/pages/asset/asset-lots-table.tsx
@@ -7,7 +7,7 @@ import {
   TableRow,
 } from "@/components/ui/table";
 import { Lot } from "@/lib/types";
-import { formatAmount } from "@wealthfolio/ui";
+import { formatAmount, formatQuantity } from "@wealthfolio/ui";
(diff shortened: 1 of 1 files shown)


afadil:
is it possible to update this PR?
Thanks

afadil:
done in v3.0.4

Issue #597, opened 2026-02-16
v3.0.0-beta.4: Impossible to import CSV with manual symbols


leppa:
In v2.x importing CSVs would automatically create manual symbols for the ones that don't exist on Yahoo Finance. v3.x introduced symbol validation during import. Unfortunately, this completely prevents importing records with manual symbols. See the detailed steps below.

## Environment

**main** branch (eb502cf3737f78420ab799acf69f663ab2528215) with PR #593 applied, running in a dev container.

## Steps to reproduce

1. Start the Import from CSV process and upload a CSV file with the symbol no longer exists (delisted company):

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/33690a12-303f-4053-a9a9-fa23a7bc40ce" />

2. Map the fields and activity types, if necessary:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/e645f990-a964-447f-bc7a-f166c7816eac" />

3. Reach the Review page.

3. 1.  Observe the unrecognized symbol error:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/5e03154e-0239-4200-b166-219ec39c9749" />

3. 2. Click on the HNWK symbol:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/1b8a67a4-ced2-4c9e-b991-b39dbff417b8" />

3. 3. Only after editing the input field (e.g., removing and re-adding a character), you will get an option to create a manual symbol:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/68cea562-05d7-41f0-bd0d-75e96b8faebc" />

3. 4. Create the symbol:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/b0f8eec4-b1d1-4d47-bc9d-c84eca9d91d2" />

4. 4. After creating the symbol, press the "Apply 1 mapping" button:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/9414f515-a24e-40d9-8745-00dc69d83825" />

### Actual behavior

You are back on the unrecognized symbol error from step 3.1.:

<img width="1866" height="960" alt="Image" src="https://github.com/user-attachments/assets/3fa2c61d-a17e-4565-b86f-5fc9377fb9b5" />

### Expected behavior

The unrecognized symbol error disappears and the transaction is imported in the next step.

## Additional information

[...]

PR #598, opened 2026-02-18
Some improvements to CSV import of manual symbols

leppa:
## Description

This change includes several improvements:

### Include existing manual symbols in activity import validation

Existing manual symbols were not included in the `check_activities_import` reply and were incorrectly marked as invalid with "Could not find SYMBOL in market data" error, forcing users to manually remap already-existing symbols.

Reorder validation logic to look up existing assets before validating market data requirements. When an existing asset is found, its `quote_mode` is populated, allowing manual symbols to bypass market data
validation.

### Preserve user-provided symbol name on import

When user created a manual symbol during import the user provided name was "lost" and symbol ticker was set as the name instead.

Update both UI and server code to include `symbolName` into the request and pull `symbolName` from the response. Don't overwrite symbol name with `normalized_symbol` in the `check_activities_import` if it was already set.

### Add "Mark Custom" and "Mark All Custom" buttons in the CSV import

Alow to quickly mark unresolved symbols as manual during CSV import, instead of requiring manual search and mapping one-by-one.

Two buttons were added:
- "Mark Custom" to the right of each unresolved symbol - will mark only corresponding symbol as manual
- "Mark All Custom" at the top right - will mark all symbols as manual

Manual symbols will be created as an equity with the symbol as both its ticker and its name.

image

### Set `quoteMode` in the `symbolMappingMeta`

`quoteMode` was defined in the `importMappingSchema` but was never set.

### Add type safety for `quoteMode`

Change `quoteMode` in `DraftActivity`, `importMappingSchema`, and `importActivitySchema` from generic string to `quoteModeSchema` (`"MANUAL" | "MARKET"`), ensuring that it can only take valid values.

## Checklist

### DIFF
--- a/apps/frontend/src/components/ticker-search.tsx
+++ b/apps/frontend/src/components/ticker-search.tsx
@@ -315,7 +315,22 @@ const TickerSearchInput = forwardRef<HTMLButtonElement, SearchProps>(
     );
 
     useEffect(() => {
-      if (selectedResult) return;
+      if (selectedResult) {
+        const exchangeDisplay =
+          selectedResult.exchangeName || getExchangeDisplayName(selectedResult.exchange);
+        const exchangeSuffix = exchangeDisplay ? ` (${exchangeDisplay})` : "";
+        const displayText = `${selectedResult.symbol} - ${
+          selectedResult.longName || selectedResult.shortName || selectedResult.symbol
+        }${exchangeSuffix}`;
+        setSelected(displayText);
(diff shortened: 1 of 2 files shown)


leppa:
@afadil, thanks for fixing #597 so quickly. I rebased this PR on top of your latest changes. It fixes few remaining issues that weren't addressed in your fix and also adds buttons that allow to quickly mark unrecognized symbols as custom.

PR #634, opened 2026-03-01
Optimize portfolio updates back to earliest activity

geordie:
## Description

This PR changes the logic for how portfolio snapshot and valuations recalculation works. Previously there were two styles of updates, full & incremental. Full updates would recalculate snapshots and valuations for every day since the beginning of the portfolio. For portfolios with long histories, this can take a long time. Whether actions resulted in full or incremental recalculations was determined by a boolean parameter. For activity adds / updates / deletes, a full re-calculation was done every time.

This is more work than the app needs to do.

This implementation changes the recalculation behaviour on activity CRUD operations to only go back as far as the date of impacted activity.

To make this work, I added two enums that provide the option to do full recalculation, incremental since last date, or from a specified date

```
pub enum ValuationRecalcMode {
    Full,
    IncrementalFromLast,
    SinceDate(NaiveDate),
} 
```

```
pub enum SnapshotRecalcMode {
    Full,
    IncrementalFromLast,
    SinceDate(NaiveDate),
}
```

Then, when activities change, we trigger the portfolio updates to only go back as far as the activity date using the SinceDate(activity_date) value in the enum.

All callers other than activity changing callers trigger have been updated to use these enums to update portfolios the same way they did before. Tests are also updated.

### DIFF
--- a/crates/core/src/portfolio/snapshot/snapshot_service_tests.rs
+++ b/crates/core/src/portfolio/snapshot/snapshot_service_tests.rs
@@ -2561,7 +2561,10 @@ mod tests {
             fx,
         );
 
-        let _ = svc.calculate_holdings_snapshots(None).await.unwrap();
+        let _ = svc
+            .recalculate_holdings_snapshots(None, SnapshotRecalcMode::IncrementalFromLast)
+            .await
+            .unwrap();
 
         let frames = snapshot_repo.get_saved_snapshots();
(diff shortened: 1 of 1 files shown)


geordie:
hmmm. the failing test passes locally for me, but not in CI. looks like I missed pulling a commit or something.

geordie:
aborting pull request because I don't like the force push. starting again.

PR #644, opened 2026-03-02
ci(rust): align test feature flags with clippy to enable artifact reuse

### LINKED ISSUE #643: [CI][rust] Faster rust pr-checks
The rust pr-check is currently slow (~7-15 minutes). There are a couple of ways it could be sped up. The least intrusive one is probably to add --all-features to the test build. That should allow for asset resuse between the clippy build and the test build.

geordie:
Add --all-features flag to the Test step of the rust pr-check. This should trigger asset resuse between the clippy step and the test step.

Resolves #643


geordie:
this didn't help

PR #666, opened 2026-03-04
Add support for options, bonds, and precious metals as instrument types

triantos:
## Summary

This PR adds first-class support for **options**, **bonds**, and **precious metals** as distinct instrument types throughout the Wealthfolio stack. Previously, all assets were implicitly treated as equities. Now each asset carries an `instrument_type` field that drives type-specific validation, market data fetching, and UI presentation.

### Key changes across 8 commits:

- **Data model**: New `InstrumentType` enum (Equity, Crypto, Fx, Option, Metal, Bond) with DB migration adding `instrument_type` columns to `assets` and `activities`
- **Parallel enrichment**: Asset enrichment during import now runs up to 5 concurrent tasks with SSE progress events
- **Asset enrichment before portfolio**: Enrichment job runs before portfolio calculations, not just during import
- **Frontend UI**: Instrument type badges, enrichment progress indicators, bond/option-specific form fields
- **Bond providers**: New `BondQuoteMetadata` and provider infrastructure for bond market data
- **CSV import**: `instrumentType` column support, typed symbol prefixes (`bond:ISIN`, `option:OCC`), alias normalization (e.g., `FIXED_INCOME` → `BOND`)
- **Activity filter**: New instrument type dropdown filter on the activities page
- **Tests & docs**: 7 backend tests, 32 frontend tests, architecture documentation

## Test plan

- [x] Backend: 1,177 tests pass (`cargo test`)
- [x] Backend lint: `cargo clippy` — 0 errors
- [x] Frontend: 63 tests pass across 4 test suites (instrument-type, validation-utils, activity-type-mapping, activity-utils)
- [x] Frontend lint: ESLint — 0 errors
- [x] `parse_instrument_type` recognizes all canonical types and aliases including BOND
- [x] `normalizeInstrumentType` (frontend) handles canonical types, aliases, case-insensitivity, whitespace/hyphens
- [x] `splitInstrumentPrefixedSymbol` parses `bond:`, `option:`, `crypto:`, `equity:` prefixes
- [x] CSV import: instrument type from column, from typed prefix, column takes precedence over prefix
- [x] OCC option symbols (up to 21 chars) pass schema validation
- [x] ISIN bond symbols (12 chars) pass schema validation
- [x] Bond import with existing asset enriches name and currency
- [x] Option import with existing asset enriches name
- [x] DB migration is idempotent (`INSERT OR IGNORE`)
- [ ] Manual: Add a bond activity via UI form and verify it saves with `instrument_type = BOND`
- [ ] Manual: Import CSV with `instrumentType` column containing bond/option rows
- [ ] Manual: Filter activities page by instrument type

### DIFF
--- a/apps/frontend/src/pages/activity/components/forms/__tests__/buy-form.test.tsx
+++ b/apps/frontend/src/pages/activity/components/forms/__tests__/buy-form.test.tsx
@@ -50,6 +50,10 @@ vi.mock("../fields", () => ({
     </div>
   ),
   AdvancedOptionsSection: () => <div data-testid="advanced-options-section" />,
+  AssetTypeSelector: ({ name }: { name: string }) => (
+    <div data-testid={`asset-type-selector-${name}`} />
+  ),
+  OptionContractFields: () => <div data-testid="option-contract-fields" />,
   createValidatedSubmit: vi.fn((_form, handler) => handler),
 }));
 
--- a/apps/frontend/src/pages/activity/components/forms/buy-form.tsx
+++ b/apps/frontend/src/pages/activity/components/forms/buy-form.tsx
@@ -33,53 +33,64 @@ const assetMetadataSchema = z
   .optional();
 
 // Zod schema for BuyForm validation
-export const buyFormSchema = z.object({
-  assetType: z.enum(["stock", "option", "bond"]).default("stock"),
-  assetKind: z.string().optional(),
-  accountId: z.string().min(1, { message: "Please select an account." }),
-  assetId: z.string().default(""),
-  activityDate: z.date({ required_error: "Please select a date." }),
-  quantity: z.coerce
-    .number({
-      required_error: "Please enter a quantity.",
-      invalid_type_error: "Quantity must be a number.",
-    })
-    .positive({ message: "Quantity must be greater than 0." }),
-  unitPrice: z.coerce
-    .number({
-      required_error: "Please enter a price.",
-      invalid_type_error: "Price must be a number.",
-    })
-    .positive({ message: "Price must be greater than 0." }),
-  fee: z.coerce
-    .number({
-      invalid_type_error: "Fee must be a number.",
-    })
-    .min(0, { message: "Fee must be non-negative." })
-    .default(0),
-  comment: z.string().optional().nullable(),
-  // Advanced options
-  currency: z.string().min(1, { message: "Currency is required." }),
-  fxRate: z.coerce
-    .number({
-      invalid_type_error: "FX Rate must be a number.",
-    })
-    .positive({ message: "FX Rate must be positive." })
-    .optional(),
-  // Internal fields
-  quoteMode: z.enum([QuoteMode.MARKET, QuoteMode.MANUAL]).default(QuoteMode.MARKET),
-  exchangeMic: z.string().nullable().optional(),
-  symbolQuoteCcy: z.string().nullable().optional(),
-  symbolInstrumentType: z.string().nullable().optional(),
-  // Asset metadata for custom assets (name, etc.)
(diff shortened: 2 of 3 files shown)


afadil:
planning to merge this soon.
is it possible to double check this finding from a review by Claude:
 

  ---
  🔴 Open option positions valued at 1/100th actual value

  valuation_calculator.rs:139:
  let market_value = position.quantity * normalized_price * quote_fx_rate;

  contract_multiplier is correctly applied at trade time (BUY/SELL handlers multiply unit_price *
  multiplier for lot cost basis and cash flows), but the valuation path never uses it. Since Yahoo returns
  per-share option premiums:

  - Cost basis for 1 contract at $1.50: 1 × $1.50 × 100 = $150 ✓
  - Market value for 1 contract at $2.00 quote: 1 × $2.00 = $2.00 ✗ (should be $200)

  Fix:
  let market_value = position.quantity * normalized_price * position.contract_multiplier * quote_fx_rate;

  contract_multiplier defaults to Decimal::ONE via serde, so existing equity positions are unaffected.

  ---
  🔴 Option expiry inverts P&L sign

  handle_adjustment for OPTION_EXPIRY calls reduce_lots_fifo(qty) and discards the cost basis:
  let _cost_basis_removed = position.reduce_lots_fifo(qty);
  // No cash effect for expiry

  Since there's no realized gain accumulator in the system, the premium loss doesn't just disappear — it
  gets inverted. The performance service measures delta(market_value - cost_basis) between snapshots:

  1. Before expiry: cost_basis = $500, market_value ≈ $0 → P&L = -$500
  2. After expiry: lots removed, cost_basis = $0, market_value = $0 → P&L = $0
  3. Performance sees: P&L went from -$500 to $0 → records +$500 gain

  A $500 loss shows as a $500 gain in returns.

  Fix: Treat expiration as a sell at zero price instead of raw lot removal. A SELL @ unit_price=0 flows
  through existing FIFO lot matching and naturally realizes the loss (proceeds $0 - cost $500 = -$500). No
  new activity types needed — just change what the OPTION_EXPIRY handler emits internally.

  ---
  🟡 Exercise form without backend handler

  exercise-form.tsx exists but there's no backend subtype constant, compiler expansion, or calculator
  handler. Users would need to manually create separate BUY (underlying) + SELL (option) activities with
  the right quantities (contracts × multiplier shares at strike price). Consider removing the form from
  this PR or adding a clear disclaimer that it's manual-entry only.

triantos:
Apologies, I lost one change when I refactored, that handled the first two. I'll have to look at the second one shortly. I'll make sure to send an updated PR by late tonight if I can.

triantos:
The first item is a legit bug and a one-line fix, which I will push.

As far as I can tell, the second option is actually not a bug. OPTION_EXPIRY is only a thing for accounts in transactions mode, not holdings mode. The bug would only exist for holdings mode accounts. I've confirmed this manually, and also had Claude run a test to verify it as well.

The third one was dead code so I've stripped it out. I had a lot more I've been doing in my branch, but I figured I'd rather get this in and then add to it, than to overwhelm you again. :-)

thank you!

mmaha:
@triantos Thank you for the PR - Very nice. I'll have to test this later this week. 

Quick question - will this allow for short options (-ve quantities) as well? From what I can gather this is not the case currently, correct?

triantos:
> @triantos Thank you for the PR - Very nice. I'll have to test this later this week.
> 
> Quick question - will this allow for short options (-ve quantities) as well? From what I can gather this is not the case currently, correct?

I hadn't done any short options so I'll admit I forgot to test this case. Should be easy to add once these changes are in the core.

mmaha:
Looks like there is a PR for this: https://github.com/afadil/wealthfolio/pull/399

PR #713, opened 2026-03-10
Add bond market data support and fix sync error handling

triantos:
## Summary

- **Register bond market data providers** — US_TREASURY_CALC, BOERSE_FRANKFURT, and OPENFIGI were defined in the database but never instantiated in `create_provider()`, so bond quote fetches silently fell through to Yahoo (which cannot price bonds).
- **Wire up bond metadata for pricing** — `QuoteContext.bond_metadata` was hardcoded to `None`, preventing the US Treasury yield-curve calculator from receiving coupon/maturity data. Now populated from `Asset.bond_spec()`.
- **Skip matured bonds in sync** — Adds `AssetSkipReason::MaturedBond` to avoid wasted provider calls. Writes a par-value quote for matured bonds that have no existing quote data, so they do not show as "no data" in health checks.
- **Enable free bond providers by default** — US_TREASURY_CALC, BOERSE_FRANKFURT, and OPENFIGI require no API keys; enabled by default in migration.
- **Clear sync error state on Manual pricing** — When an asset quote mode is set to Manual, delete its `quote_sync_state` row so stale errors no longer appear in the health panel.
- **Fix bond provider routing** — Asset creation hardcoded `preferred_provider: YAHOO` for all MARKET assets including bonds, preventing US_TREASURY_CALC and BOERSE_FRANKFURT from being selected. Bonds now get no preferred_provider override, letting the rules resolver pick the correct provider.
- **Fix zero-coupon bond metadata** — `build_quote_context()` used `?` on `coupon_rate`, dropping the entire `bond_metadata` when coupon was None. T-bills (zero-coupon) could never be priced. Now defaults missing coupon_rate to 0.
- **Auto-detect T-bill metadata on import** — Bonds with US912797 CUSIP prefix automatically get `coupon_rate=0` and `coupon_frequency=ZERO` during asset creation.
- **Auto-convert CUSIP to ISIN for bonds** — 9-char CUSIPs are converted to 12-char ISINs in `canonicalize_market_identity()` using the existing `cusip_to_isin()` utility, enabling proper provider routing.
- **Add provider logo assets** for Boerse Frankfurt, OpenFIGI, and US Treasury.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes
- [x] `cargo test --workspace` passes (8 new tests for bond routing, zero-coupon metadata, CUSIP-to-ISIN conversion, T-bill pricing)
- [ ] Fresh install creates bond providers as enabled
- [ ] Matured bonds are skipped during sync and get par-value quote
- [ ] Switching asset to Manual clears error count from health panel
- [ ] Bond quote fetches route to US_TREASURY_CALC / BOERSE_FRANKFURT instead of Yahoo
- [ ] Importing a bond with a CUSIP auto-converts to ISIN
- [ ] T-bill imports get zero-coupon metadata automatically

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,4 +92,3 @@ db/
 apps/db/
 apps/tauri/db/
 tasks
-wf.sh
--- a/crates/core/src/assets/assets_model.rs
+++ b/crates/core/src/assets/assets_model.rs
@@ -1010,7 +1010,12 @@ pub fn canonicalize_market_identity(
                 // Auto-convert 9-char CUSIPs to 12-char ISINs for proper
                 // provider routing (US_TREASURY_CALC, BOERSE_FRANKFURT).
                 instrument_symbol = Some(if crate::utils::cusip::looks_like_cusip(&upper) {
-                    crate::utils::cusip::cusip_to_isin(&upper, "US")
+                    let country = match normalized_quote.as_deref() {
+                        Some("CAD") => "CA",
+                        Some("BMD") => "BM",
+                        _ => "US",
+                    };
+                    crate::utils::cusip::cusip_to_isin(&upper, country)
                 } else {
                     upper
                 });
@@ -1205,5 +1210,42 @@ mod tests {
 
         let sym = result.instrument_symbol.expect("should have symbol");
         assert_eq!(sym, "US912797NQ65", "ISIN should pass through unchanged");
+    }
+
+    #[test]
+    fn test_canonicalize_market_identity_cusip_to_isin_canadian() {
+        // Canadian bond CUSIP should produce a CA-prefixed ISIN
+        let result = canonicalize_market_identity(
+            Some(InstrumentType::Bond),
+            Some("135087D26"),
+            None,
+            Some("CAD"),
+        );
+
+        let sym = result.instrument_symbol.expect("should have symbol");
+        assert_eq!(sym.len(), 12, "CUSIP should be converted to 12-char ISIN");
+        assert!(
+            sym.starts_with("CA"),
+            "Canadian bond ISIN should start with CA, got {}",
+            sym
+        );
+    }
+
+    #[test]
+    fn test_canonicalize_market_identity_cusip_defaults_to_us() {
+        // When no currency is provided, CUSIP should default to US
+        let result = canonicalize_market_identity(
+            Some(InstrumentType::Bond),
+            Some("912797NQ6"),
+            None,
+            None,
+        );
+
+        let sym = result.instrument_symbol.expect("should have symbol");
+        assert!(
+            sym.starts_with("US"),
+            "CUSIP with no currency should default to US, got {}",
+            sym
+        );
     }
 }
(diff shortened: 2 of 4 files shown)


afadil:
Thanks for the PR! Solid work. A few items:

  ## Description Accuracy
  I already this some work to wire the new providers(fixed in 43bce1b3 on main.)

  Issues

  **1. Missing metal-price-api.png logo**
  The migration sets logo_filename = 'metal-price-api.png' for METAL_PRICE_API, but no image file is added in this PR (the other three logos are included). This will show a broken
  image in the UI.

  **2. CUSIP-to-ISIN hardcodes "US" country code**
  In canonicalize_market_identity:
  crate::utils::cusip::cusip_to_isin(&upper, "US")
  CUSIPs are used for both US and Canadian securities. A Canadian bond CUSIP would get an incorrect ISIN. Consider deriving the country from the currency parameter (e.g., CAD → "CA"),
  or at minimum add a comment documenting the US-only assumption.

  **3. Unrelated .gitignore change**
  Adding wf.sh to .gitignore looks like a personal script — this should go in your global gitignore (~/.gitignore_global), not in the project.

  ## Improvements

  4. Duplicated sync-state clearing logic
  The same block appears in both update_quote_mode and update_quote_mode_silent:
  if asset.quote_mode == QuoteMode::Manual {
      if let Err(e) = self.quote_service.delete_sync_state(asset_id).await {
          warn!("Failed to clear sync state for {}: {:?}", asset_id, e);
      }
  }
  Consider having update_quote_mode call update_quote_mode_silent + emit the event, to avoid duplication.

  5. Provider config tests duplicate logic instead of testing the actual method
  test_bond_provider_config_is_none and test_equity_provider_config_is_yahoo copy-paste the match expression from the service rather than calling the service method. If the logic
  changes in the service but not the test, the test would still pass. Would be more valuable as integration tests against the actual code path.

  6. and_then + ? pattern in build_quote_context
```
  let bond_metadata = asset.bond_spec().and_then(|spec| {
      Some(BondQuoteMetadata {
          coupon_rate: spec.coupon_rate.unwrap_or(Decimal::ZERO),
          maturity_date: spec.maturity_date?, // ← ? inside and_then
          ...
      })
  });
```
  This works but mixing and_then with ? is easy to misread. Minor readability.

triantos:
Thanks @afadil. To address your points you've raised:
1. The file is already committed (in my original options-bonds-metals change), I just neglected to hook it up to the data provider. This fixes that. A hard refresh of the frontend should pick it up. I confirmed it's in your repo already
2. CUSIP-to-ISIN: Good catch, thank you. I'll push a fix.
3. Fixed
4. Also fixed
5. In this case, I think to fix this properly will require fairly nontrivial work. I will address this in a separate PR, if that's ok. (It looks like I'd need to wire up a mock AssetRepository and QuoteService since new_asset_from_spec is async).
6. Implemented your suggestion. Both work, but I'm happy to match the style you want for the code.

afadil:
Thank you!

PR #727, opened 2026-03-15
refactor(activity): create paired deposit for funded buys

kwaich:
## Description

When a buy activity is marked as "Include cash deposit", automatically create a paired `DEPOSIT` transaction for the same account, amount, and date. This ensures the portfolio cash balance reflects the inflow before the buy executes, giving accurate performance and cash tracking.

### Changes

- **Buy form** — Added "Include cash deposit" checkbox in the Advanced Options section. Creates a separate deposit activity for the total purchase cost.
- **Form schema + config** — Added `includeCashDeposit` boolean field. Payload builder passes `includeCashDeposit: true` when enabled. Preserves existing metadata keys and strips legacy `include_cash_deposit` from metadata to avoid pollution.
- **Mutation logic (`use-activity-mutations.ts`)** — When `includeCashDeposit` is enabled on a BUY, the hook bulk-creates both a `DEPOSIT` (for `quantity * unitPrice * multiplier + fee`) and the `BUY` via `saveActivities`. Option contract multiplier is respected in the deposit amount calculation. Falls back to single `createActivity` when disabled.
- **Duplicate activity** — Preserves metadata when duplicating.
- **Tests** — Added coverage for form schema payload behavior, checkbox default state, and the bulk deposit+buy mutation path.

### Checklist


afadil:
@codex review

afadil:
Not sure about this one. This is what external TransferIn does.

If it's for UX convenience, I wouldn't impact the calculator and the backend; I would just make the frontend create two activities.

Jonjon-prog:
Agree with Afadil on that. 
Better to keep the 2 activities as the same time, like this you don't have any cash balance hide.

kwaich:
yeah that make sense and keeps it simple, thanks

kwaich:
hi @afadil , made the changes to keep it simple and create 2 activities. please help to check, thanks!

PR #736, opened 2026-03-16
Fix metal price lookups for weight-suffixed symbols

triantos:
## Summary
- Strip weight suffix (e.g. "XAU-500G" → "XAU", "XAU-1KG" → "XAU") from metal symbols
  before passing to price providers
- Weight conversion is handled separately by `metal_weight_oz()`, but the suffixed symbol
  was being sent to providers as-is, causing lookup failures

## Test plan
- [ ] Verify price lookups work for weight-suffixed metal symbols (XAU-500G, XAU-1KG)
- [ ] Verify standard metal symbols (XAU, XAG) continue working
- [ ] All existing tests pass


triantos:
Combining with #796 into a single metals PR that also adds quote-pipeline weight scaling.

PR #738, opened 2026-03-17
Normalize OCC option symbols from broker sync to compact format

triantos:
## Summary
- Brokers (via SnapTrade) may return space-padded OCC symbols (e.g. `GE    270115C00340000`) which creates duplicate assets when the same contract also appears in compact form (`GE270115C00340000`)
- Parse incoming option symbols in the broker mapping layer with the existing `parse_occ_symbol()` utility and rebuild in compact format via `build_occ_symbol()` to ensure consistent asset identity

## Test plan
- [ ] Verify existing option positions from Connect sync are not duplicated
- [ ] Verify option symbols from non-Connect sources (CSV import, manual entry) are unaffected
- [ ] All existing tests pass

### DIFF
--- a/crates/core/src/assets/assets_model.rs
+++ b/crates/core/src/assets/assets_model.rs
@@ -382,6 +382,8 @@ impl Asset {
     /// Returns the contract multiplier for this asset.
     ///
     /// For options, this is the number of shares per contract (typically 100).
+    /// For precious metals with a weight suffix (e.g. XAU-1KG), this is the
+    /// weight in troy ounces so that a per-oz spot quote is scaled correctly.
     /// For all other instruments it is 1.
     pub fn contract_multiplier(&self) -> Decimal {
         if let Some(spec) = self.option_spec() {
@@ -389,6 +391,8 @@ impl Asset {
         } else if self.is_option() {
             // Option without metadata — default to standard 100 multiplier
             Decimal::from(100)
+        } else if self.is_metal() {
+            self.metal_weight_oz()
         } else {
             Decimal::ONE
         }


afadil:
**[P2] Option symbol normalization is already handled downstream  This looks redundant now.**
Option symbol normalization is already happening in the current sync path: option positions are normalized in broker sync, and activity symbols are normalized again when `prepare_activities_for_sync()` builds the asset spec. That means asset identity is already canonicalized before `instrument_key` generation, so doing another normalization here does not fix a missing case in the final path.

If we keep any normalization at this layer, it should call the shared `normalize_option_symbol()` helper rather than duplicating parse/rebuild logic, otherwise this can drift from the canonical path and miss broker-specific compact formats.

  **[P1] Metal weight is dropped during pricing I don’t think the metal change is safe to merge as written.**
  
Normalizing `XAU-500G` / `XAU-1KG` to `XAU` fixes provider lookup, but it also changes the pricing unit to a 1 oz spot quote. The valuation path still uses `quantity * price * contract_multiplier`, and `contract_multiplier()` is `1` for metals, so `metal_weight_oz()` is never applied anywhere. As a result, one `XAU-500G` holding would now be valued as one ounce of gold instead of ~16.0754 oz.

So I agree with normalizing for provider lookup, but we also need a weight-aware adjustment before merging this, for example by scaling latest/historical metal quotes by `metal_weight_oz()` (or otherwise carrying the weight through valuation).

triantos:
**Re option normalization**: OK, yes this is no longer needed now that you've merged my other change and normalize_option_symbol().

On the topic of metals, this got lost during one of my refactors. Sorry. I have fixed with an updated [trivial] commit.

PR #746, opened 2026-03-18
fix(frontend): stabilize grouped dashboard account metrics

AtomicTrxn:
## Description

Root cause:
- grouped child account rows on the dashboard hid the secondary metric line whenever gain data was zero or unavailable
- that made some expanded rows render with two lines and others with one line, which looked like rows were switching between balance and gain displays
- the primary metric was already correct; the bug was inconsistent secondary-metric rendering

Fix summary:
- keep the primary line as account value for all rows
- for nested expanded group rows, always render a secondary line
- show gain amount/percent when gain data exists
- show neutral zero-gain output when gain is zero
- show `-` when valuation/gain data is unavailable
- keep group headers and standalone account cards unchanged

Validation:
- `pnpm --filter frontend test -- accounts-summary`
- `pnpm --filter frontend lint`
- `pnpm exec prettier --check apps/frontend/src/pages/dashboard/accounts-summary.tsx apps/frontend/src/pages/dashboard/accounts-summary.test.tsx`

## Checklist


afadil:
changed to logic in this pr https://github.com/afadil/wealthfolio/pull/759

PR #756, opened 2026-03-18
Fix cross-currency buy/sell cash conversion without explicit fx_rate

afadil:
## Summary

When processing buy/sell activities with cross-currency cash in accounts without an explicit `fx_rate`, the calculator now converts the cash amount to the account currency using historical FX rates from the FxService, rather than booking it in the activity currency.

### Holdings Calculator
- Modified buy cash handling to convert activity currency to account currency when `fx_rate` is missing/zero
- Modified sell cash handling to convert activity currency to account currency when `fx_rate` is missing/zero
- Added fallback warning logs when FX conversion fails
- Prevents phantom multi-currency cash positions in single-currency accounting

### Tests
- Updated 7 test cases to expect converted CAD balances instead of activity currency balances
- Updated assertions to verify no phantom USD/activity currency cash balances are created
- Clarified test documentation to reflect new conversion behavior
- Tests now verify FxService historical rates are applied correctly

### DIFF
--- a/crates/core/src/portfolio/snapshot/holdings_calculator.rs
+++ b/crates/core/src/portfolio/snapshot/holdings_calculator.rs
@@ -316,8 +316,24 @@ impl HoldingsCalculator {
                 // Broker converted at transaction time — book in account currency
                 add_cash(state, account_currency, -(total_cost * fx_rate));
             } else {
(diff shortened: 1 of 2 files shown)


afadil:
@codex review

PR #763, opened 2026-03-19
Fix CSV re-import dedup to use raw symbol for stable keys

afadil:
## Problem
CSV re-imports were using resolved asset UUIDs for idempotency key computation, causing keys to differ between the review step and import step, breaking deduplication.

## Solution
- Extract symbol-based asset identifier logic into `symbol_based_asset_id()` helper that uses raw symbol + exchange MIC instead of UUID
- Update `build_import_idempotency_key()` to use the new helper for consistent key generation
- Update CSV import deduplication in repository to call the same helper, ensuring keys match across steps
- Improve duplicate handling with explicit error catching for `UniqueViolation` constraints with debug logging

## Changes
- **activities_service.rs**: Add `symbol_based_asset_id()` helper and refactor key building logic
- **repository.rs**: Switch from bulk insert to per-activity insert with duplicate detection and logging

### DIFF
--- a/crates/core/src/activities/activities_service.rs
+++ b/crates/core/src/activities/activities_service.rs
@@ -306,6 +306,21 @@ impl ActivityService {
             .ok()
     }
 
+    /// Build a symbol-based asset identifier for idempotency key computation.
+    /// Uses raw symbol text + exchange MIC instead of resolved UUID so that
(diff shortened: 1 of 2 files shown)


afadil:
@codex review

PR #782, opened 2026-03-27
feat(fire-planner): add FIRE planner as core feature

alexghirelli:
Adds a FIRE (Financial Independence, Retire Early) Planner directly into the core app, replacing the previous addon approach.

## Rust backend (crates/core + apps/tauri)
- New portfolio/fire module with model.rs (all domain types: FireSettings, FireProjection, MonteCarloResult, ScenarioResult,
SorrScenario, etc.) and calculator.rs (full math engine)
- calculator.rs implements: deterministic projection, Monte Carlo with parallel rayon workers + Box-Muller sampling +
two-regime fat-tailed returns (85% normal / 15% stress), sequence-of-returns risk, scenario analysis, sensitivity analysis,
and strategy comparison
- 8 Tauri commands: get_fire_settings, save_fire_settings (persisted via the app settings store), calculate_fire_projection,
run_fire_monte_carlo (10,000 sims), run_fire_scenario_analysis, run_fire_sorr, run_fire_sensitivity,
run_fire_strategy_comparison (5,000 sims)

## Frontend (apps/frontend)
- Typed invoke adapter at adapters/tauri/fire-planner.ts
- All pages migrated off the addon SDK to native hooks (useFireSettings, usePortfolioData) and direct Tauri/adapter calls
- Monte Carlo and strategy comparison run via Tauri (parallel Rust); Dashboard, Scenario, Sensitivity, SORR remain sync
TypeScript for instant rendering
- /fire-planner route and nav item added

## Checklist

- I have read and agree to the Contributor License Agreement.

By submitting this PR, I agree to the CLA.

### DIFF
--- a/apps/frontend/src/adapters/web/fire-planner.ts
+++ b/apps/frontend/src/adapters/web/fire-planner.ts
@@ -39,12 +39,14 @@ export const calculateFireProjection = async (
   return projectFireDate(settings, currentPortfolio);
 };
 
+const WEB_MAX_SIMS = 10_000;
+
 export const runFireMonteCarlo = async (
   settings: FireSettings,
   currentPortfolio: number,
-  nSims = 10_000,
+  nSims = WEB_MAX_SIMS,
 ): Promise<MonteCarloResult> => {
-  return runMonteCarlo(settings, currentPortfolio, nSims);
+  return runMonteCarlo(settings, currentPortfolio, Math.min(nSims, WEB_MAX_SIMS));
 };
 
 export const runFireScenarioAnalysis = async (
--- a/apps/frontend/src/pages/settings/fire-planner/fire-planner-settings-page.tsx
+++ b/apps/frontend/src/pages/settings/fire-planner/fire-planner-settings-page.tsx
@@ -1,4 +1,10 @@
-import { createGoal, getGoalsAllocation, updateGoal, updateGoalsAllocations } from "@/adapters";
+import {
+  createGoal,
+  getGoals,
+  getGoalsAllocation,
+  updateGoal,
+  updateGoalsAllocations,
+} from "@/adapters";
 import { useFireSettings } from "@/pages/fire-planner/hooks/use-fire-settings";
 import { usePortfolioData } from "@/pages/fire-planner/hooks/use-portfolio";
 import { timezoneToCountry } from "@/pages/fire-planner/lib/timezone";
@@ -19,8 +25,14 @@ export default function FirePlannerSettingsPage() {
   const handleSave = async (updated: FireSettings) => {
     const grossTarget = calculateFireTarget(updated);
 
+    // Prefer the stored linkedGoalId; if missing (e.g. web localStorage cleared),
+    // recover the existing FIRE goal by name to avoid creating duplicates.
     let goalId = updated.linkedGoalId ?? null;
     try {
+      if (!goalId) {
+        const existing = await getGoals();
+        goalId = existing.find((g) => g.title === "FIRE (gross)")?.id ?? null;
+      }
       if (goalId) {
         await updateGoal({
           id: goalId,


afadil:
Hi Alex, thanks for this great feature. I did a quick review with Codex and Claude; I dropped the findings below:
---

## 2. Business Logic & Calculation Review

### 2.1 Core FIRE Formulas — Correct

| Calculation              | Formula                                        | Status                                                                 |
| ------------------------ | ---------------------------------------------- | ---------------------------------------------------------------------- |
| Gross FIRE Target        | `expenses*12 / SWR`                            | Correct                                                                |
| Net FIRE Target          | `max(0, expenses-dayOneIncome)*12 / SWR`       | Correct — only subtracts income streams active from day 1 of FIRE      |
| Coast FIRE               | `netTarget * ((1+inf)/(1+return))^years`       | Correct — present value of nominal target discounted at nominal return |
| Deterministic Projection | Year-by-year compound with contribution growth | Correct                                                                |

### 2.2 Issues Found

#### HIGH: Pension fund balance is disconnected from payout amount

**Files**: `calculator.rs:95-120`, `fire-math.ts:94-120`

The pension fund accumulation (`stepPensionFunds`) tracks a growing fund balance
over time (contributions + investment return). At payout age, the fund is zeroed
out and the user's manually-entered `monthly_amount` kicks in. But there is **no
connection** between the accumulated fund value and the payout amount.

Example: A user sets a pension fund with EUR 50k initial + EUR 200/mo
contribution growing at 4%. By age 67 it accumulates to ~EUR 180k. But the
payout `monthly_amount` is whatever the user typed (e.g., EUR 800/mo = EUR
9,600/yr). The planner shows a large pension fund growing and then just
disappearing when payout starts, replaced by an unrelated income stream.

**Impact**: Users can't tell whether their pension fund is on track to deliver
the income they expect. The fund balance display gives false confidence.

**Recommendation**: Either (a) compute an annuity from the fund balance at
payout age, or (b) clearly label the fund balance as "informational only" and
note that the payout amount is a manual estimate.

#### MEDIUM: Income streams use deterministic inflation in Monte Carlo

**Files**: `calculator.rs:234-238`, `fire-math.ts:250-256`

In Monte Carlo, expenses use stochastic inflation
(`cumulativeInflation *= 1 + stochastic_draw`), but income streams use the
deterministic inflation rate via `additional_income_at_age()`. This creates an
asymmetry: in a high-inflation simulation, expenses spike but income doesn't
keep pace — even for inflation-indexed streams. The reverse is also wrong:
low-inflation simulations understate income.

This systematically **overstates risk** when inflation is above average and
**understates risk** when below average.

**Fix**: Pass `cumulativeInflation` into the income calculation for
inflation-indexed streams, or apply the stochastic inflation factor consistently
to both expenses and income.

#### HIGH: No input validation — division by zero

**Files**: `calculator.rs:68`, `fire-math.ts:62`, `fire.rs` (all commands)

`calculate_fire_target` divides by `safe_withdrawal_rate`. If the user sets SWR
to 0 (the slider allows 2.5% minimum, but the backend accepts any value via
IPC), this produces `Infinity` / `f64::INFINITY`, which propagates through all
calculations silently.

Similarly, negative ages, `current_age > target_fire_age`,
`planning_horizon_age < current_age`, and negative monetary values are not
validated.

**Recommendation**: Add validation at the Tauri command boundary:

```rust
if settings.safe_withdrawal_rate <= 0.0 {
    return Err("SWR must be positive".into());
}
if settings.current_age >= settings.planning_horizon_age {
    return Err("Planning horizon must exceed current age".into());
}
```

#### MEDIUM: No cap on n_sims — potential OOM

**Files**: `fire.rs:63`, `fire.rs:95`

`n_sims` is passed from the frontend. Default is 100,000 for MC and 5,000 for
strategy comparison. Each simulation allocates an in-memory path of ~55 `f64`
values inside the Tauri process. 100k sims = ~42 MB of transient heap usage —
fine. But the parameter is unbounded. A value of 10M would allocate ~4.2 GB of
simulation paths.

**Fix**: Cap at the Tauri command level:

```rust
let n = n_sims.unwrap_or(10_000).min(500_000);
```

#### MEDIUM: Constant-percentage "survival" is misleading

The constant-percentage withdrawal strategy (`SWR% * portfolio`) mathematically
never depletes the portfolio — it only approaches zero. The success rate will
therefore always be ~100%. But a portfolio of EUR 0.01 has "survived" in a
meaningless way.

**Recommendation**: Define failure as "portfolio drops below X% of initial"
(e.g., 10%) rather than just `> 0.0`. Or document the 100% rate as expected
behavior in the Guide.

#### MEDIUM: Withdrawal timing bias

**Files**: `calculator.rs:182`, `fire-math.ts:184`

The formula `portfolio = portfolio * (1 + r) - withdrawal` applies returns
**before** withdrawal (beginning-of-year returns). This slightly overestimates
portfolio longevity compared to end-of-year or mid-year models. This is a
standard simplification used in many FIRE calculators but should be documented.

#### HIGH: State model conflates financial independence with target-age retirement

**Files**: `calculator.rs:157-163`, `fire-math.ts:151-157`, `model.rs:75-81`

The FIRE trigger condition is
`portfolio >= nominalFireTarget || age >= targetFireAge`. The `||` means the
planner switches from accumulation to withdrawal at `target_fire_age` even if
the portfolio never reaches the FIRE target. The depletion math after the
trigger is internally consistent — the real problem is that the trigger is a
core model decision, not just a display choice:

1. It switches the simulation from accumulation to withdrawal phase
2. It sets `portfolio_at_fire`, which feeds SORR analysis
3. It sets `fire_age`, which feeds scenario analysis, sensitivity heatmaps, and
   MC median FIRE age
4. Sensitivity heatmap cells show `target_fire_age` instead of `None` for
   parameter combinations where FI is never reached — the heatmap loses its
   ability to distinguish viable from unviable plans
5. Scenario analysis shows the same `fire_age` for all three scenarios in
   underfunded cases, making the comparison meaningless
6. MC `median_fire_age` clusters at `target_fire_age` for underfunded paths,
   reporting a reassuring number while success rate is 30%

The current behavior models this policy: "retire as soon as FIRE target is
reached, otherwise retire at target age anyway." That is a valid policy, but the
outputs label the result as "FIRE age" — implying financial independence — not
"retirement age under a forced-at-target assumption."

**Fix**: Split the state explicitly in the domain model:

```rust
pub struct FireProjection {
    pub fi_age: Option<u32>,                    // age when portfolio >= target (None if never)
    pub retirement_age: u32,                     // actual age retirement starts
    pub retirement_trigger: RetirementTrigger,   // FiReached | TargetAge
    pub funded_at_retirement: bool,              // portfolio >= target at retirement
    // ...
}

pub enum RetirementTrigger {
    FiReached,
    TargetAge,
}
```

Then downstream consumers become coherent:

- Sensitivity/scenario tables show `fi_age` (None when unreachable)
- MC reports both: probability of FI by target age AND success rate if retiring
  at target age
- SORR labels correctly whether starting from FI or from underfunded retirement
- Dashboard warns when retirement is triggered by age, not by FI

#### HIGH: Selected-account scope is inconsistent across planner features

**Files**: `use-portfolio.ts`, `auto-config.ts`, `AllocationPage.tsx`

The planner exposes `includedAccountIds`, and `usePortfolioData()` correctly
uses that selection when computing `totalValue` via `getLatestValuations()`. But
the rest of the feature does not follow the same scope:

- `getHoldings("TOTAL")` pulls global holdings, not holdings for the selected
  FIRE accounts
- `getActivities()` pulls all activities, so contribution detection includes
  excluded accounts
- `deriveExpectedReturn()` uses all active accounts, not the selected FIRE
  accounts
- `AllocationPage` and drift monitoring therefore reflect the total portfolio,
  not the planner portfolio

**Impact**: A user can exclude an account from the FIRE plan but still have that
account influence auto-config, target allocations, drift alerts, and other
supporting visuals. That undermines trust in the planner because the main
projection and the supporting analysis are based on different account sets.

**Recommendation**: Thread selected account IDs through all holdings,
activities, and performance queries used by the planner. If that is not
possible in v1, remove the account-selection feature until the scope can be
applied consistently.

---

## 3. Algorithm Implementation Review

### 3.1 Random Number Generation

**Box-Muller transform** (hand-rolled): Correct implementation. Uses only the
cosine component, discarding the sine — wastes 50% of the random draws.

**Alternative**: The `rand_distr` crate provides `Normal` distribution:

```rust
use rand_distr::{Normal, Distribution};
let normal = Normal::new(mean, std).unwrap();
let sample = normal.sample(&mut rng);
```

This uses the ziggurat algorithm (faster) and avoids the transcendental function
calls. Already a transitive dependency via `rand`.

### 3.2 Two-Regime Fat-Tailed Distribution

The mixture model is novel and reasonable:

- 85% normal years: mean+1.5%, sigma\*0.8
- 15% stress years: mean-8.5%, sigma\*1.8
- Mean preservation: 0.85*(mean+0.015) + 0.15*(mean-0.085) = mean

**Variance is not preserved**: The mixture variance is higher than sigma^2 alone
(by design — this is the "fat tail" effect). The user-set sigma is the
per-regime base, not the portfolio's overall volatility. This should be
documented — a user who enters sigma=12% based on historical data will get a
wider fan chart than expected.

### 3.3 Stochastic Inflation

sigma=0.01 (1%) is hardcoded. Historical CPI volatility is roughly 1-2%, so this
is reasonable for most developed economies. Not configurable — acceptable for
v1.

### 3.4 Rayon Parallelism

Good use of `into_par_iter()` for embarrassingly parallel MC loop. RNG
initialized per task via `thread_rng()` — correct for rayon's work-stealing
model. No shared mutable state in the closure.

### 3.5 Non-reproducibility

Neither Rust nor TS implementation supports seed-based RNG. Two runs of the same
Monte Carlo will produce different results. This could confuse users who see
their success rate jump between 87% and 92% on re-runs.

**Recommendation**: Consider using `StdRng::from_seed()` with a deterministic
seed derived from the settings hash. This way, same inputs = same outputs. The
user can still "re-roll" by changing any parameter.

### 3.6 Alternative Crates

| Crate         | Purpose                    | Verdict                                   |
| ------------- | -------------------------- | ----------------------------------------- |
| `rand_distr`  | Normal distribution        | Recommended — better perf than Box-Muller |
| `statrs`      | Full statistical toolkit   | Overkill for this use case                |
| `ndarray`     | Matrix ops for sensitivity | Unnecessary — arrays of arrays work fine  |
| `rand_chacha` | Reproducible seeding       | Nice-to-have for reproducibility          |

None of the crate substitutions are mandatory. The statistical engine is
broadly reasonable and performant; the larger risks are in product semantics,
validation, and keeping the duplicated TS/Rust logic aligned.

---

## 4. Code Quality

### 4.1 Dual-Stack Duplication

`fire-math.ts` (552 lines) is a near-exact port of `calculator.rs` (709 lines).
Both implementations of every algorithm must stay in sync. The TS version runs
synchronously on the main thread; the Rust version runs MC/strategy comparison
in parallel.

**Risk**: A bug fix in one won't automatically land in the other. There are
already maintainability drifts:

1. `SimulationsPage` invokes Monte Carlo and strategy comparison through the
   Tauri adapter, but `fire-math.ts` still carries its own MC and
   strategy-comparison implementation. That leaves a second copy of stochastic
   logic in the tree with no clear product path.
2. The same conceptual bugs are duplicated in both copies, including the
   `portfolio >= nominalFireTarget || age >= targetFireAge` trigger and the MC
   inflation asymmetry. Any fix has to be made twice.
3. The TS `stepPensionFunds` returns a `total` value that callers ignore, while
   the Rust version is `void`. Small API drifts like this are a warning sign
   that the copies will continue to diverge.

**Recommendation**: Consider whether the TS computation library can be
eliminated entirely in favor of Tauri calls. The latency for deterministic
projections is <1ms — the IPC overhead is the only concern. Alternatively,
clearly mark the TS version as the source of truth for sync calculations and the
Rust version for async/parallel.

### 4.2 Desktop-Only Integration

The latest commit fixes the settings flow by adding
`fire-planner-settings-page.tsx` and `linkedGoalId`, but the feature is still
wired only through `apps/frontend/src/adapters/tauri/fire-planner.ts`. There
are no `apps/server/src/api/` endpoints or web adapter exports for the FIRE
planner, so the feature remains **desktop-only**. If that is intentional for
v1, the route should be explicitly gated or hidden on web builds rather than
half-integrated through shared frontend code.

### 4.3 Settings Schema Versioning

Settings are stored as a JSON blob via `settings_service.set_setting_value()`.
No schema version field. This is no longer hypothetical: the second commit adds
`linked_goal_id` to `FireSettings` without `#[serde(default)]`, so any existing
saved FIRE settings JSON that lacks that field will fail deserialization in
`get_fire_settings`. The `#[serde(default)]` attribute is only on
`withdrawal_strategy`.

**Fix**: Add `#[serde(default)]` to all optional/new fields, or add a version
field to the JSON envelope.

### 4.4 ID Generation

`generateId()` in `SettingsPage.tsx` uses
`Math.random().toString(36).slice(2,9)`. This produces 7 alphanumeric chars
(~36^7 ~ 78 billion combinations). Collision risk is low for a handful of income
streams, but `crypto.randomUUID()` is already used elsewhere in the PR
(`fire-planner-settings-page.tsx`) and is the better choice.

### 4.5 TypeScript Adapter Resolution Is Hardwired To Tauri

The PR adds hardcoded path aliases:

```json
"@/adapters": ["./src/adapters/tauri/index.ts"],
"@/adapters/*": ["./src/adapters/tauri/*"],
```

The repo already uses build-target-aware Vite aliases for `@/adapters`, but
this change hardwires TypeScript resolution to the Tauri implementation. That
means editor tooling, `tsc`, and any path-based tooling now disagree with the
existing dual-runtime pattern even if the Vite runtime alias still points
correctly at web vs Tauri builds.

### 4.6 SettingsPage Size

`SettingsPage.tsx` is 898 lines — the largest file in the PR. It mixes form
logic, auto-config, pension fund sync, allocation management, and UI. Consider
extracting the income stream editor and allocation editor into separate
components.

### 4.7 Missing Error Boundaries

Monte Carlo and strategy comparison run asynchronously via Tauri. The
`MonteCarloSection` component catches errors with try/finally but doesn't show
error messages — `setRunning(false)` silently resets state. If the Rust backend
panics or returns an error string, the user sees nothing.

### 4.8 Test Coverage

Rust: 9 unit tests covering target calculations, projection, MC range, SORR, and
sensitivity dimensions. This is a good baseline but misses:

- Edge cases: zero portfolio, zero contributions, age >= horizon
- Constant-percentage withdrawal strategy tests
- Pension fund accumulation verification
- Income stream growth calculations

Frontend: No tests.

### 4.9 Frontend File Naming Drift

The new FIRE page files under `apps/frontend/src/pages/fire-planner/pages/`
use PascalCase filenames (`AllocationPage.tsx`, `GuidePage.tsx`, etc.), while
the existing frontend codebase is overwhelmingly kebab-case
(`performance-page.tsx`, `goals-page.tsx`, `settings-layout.tsx`, etc.). This
is not a functional bug, but it introduces avoidable style drift in an
otherwise consistent area of the repo.

**Recommendation**: Rename the new page files to kebab-case to match the
existing frontend convention.

---

## 5. Security & Reliability

| Risk                                      | Severity | Notes                                                           |
| ----------------------------------------- | -------- | --------------------------------------------------------------- |
| No `n_sims` cap -> OOM                    | Medium   | 100k sims = 42MB; uncapped input from IPC                       |
| Division by zero (SWR=0)                  | Medium   | Produces Infinity, propagates silently                          |
| No input validation on Tauri commands     | Medium   | Accepts any JSON via IPC                                        |
| Settings schema incompatibility           | Medium   | `linked_goal_id` was added without `#[serde(default)]`; older saved JSON will fail to load |
| Mixed account scopes across planner UI    | Medium   | Main projection uses selected accounts; auto-config/allocation still use total portfolio |
| Settings stored as plain JSON             | Low      | Financial plan data, no credentials — acceptable                |
| `Math.random()` for IDs                   | Low      | Not cryptographic but fine for stream IDs                       |
| Rayon panic propagation                   | Low      | A panic in a MC closure would propagate as a Tauri error string |
| No rate limiting on computation endpoints | Low      | User can rapid-fire MC runs; desktop-only so acceptable         |

---

## 6. Product Strategy Assessment

### 6.1 Strategic Fit — Strong

Moving FIRE planning from addon to core is the right call. For a local-first
investment tracker, FIRE planning is a natural extension that:

- Increases user engagement (daily check of FIRE progress)
- Creates a "sticky" feature (users configure their plan and revisit it)
- Differentiates from simpler portfolio trackers
- Leverages existing portfolio data (auto-config from activities/holdings)

### 6.2 Competitive Position

Free web alternatives (FIRECalc, cFIREsim, ProjectiFi, FI Calc) exist but
require manual data entry. Wealthfolio's advantage is **integration with actual
portfolio data** — auto-detect contributions, derive expected returns from
historical performance, pull live pension fund balances. That integration angle
is the right differentiator, but the pension-fund coupling and account-scoping
issues in this PR still undermine trust in the numbers.

### 6.3 Italian Market Focus

The TFR/INPS/fondo pensione modeling and Italy-specific guide content suggest
the contributor's primary market. The timezone-based country detection is a
reasonable heuristic. The Italy content is gated behind the Guide tab and
doesn't affect calculations.

**Risk**: The Italian-specific pension modeling could fragment the feature.
Future contributors from other countries may add their own pension systems,
creating maintenance sprawl. Consider a more generic "defined benefit" /
"defined contribution" framework instead of country-specific modeling.

### 6.4 Missing Features for Production

These are not blockers for an initial release but should be on the roadmap:

| Feature                                          | Priority | Notes                                                          |
| ------------------------------------------------ | -------- | -------------------------------------------------------------- |
| Tax-aware projections                            | High     | Capital gains, pension tax, dividend tax vary by country       |
| Variable withdrawal strategies (guardrails, VPW) | Medium   | Only constant-dollar and constant-% are implemented            |
| Historical backtesting                           | Medium   | Test plan against actual market history, not just MC           |
| Spouse/partner combined planning                 | Medium   | Common FIRE scenario — two incomes, different retirement ages  |
| Healthcare cost modeling                         | Medium   | Separate expense bucket with higher inflation                  |
| Glide path / age-based allocation shift          | Low      | Fixed allocation throughout — unrealistic for 40+ year horizon |

### 6.5 Goal Integration — Good but Confusing

The settings page auto-creates a "FIRE" goal with `grossTarget`
(expenses\*12/SWR). But the FIRE planner dashboard prominently displays the
**net** target. Users will see two different numbers for their FIRE target:

- Goals page: EUR 900,000 (gross)
- FIRE planner: EUR 540,000 (net, with income streams)

This should be clearly explained in the UI or harmonized.

---

## 7. Summary of Recommendations

### Must Fix Before Merge

1. **Split FI attainment from retirement trigger in the domain model** — the
   planner conflates "financial independence reached" with "retirement starts at
   target age." Add `fi_age`, `retirement_trigger`, `funded_at_retirement` to
   `FireProjection` so sensitivity, scenario, MC, and SORR outputs are coherent
   for underfunded cases
2. **Make the pension model internally consistent** — either derive payout from
   the accumulated fund or clearly demote the balance display to
   informational-only so users do not mistake it for a funded income promise
3. **Apply selected account scope consistently** — the same included-account
   set must drive projections, auto-config, holdings, activities, and
   allocation/drift analysis
4. **Validate inputs at the Tauri command boundary and cap `n_sims`** — SWR >
   0, ages in valid range, no silent `Infinity`, no unbounded simulation size
5. **Remove the tsconfig hardwiring to Tauri** — keep TypeScript resolution
   aligned with the repo's existing build-target abstraction; if the feature is
   desktop-only for v1, isolate it explicitly

### Should Fix

6. Fix Monte Carlo inflation asymmetry — apply stochastic inflation
   consistently to both expenses and inflation-linked income streams
7. Add `#[serde(default)]` to new/optional `FireSettings` fields — the current
   `linked_goal_id` addition is already a backward-compatibility break
8. Use `rand_distr::Normal` instead of hand-rolled Box-Muller
9. Add error display in Monte Carlo UI when Rust returns an error
10. Replace `Math.random()` ID generation with `crypto.randomUUID()`
11. Rename new frontend page files to kebab-case for consistency with the
   existing `apps/frontend/src/pages` convention

### Nice to Have

12. Reproducible MC with deterministic seed
13. Web adapter for server mode
14. Extract SettingsPage sub-components (income streams, allocations)
15. Document the two-regime model's effect on user-entered volatility
16. Add property-based / edge-case tests for calculator

---

## 8. Validation Notes

- Rust verification: `cargo test -p wealthfolio-core fire` passed in an
  isolated PR worktree.
- Frontend type-checking was not a clean PR-only signal because the workspace
  expects referenced package outputs to be built first, so the frontend/runtime
  findings above were verified by source inspection against the latest PR head.

---

## Appendix: File-by-File Summary

| File                                               | Lines | Purpose                             | Notes                          |
| -------------------------------------------------- | ----- | ----------------------------------- | ------------------------------ |
| `crates/core/src/portfolio/fire/model.rs`          | 162   | Domain types                        | Clean, well-structured         |
| `crates/core/src/portfolio/fire/calculator.rs`     | 709   | All calculation algorithms + tests  | Core of the PR                 |
| `apps/tauri/src/commands/fire.rs`                  | 105   | 8 Tauri IPC commands                | Thin wrappers, no validation   |
| `apps/frontend/.../fire-math.ts`                   | 552   | TS port of calculator               | Duplicates Rust logic          |
| `apps/frontend/.../types.ts`                       | 120   | TS interfaces                       | Mirrors Rust models            |
| `apps/frontend/.../auto-config.ts`                 | 193   | Derive settings from portfolio      | Useful, but not scoped to selected FIRE accounts |
| `apps/frontend/.../SettingsPage.tsx`               | 898   | Full settings form                  | Oversized, needs decomposition |
| `apps/frontend/.../SimulationsPage.tsx`            | 889   | MC, scenarios, SORR, sensitivity    | Good data visualization        |
| `apps/frontend/.../DashboardPage.tsx`              | 496   | KPIs, progress, budget breakdown    | Well-designed dashboard        |
| `apps/frontend/.../AllocationPage.tsx`             | 317   | Drift monitor, pie charts           | Useful drift view, but currently total-portfolio scoped |
| `apps/frontend/.../GuidePage.tsx`                  | 217   | User guide with IT-specific content | Helpful documentation          |
| `apps/frontend/.../fire-planner-settings-page.tsx` | 112   | Settings + Goal integration wrapper | Goal upsert logic              |
| `apps/frontend/.../fire-planner-page.tsx`          | 100   | Main page with tab layout           | Clean routing                  |

afadil:
## Remaining Findings

### 1. HIGH: Pension fund balance is still disconnected from payout amount

**Files**: `crates/core/src/portfolio/fire/calculator.rs`,
`apps/frontend/src/pages/fire-planner/lib/fire-math.ts`,
`apps/frontend/src/pages/fire-planner/pages/dashboard-page.tsx`

The planner still models pension/retirement funds in two unrelated ways:

- an accumulating balance (`current_value`, `monthly_contribution`,
  `accumulation_return`) that grows over time and is shown in the projection
- a separate manual `monthly_amount` that suddenly becomes retirement income at
  `start_age`

At payout age, the accumulated balance is zeroed out and the manual income
stream takes over, but there is still no formula connecting one to the other.

That means the UI can show a pension fund growing to a large balance and then
silently replacing it with an arbitrary income amount that may be too high, too
low, or unrelated. This is still the biggest business-logic trust problem in
the PR.

**Recommendation**:

- either derive payout from the accumulated balance
- or explicitly label pension balances as informational only and state that the
  payout amount is a manual estimate

### 2. HIGH: Selected-account scope is still inconsistent, and goal sync widens the mismatch

**Files**: `apps/frontend/src/pages/fire-planner/hooks/use-portfolio.ts`,
`apps/frontend/src/pages/fire-planner/lib/auto-config.ts`,
`apps/frontend/src/pages/fire-planner/pages/allocation-page.tsx`,
`apps/frontend/src/pages/settings/fire-planner/fire-planner-settings-page.tsx`

The planner still does not use one consistent account universe:

- `totalValue` respects `includedAccountIds` via `getLatestValuations(activeAccountIds)`
- holdings still come from `getHoldings("TOTAL")`
- activities still come from `getActivities()`
- auto-config still derives expected return from all active accounts
- allocation/drift still uses total holdings and total activities

The latest commit also keeps automatically adding `linkedAccountId` values from
income streams into goal allocations. That means the linked goal can now reflect
accounts that the planner projection itself does not include in `totalValue`.

So the user can end up with:

- one portfolio scope for the projection
- a different scope for allocation/drift and auto-config
- a third scope for goal progress

This remains a major trust issue because supporting analytics are still not
based on the same portfolio the main planner uses.

**Recommendation**:

- thread one explicit account set through projection, holdings, activities,
  auto-config, allocation, and goal sync
- if that is not possible in v1, remove account selection or clearly narrow the
  feature scope

### 3. MEDIUM: Web integration is now build-safe, but still only stubbed

**Files**: `apps/frontend/src/adapters/web/fire-planner.ts`,
`apps/frontend/src/pages/fire-planner/hooks/use-fire-settings.ts`

The new web adapter fixes the previous missing-module problem, but it is still a
desktop-only stub that throws `"FIRE planner requires the desktop app"` for all
operations.

That is acceptable only if the feature is explicitly desktop-only for v1 and is
gated accordingly. Otherwise the app now has a feature that can render in web
mode, default settings successfully, and then fail on save/simulate at runtime.

**Recommendation**:

- either hide/gate the feature on web builds
- or add the missing server API and real web adapter implementation

### 4. MEDIUM: One downstream consumer still confuses FI with retirement start

**File**: `apps/frontend/src/pages/fire-planner/pages/simulations-page.tsx`

The core model now distinguishes genuine FI from target-age retirement much
better, but `SorrSection` still checks:

```ts
const fireReached = proj.portfolioAtFire > 0;
```

That is the wrong predicate. `portfolioAtFire` is set whenever the withdrawal
phase starts, including the "forced by target age" path. For an underfunded
plan with any non-zero portfolio, this branch still treats retirement start as
if FIRE were reached.

Impact:

- the yellow "illustrative only" warning is suppressed for underfunded plans
- the copy says "FIRE has not been reached" only in cases where
  `portfolioAtFire === 0`, which is not the real condition

This should use `proj.fundedAtRetirement` or `proj.fireAge !== null`.

Related semantic residue remains in the UI too: some sections still fall back to
labeling `targetFireAge` as "FIRE age" when FI is not actually reached.

### 5. MEDIUM: TypeScript adapter resolution still depends on Tauri hardwiring

**Files**: `apps/frontend/tsconfig.json`,
`apps/frontend/src/adapters/web/fire-planner.ts`

`tsconfig.json` still maps:

```json
"@/adapters/*": ["./src/adapters/tauri/*"]
```

That means TypeScript tooling still resolves direct adapter imports to the Tauri
implementation even for code that conceptually belongs to the web side.

The new web stub now depends on that behavior:

```ts
import type { SensitivityResult, StrategyComparisonResult } from "@/adapters/fire-planner";
```

Under the actual Vite web alias, that import points back to the same web file,
not to a shared type source. It works today only because the `tsconfig` path
forces type resolution through the Tauri file instead.

So the compile failure is fixed, but the adapter layering is still brittle and
tooling-dependent.

**Target fix**:

- export the FIRE planner API through the existing platform entrypoints:
  `src/adapters/tauri/index.ts` and `src/adapters/web/index.ts`
- move shared FIRE adapter result types such as `SensitivityResult` and
  `StrategyComparisonResult` into a platform-neutral module, preferably
  `apps/frontend/src/pages/fire-planner/types.ts` or a new shared adapter-types
  file
- update feature code to import from `@/adapters`, not `@/adapters/fire-planner`
- remove the `@/adapters/* -> ./src/adapters/tauri/*` override from
  `apps/frontend/tsconfig.json`

That would bring the feature back in line with the repo's normal adapter model:
Vite chooses the platform implementation at build time, while TypeScript sees
only platform-neutral exports and types.

### 6. MEDIUM: Dual TS/Rust implementation drift is still a maintainability risk

**Files**: `crates/core/src/portfolio/fire/calculator.rs`,
`apps/frontend/src/pages/fire-planner/lib/fire-math.ts`

The latest commit kept both implementations aligned, which is good. But the
structural duplication is still large:

- deterministic projection, SORR, scenario analysis, and sensitivity exist in
  both Rust and TypeScript
- the frontend currently calls backend Monte Carlo/strategy comparison, while
  `fire-math.ts` still carries its own local Monte Carlo and strategy-comparison
  code

So the current tree still contains duplicate business logic with no single clear
source of truth. That is not an immediate merge blocker, but it will keep
raising maintenance cost.

### 7. LOW: Goal semantics are still a little confusing

**File**: `apps/frontend/src/pages/settings/fire-planner/fire-planner-settings-page.tsx`

The goal created from settings still uses the **gross** FIRE target, while the
planner dashboard emphasizes the **net** FIRE target. The dashboard tooltip now
explains the distinction, which helps, so this is no longer a major issue. But
combined with linked pension accounts being added into goal allocations, the
goal surface can still diverge from the planner surface in ways users will find
hard to reason about.

### 8. LOW: Prefer the shared `generateId()` helper for frontend consistency

**File**: `apps/frontend/src/pages/fire-planner/pages/settings-page.tsx`

The PR already fixed the earlier weak ID generation by moving away from
`Math.random()`. That part is good. But the repo already has a shared frontend
ID helper in `apps/frontend/src/lib/id.ts`, and this form now calls
`crypto.randomUUID()` directly instead.

This is not a correctness or security issue here. The IDs are local planner
stream IDs, and `crypto.randomUUID()` is perfectly valid. I would still prefer
using the shared `generateId()` helper for consistency with the rest of the
frontend and to keep the fallback behavior in one place.

---

## Algorithm & Crate Assessment

### Correctness

The core FIRE formulas remain sound:

- gross FIRE target: `expenses * 12 / SWR`
- net FIRE target: subtract only income available from day 1 of retirement
- coast FIRE: nominal target discounted back at nominal return
- deterministic yearly stepping: internally consistent

The latest commit also fixes the earlier "fake FIRE age" problem in the main
calculation outputs. That is a real improvement to the algorithmic integrity of
the feature.

### `rayon`

`rayon` is still a defensible choice here. This workload is CPU-bound and
embarrassingly parallel, so `into_par_iter()` is a better fit than trying to
force the loop through Tokio-style async patterns.

### `rand_distr`

Switching to `rand_distr::Normal` is the right improvement. It is simpler and a
better fit than hand-rolling Gaussian draws.

### Remaining modeling limitation

The main algorithmic weakness is no longer the Monte Carlo engine. It is the
product model around pension accumulation vs payout and the inconsistent account
scope feeding the planner.

---

## Code Quality & Reliability

- The Tauri command layer is in much better shape now: validation exists, and
  simulation size is capped.
- The settings schema backward-compatibility break from `linked_goal_id` was
  fixed.
- The frontend naming and UUID issues are fixed.
- The settings form is still large and would benefit from decomposition, but
  that is not urgent.
- Frontend test coverage is still absent.
- Monte Carlo remains non-reproducible across runs because no deterministic seed
  is used. That is acceptable for v1, but worth documenting.

alexghirelli:
Hey @afadil some points to keep in mind for a V2 after the merge

## Post-merge feature ideas

  A few things worth considering once this lands. Grouped roughly by effort/impact.

  ---

  ### High priority

  **One-off expenses**
  Model future lump-sum costs tied to a specific age (home repairs, college tuition, car purchase). These reduce the portfolio
  in a single year rather than recurring monthly. Most FIRE planners handle this on a spreadsheet today — integrating it would
  be a meaningful differentiator. Implementation: add an `oneOffExpenses: { age, amount }[]` array to `FireSettings` and
  subtract from the portfolio in the relevant projection year.

  **Retirement age comparison view**
  Instead of changing `targetFireAge` in settings and re-running, show "retire at 50 / 55 / 60" side by side in a single chart.
   Similar to the existing scenario analysis but varying age rather than return. Low math complexity, high UX value.

  **Withdrawal order**
  Which accounts to draw from first (taxable → tax-deferred → tax-free). Currently the model treats the portfolio as a single
  bucket. Withdrawal order materially affects after-tax outcomes and is one of the most common questions in FIRE communities.

  ---

  ### Medium priority

  **Withdrawal taxes**
  The projections are currently pre-tax. Adding a simplified tax layer (flat effective rate, or per-country capital gains rate)
   would make the numbers more realistic, especially for non-US users where the imposta sostitutiva / CGT rate is fixed and
  predictable.

  **Dynamic withdrawal strategies**
  Guyton-Klinger or Vanguard Dynamic Spending as a third withdrawal strategy alongside constant-dollar and constant-percentage.
   Both reduce sequence-of-returns risk without fully sacrificing spending floor — useful for the SORR section.

  **"When can I stop contributing?"**
  Given the current portfolio, compute the age at which coast FIRE is reached with zero future contributions. Single derived
  value, no new settings needed.

  ---

  ### Nice to have

  - **PDF export** — one-page plan summary
  - **Annual review mode** — compare current projections against last year's snapshot
  - **Debt/mortgage payoff** — model a mortgage that expires at age Y, freeing up X/month from that point

afadil:
Thanks for the latest update. Some remaining issues woth fixing before merge

### 1. Early-FI paths still use a target/income model anchored to `target_fire_age`

This is the biggest remaining logic issue.

Right now the planner can retire **before** `target_fire_age`, but parts of the target calculation still assume retirement income is available by `target_fire_age`, not by the **actual** retirement start:

- `calculate_net_fire_target()` subtracts streams with `start_age <= target_fire_age`
- `resolve_dc_payouts()` assumes DC contributions continue until `min(start_age, target_fire_age)`

That can overstate readiness in early-FI cases. Example: if the portfolio crosses the reduced target at 50, but the pension/DC income only starts at 55, the planner can still report FI too early because the target was reduced using income that is not actually available at 50.

I think this needs one consistent model:
- either this is a **target-age retirement planner**, and withdrawal never starts before `target_fire_age`
- or this is a true **FIRE planner**, and target/income availability must be computed from the actual candidate retirement age

### 2. Linked DC accounts can still be double counted

The latest commit correctly stopped widening goal allocations with linked stream accounts, but the planner still allows this:

- include an account in `includedAccountIds`
- also link that same account to a DC stream
- count it once as portfolio capital and again as future income derived from that capital

That can materially overstate retirement readiness.

I think the safest fix is to make those scopes mutually exclusive:
- either automatically exclude linked DC accounts from the FIRE portfolio
- or block save with a validation error when the same account appears in both places

### 3. SORR still starts from `target_fire_age`, not the actual retirement start

`run_sequence_of_returns_risk()` still derives:
- retirement duration from `planning_horizon_age - target_fire_age`
- age progression from `target_fire_age + i`

That means if FI is reached earlier, the SORR analysis is still using the wrong starting age and wrong retirement window.

It should take `retirementStartAge` explicitly instead of reconstructing the path from `target_fire_age`.

### 4. The desktop-only settings route is still exposed in web mode

The main planner page is gated, but `/settings/fire-planner` is still routed and visible in navigation/settings on web. Since the web adapter still throws desktop-only errors, the settings page looks usable until the user tries to save.

This should either:
- be hidden in web mode
- or use the same desktop-only gate as the main planner page

## Lower-priority follow-ups

- if the selected FIRE account list becomes empty, goal allocations should be cleared rather than leaving stale rows behind
- the new DC/early-FI behavior still needs direct test coverage; the current FIRE tests pass, but they do not cover the new timing-sensitive cases

## Bottom line

This PR is close, and the latest commit fixed several real problems. I would still ask for one more round before merge because the remaining issues affect the planner’s meaning, not just presentation. The key thing to resolve is the distinction between `target_fire_age` and the actual retirement/FI start, and then make the target, DC payout logic, and SORR all follow that same model.

PR #783, opened 2026-03-27
feat(market-data): add historical quotes to Metal Price API provider

triantos:
## Summary
- Implement `get_historical_quotes()` for the Metal Price API provider using the `/v1/timeframe` endpoint
- Set `supports_historical: true` so the quote sync engine can backfill metal prices automatically
- Previously, metals with MARKET quote mode were skipped during sync because the provider reported no historical support, leaving prices stale indefinitely
- Add integration tests gated behind `METAL_PRICE_API_KEY` env var

## Test plan
- [ ] Unit tests pass: `cargo test -p wealthfolio-market-data` (232 passed)
- [ ] Integration tests pass: `METAL_PRICE_API_KEY=<key> cargo test -p wealthfolio-market-data -- --ignored metal_price` (4 passed)
- [ ] Full workspace: `cargo test --workspace` (1,282 passed, 0 failed)
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all -- --check` clean

### DIFF
--- a/crates/market-data/src/provider/metal_price_api/mod.rs
+++ b/crates/market-data/src/provider/metal_price_api/mod.rs
@@ -21,6 +21,8 @@ use serde::Deserialize;
 use std::collections::HashMap;
 use std::time::Duration;
 
+use tracing::warn;
+
 use crate::errors::MarketDataError;
 use crate::models::{Coverage, InstrumentKind, ProviderInstrument, Quote, QuoteContext};
 use crate::provider::{MarketDataProvider, ProviderCapabilities, RateLimit};
@@ -91,16 +93,19 @@ impl MetalPriceApiProvider {
     }
 
     /// Convert a rate (1 base_currency = rate troy ounces) to price per troy ounce.
+    /// Division is done in Decimal space to avoid f64 precision loss.
     fn rate_to_price(rate: f64) -> Result<Decimal, MarketDataError> {
-        if rate == 0.0 {
+        let rate_decimal =
+            Decimal::try_from(rate).map_err(|_| MarketDataError::ValidationFailed {
+                message: "Failed to convert rate to decimal".to_string(),
+            })?;
+        if rate_decimal.is_zero() {
             return Err(MarketDataError::ProviderError {
                 provider: PROVIDER_ID.to_string(),
                 message: "Invalid rate (zero)".to_string(),
             });
         }
-        Decimal::try_from(1.0 / rate).map_err(|_| MarketDataError::ValidationFailed {
-            message: "Failed to convert rate to decimal".to_string(),
-        })
+        Ok(Decimal::ONE / rate_decimal)
     }
 }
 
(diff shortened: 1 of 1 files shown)


afadil:
### Issues

#### 1. Incorrect comment about free tier date range limit (clarification needed)

The test has this comment:

```rust
// Free tier limits timeframe to 5 days
let end = Utc::now();
let start = end - chrono::Duration::days(4);
```

The [[official API docs](https://metalpriceapi.com/documentation)](https://metalpriceapi.com/documentation) state the timeframe endpoint has a **365-day maximum range** across all plans. There's no mention of a 5-day restriction for the free tier.

Can you confirm where the 5-day limit comes from? Did you hit an error with a wider range on the free tier? If this is a quota/request-count issue rather than a timeframe restriction, the comment should be updated to avoid misleading future contributors.

#### 2. Prefer header-based authentication over query parameter

The API supports both:
- Query param: `?api_key=KEY`
- Header: `X-API-KEY: KEY`

The current code (both existing `get_latest_quote` and new `get_historical_quotes`) embeds the key in the URL. API keys in URLs can leak via server logs, browser history, and `Referer` headers.

**Suggestion:** Switch to header auth. This is pre-existing and could be a follow-up PR:

```rust
self.client
    .get(&url)
    .header("X-API-KEY", &self.api_key)
    .send()
    .await
```

#### 3. `f64` division before `Decimal` conversion

`rate_to_price` does `1.0 / rate` in `f64` space, then converts to `Decimal`. This introduces floating-point error before the conversion. The division could be done in `Decimal` space:

```rust
Decimal::ONE / Decimal::try_from(rate)?
```

Not a regression — the existing code did the same thing. Low priority.

#### 4. `success` field unchecked on `MetalPriceResponse` (pre-existing)

`get_historical_quotes` correctly checks `tf_resp.success`. The existing `get_latest_quote` does **not** check `response.success` on `MetalPriceResponse`. Not introduced by this PR but worth fixing alongside.

#### 5. No guard on date range exceeding API max

If the sync engine passes a range wider than 365 days, the API will reject it. Consider either:
- Adding a doc comment on `get_historical_quotes` noting the 365-day limit
- Or clamping/chunking the range

Low priority if the sync engine already handles provider errors gracefully.

triantos:
Thanks for the thorough review! Pushed fixes in 2624a05a:

**1. 5-day comment**: The free tier returned `{"success":false,"error":{"statusCode":421,"message":"Timeframe queries exceeding 5 days require a paid plan."}}` during integration testing. The official docs show 365-day max but that appears to be for paid plans. I've kept the test at 4 days to stay within the free tier limit but will update the comment to reference the actual error.

**2. Header auth**: Switched both `get_latest_quote` and `get_historical_quotes` to `X-API-KEY` header.

**3. Decimal division**: `rate_to_price` now converts to `Decimal` first, then divides.

**4. `success` check on latest**: Good catch — the existing code already checked `metal_resp.success` (line 191), but I added `tracing::warn` for better observability on both endpoints.

**5. 365-day range guard**: Will add a doc comment noting the limit. The sync engine handles provider errors gracefully (retries, then moves on), so exceeding the range won't break anything.

PR #792, opened 2026-03-29
feat: materialized lots table for holdings tracking

triantos:
## Summary

- Adds a `lots` table that tracks individual tax lots (acquisitions) as first-class database rows, replacing the JSON-serialized snapshot approach for position tracking
- Lots are computed from activities using FIFO disposal, written alongside existing snapshots for validation, then used as the authoritative source for valuations, net worth, and holdings
- Supports TRANSACTIONS-mode accounts (lots derived from activity replay) and HOLDINGS-mode accounts (lots derived from manual snapshots)

## Motivation

The existing holdings system stores positions as serialized JSON inside snapshot blobs. This makes it impossible to query individual lots for tax reporting, forces full-history replay on every quote update, and prevents incremental position updates. The lots table solves all three: lots are queryable rows, valuations become `lots JOIN quotes`, and activity changes update only affected lots.

## Changes

1. **Schema & model** — `lots` table with migration, `LotRecord` Rust model, `LotRepositoryTrait`
2. **Shadow-write** — after each recalc, lot rows are written in parallel with JSON snapshots; quantity mismatches logged at ERROR
3. **Wiring & backfill** — lot repository injected into services; on first launch, existing accounts are backfilled from current snapshots
4. **Tests** — unit, integration, and service tests for lot computation, FIFO closing, splits, transfers
5. **Read switchover** — `get_net_worth`, `calculate_valuation_history`, `holdings_from_snapshot`, and `build_live_holdings` all read from the lots table instead of deserializing JSON snapshots
6. **Incremental maintenance** — activity changes trigger targeted lot updates (insert/close) rather than full-account replay
7. **HOLDINGS-mode** — manual snapshot imports create lot records so both tracking modes use the same valuation path
8. **TOTAL pseudo-account fix** — TOTAL no longer gets lot rows written; its values come from query-time aggregation across real accounts

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace -- -D warnings` passes (core, server, ai, storage)
- [x] `cargo test --workspace` — 1,316 tests pass, 0 failures
- [x] `pnpm format:check` passes
- [x] Verified on live portfolio data (multi-currency, options, bonds, metals, liabilities)


afadil:
## Blocking issues

1. [P0] Startup lot backfill is unsafe/incomplete for existing HOLDINGS accounts  
   Files: `apps/server/src/scheduler.rs`, `apps/tauri/src/scheduler.rs`, `crates/core/src/portfolio/snapshot/snapshot_service.rs:268-276`, `crates/core/src/portfolio/snapshot/snapshot_service.rs:432-470`

   Both startup backfills call `recalculate_holdings_snapshots(None, SnapshotRecalcMode::Full)` when the lots table is empty.

   There are two bad outcomes here:
   - For manual-only/HOLDINGS-only users, `Full` recalc with `all_activities.is_empty()` deletes snapshots for every account and returns. On first launch after this migration, those users have an empty `lots` table and no activities, so the backfill can erase their saved manual snapshot history instead of populating lots from it.
   - For mixed portfolios, the full recalc explicitly skips HOLDINGS-mode accounts, so only transaction accounts get lots. After that, the `lots` table is no longer empty, so future startup backfills stop running. Existing HOLDINGS accounts remain permanently unbackfilled unless each snapshot is manually re-saved.

2. [P1] Historical as-of queries use the latest lot state, not the state at the requested date  
   Files: `crates/storage-sqlite/src/lots.rs:168-188`, `crates/storage-sqlite/src/lots.rs:211-256`  
   Affected readers: `crates/core/src/portfolio/holdings/holdings_service.rs`, `crates/core/src/portfolio/net_worth/net_worth_service.rs`, `crates/core/src/portfolio/valuation/valuation_service.rs`

   `get_lots_as_of_date` reconstructs history using only `open_date` and `close_date`, but `sync_lots_for_account` mutates each row in place to the latest `remaining_quantity` and `total_cost_basis`, and fully closed lots are zeroed out.

   Example: buy 10 shares on Jan 1, sell 4 on Feb 1. Querying Jan 15 should still show 10 shares, but after the sell the stored row now has `remaining_quantity = 6`, so the historical query returns 6 instead of 10. After a full close, dates before the close can return 0. This makes the new lots-backed historical holdings/net-worth/valuation reads incorrect.

3. [P1] HOLDINGS-mode history keeps only the latest snapshot in the lots table  
   File: `crates/core/src/portfolio/snapshot/snapshot_service.rs:1774-1820`

   Manual/CSV/broker snapshots only write lot rows when the saved snapshot is the latest snapshot for the account, and they do so via `replace_lots_for_account`, which replaces the account’s entire lot set.

   Because historical holdings/net worth now read security positions from the lots table, older HOLDINGS snapshots stop being representable as soon as a newer snapshot is saved. The same problem appears after deleting the latest snapshot: the lots table is refreshed from the new latest snapshot, so earlier snapshot dates lose their original security positions and instead show the latest remaining lot set (or zero).

4. [P1] Incremental lot sync cannot remove deleted lots or reopen previously closed lots  
   Files: `crates/storage-sqlite/src/lots.rs:211-256`, `crates/core/src/portfolio/snapshot/snapshot_service.rs:392-405`

   `sync_lots_for_account` only:
   - upserts the currently open lots
   - marks the listed closures as closed

   It never removes open rows that disappeared from the recalculated state, and its `ON CONFLICT DO UPDATE` path does not reset `is_closed`, `close_date`, or `close_activity_id`.

   That causes two correctness problems during `SinceDate` / incremental recalculation:
   - If a buy activity is deleted and a lot should disappear entirely, the stale open row is left behind.
   - If a closing sell is deleted/edited and a previously closed lot should become open again, the upsert updates quantity/cost basis but leaves `is_closed = 1`, so the lot remains invisible to open-lot queries.

## Additional issues

5. [P2] Web `get_snapshot_by_date` no longer matches desktop / previous behavior  
   Files: `apps/server/src/api/holdings/handlers.rs:186-199`, `apps/tauri/src/commands/portfolio.rs:925-947`

   The previous web behavior returned an error when no snapshot existed for the requested date. The new web handler now returns `200` with an empty holdings list, while the Tauri command still returns `"No snapshot found for date ..."`.

   This is a silent API contract change and also creates inconsistent web vs desktop behavior for the same frontend call.

6. [P3] `count_open_lots` counts all rows, not open rows  
   File: `crates/storage-sqlite/src/lots.rs:269-278`

   The method name says “open lots”, but the implementation does `count_star()` over the entire table with no `is_closed = 0` filter.

   Current callers only use it as “is the table empty?”, so this is not immediately breaking, but the name is misleading and will cause mistakes later.

7. [P3] Decimal parse failures are silently converted to zero in lots-backed readers  
   Files: `crates/core/src/portfolio/holdings/holdings_service.rs:672-681`, `crates/core/src/portfolio/net_worth/net_worth_service.rs:297-301`, `crates/core/src/portfolio/valuation/valuation_service.rs:326-327`

   Multiple readers parse lot numeric fields with `unwrap_or_default()`. If the DB ever contains malformed decimal text, the app will silently undervalue positions instead of surfacing a corruption/error signal.

   For lots, parse failure should be treated as a data integrity problem and at minimum logged loudly rather than quietly coerced to zero.

triantos:
I've addressed all of the issues you raised. Note there were some corner case bugs in previous handling of lots which are now fixed.

triantos:
Superseded by #856 — the three open data-model PRs (#792, #831, #842) have been consolidated into a single branch (`refactor_data_model`) with a single migration and a clean commit layout for easier review.

PR #801, opened 2026-04-01
Add possibility to skip de-duplication

### LINKED ISSUE #779: Duplicate activities with the same date and amount cannot both be imported
# Duplicate FEE activities with the same date and amount cannot both be imported

## Description

When importing activities, Wealthfolio rejects a second `FEE` activity if it has the same date and value as an existing activity. This is a problem for trade-level fees, where multiple trades on the same day can legitimately generate the same fee amount. [file:1]

The attached CSV contains three records: two identical `FEE` rows dated `10/04/2025` for `$CASH-GBP` with a fee of `7.5`, plus one separate `FEE` row dated `23/01/2026` with a fee of `63.27`.

The `63.27` fee is parsed correctly, but only one of the two `7.5` fee rows is accepted. [file:1]

## Steps to reproduce

1. Import the attached CSV into Wealthfolio.
2. Use the file containing the two identical `10/04/2025` `FEE` rows and the separate `23/01/2026` `FEE` row. [file:1]
3. Observe that the second identical `7.5` fee fails with the following error:

```text
Error: Activity error: Invalid data: Duplicate activity added. A matching activity already exists (id: [GUID]).
```

## Expected behavior

Both `7.5` fee rows should be imported as separate activities when they represent distinct trade-level fees reported on the same day.

## Actual behavior

Only one of the duplicate `7.5` fee rows is imported, and the other is rejected as a duplicate. This occurs both in the program and via CSV import. 

## Possible cause

The duplicate-detection logic for `FEE` activities likely treats same-day, same-amount fee entries as duplicates, even when they are legitimate separate records.

## Possible workaround / fix

One possible workaround would be to specify the fee `Symbol` to associate the fee with the symbol responsible for generating it. Another option would be to include an additional discriminator or unique identifier for fee entries so duplicate detection does not rely only on date and amount.

[fee_duplicate.csv](https://github.com/user-attachments/files/26272987/fee_duplicate.csv)

leppa:
## Description

Import logic automatically de-duplicates the imported data. Moreover, the logic doesn't just de-duplicate with the existing records, but also *within* the imported ones. The end user has no control over de-duplication which may, in the case of false positives, lead to legitimate records being skipped from the import.

Add a checkmark at the last import step that will disable the de-duplication logic for this specific import run.

image

As the key that is used to detect duplicates is also a unique key in the database, duplicate records will have `\n\n-- Possible duplicate <number>` appended to the end of their notes. This will effectively make them non-duplicates.

E.g., if the record has the following notes:
```
Buy 100 AAPL @250
```
it will become:
```
Buy 100 BSTG @250

-- Possible duplicate 1
```

Resolves #779

## Checklist

### DIFF
--- a/apps/tauri/src/commands/fire.rs
+++ b/apps/tauri/src/commands/fire.rs
@@ -4,12 +4,12 @@ use crate::context::ServiceContext;
 use log::debug;
 use tauri::State;
 use wealthfolio_core::portfolio::fire::{
-    FireProjection, FireSettings, MonteCarloResult, ScenarioResult, SensitivityResult,
-    SorrScenario, StrategyComparisonResult,
-};
-use wealthfolio_core::portfolio::fire::{
     project_fire_date, run_monte_carlo, run_scenario_analysis, run_sensitivity_analysis,
     run_sequence_of_returns_risk, run_strategy_comparison,
+};
+use wealthfolio_core::portfolio::fire::{
+    FireProjection, FireSettings, MonteCarloResult, ScenarioResult, SensitivityResult,
+    SorrScenario, StrategyComparisonResult,
 };
 
 const FIRE_SETTINGS_KEY: &str = "fire_planner_settings";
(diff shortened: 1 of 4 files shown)


afadil:
implemented in https://github.com/afadil/wealthfolio/pull/803

PR #822, opened 2026-04-07
fix: preserve bond price precision (4 decimal places)

### LINKED ISSUE #821: fix: bond price precision loss — 0.9192 displayed/saved as 0.92
## Summary

Bonds quoted as a percentage of face value (e.g. 91.92%) are normalized to a decimal fraction by the Boerse Frankfurt provider (91.92 → 0.9192). However, two issues cause precision loss:

### 1. Frontend rounds to 2 decimal places (primary bug)

`quote-history-data-grid.tsx` and `formatAmount()` both cap display at 2 decimal places, so `0.9192` is shown and can be saved back to the DB as `0.92` — losing 2 significant digits.

### 2. Boerse Frankfurt provider divides in f64 space (secondary)

The `/100` normalization for `traded_in_percent` bonds was performed in `f64` floating-point before converting to `Decimal`, which is inconsistent with the pattern established in `metal_price_api` and can introduce rounding errors for certain values.

## Impact

- Bond prices stored with wrong precision after manual quote editing
- Bond market value, P&L, and performance calculations are affected
- Affects all bonds fetched from Boerse Frankfurt with `tradedInPercent: true` (Italian BTPs, German Bunds, French OATs, etc.)

## Related

- Part of the broader bond support tracked in #753

marcoscale98:
Fixes #821

- boerse_frankfurt: convert f64 to Decimal before dividing by 100, matching the pattern in metal_price_api (Decimal-space arithmetic)
- quote-history-data-grid: increase default precision from 2 to 4 so bond prices like 0.9192 are not rounded to 0.92 on save
- formatAmount: set maximumFractionDigits to 4 (min stays 2) so trailing zeros are trimmed automatically via Intl.NumberFormat

## Summary

Bonds from Boerse Frankfurt are quoted as % of face value (e.g. 91.92%), normalized to a decimal fraction (0.9192). Two issues caused precision loss:

- **Frontend rounded to 2 decimals**: `0.9192` was displayed and could be saved back as `0.92`, destroying precision
- **f64-space division in provider**: the `/100` normalization happened before `Decimal` conversion, inconsistent with the pattern in `metal_price_api`

## Changes

**`crates/market-data/src/provider/boerse_frankfurt/mod.rs`**
- Convert `f64` → `Decimal` first, then divide by `Decimal::from(100)` for `traded_in_percent` bonds (latest + historical quotes)

**`apps/frontend/src/pages/asset/quote-history-data-grid.tsx`**
- Default precision: `2` → `4` so bond prices like `0.9192` are preserved on display and save

**`packages/ui/src/lib/utils.ts`**
- `maximumFractionDigits`: `2` → `4` in `formatAmount()` — propagates to all price displays. `Intl.NumberFormat` trims trailing zeros automatically (`0.9200` → `0.92`, `0.9192` → `0.9192`)

## Test plan

- [x] `cargo test -p wealthfolio-market-data` — 232 tests pass
- [x] `pnpm type-check` — no errors
- [x] Manual: bond asset from Boerse Frankfurt shows 4 decimal places
- [x] Manual: stocks/ETFs still show 2 decimals

## Checklist

### DIFF
--- a/apps/frontend/src/lib/format-amount.test.ts
+++ b/apps/frontend/src/lib/format-amount.test.ts
@@ -1,0 +1,28 @@
+import { formatAmount } from "@wealthfolio/ui";
+import { describe, expect, it } from "vitest";
+
+describe("formatAmount", () => {
+  it("formats integer with default 2 decimals", () => {
+    expect(formatAmount(1234, "USD")).toBe("$1,234.00");
+  });
+
+  it("rounds to 2 decimals by default", () => {
+    expect(formatAmount(1234.5678, "USD")).toBe("$1,234.57");
+  });
+
+  it("shows exactly 4 decimals when precision=4", () => {
+    expect(formatAmount(0.9192, "USD", false, 4)).toBe("0.9192");
+  });
+
+  it("pads trailing zeros when precision=4 (columns align)", () => {
+    expect(formatAmount(0.92, "USD", false, 4)).toBe("0.9200");
+  });
+
+  it("pads with zeros for a 3-decimal value at precision=4", () => {
+    expect(formatAmount(0.918, "USD", false, 4)).toBe("0.9180");
+  });
+
+  it("returns - for null", () => {
+    expect(formatAmount(null, "USD")).toBe("-");
+  });
+});
--- a/apps/frontend/src/lib/utils.ts
+++ b/apps/frontend/src/lib/utils.ts
@@ -253,6 +253,31 @@ const DECIMAL_FORMAT_OPTIONS: Intl.NumberFormatOptions = {
 
 const decimalFormatter = new Intl.NumberFormat("en-US", DECIMAL_FORMAT_OPTIONS);
 const currencyFormatterCache = new Map<string, Intl.NumberFormat>();
+const precisionFormatterCache = new Map<string, Intl.NumberFormat>();
+
+const getPrecisionFormatter = (
+  precision: number,
+  currency?: string,
+  displayCurrency?: boolean,
+): Intl.NumberFormat => {
+  const key = `${currency ?? ""}|${displayCurrency}|${precision}`;
+  if (precisionFormatterCache.has(key)) return precisionFormatterCache.get(key)!;
+  const opts: Intl.NumberFormatOptions = {
+    minimumFractionDigits: precision,
+    maximumFractionDigits: precision,
+  };
+  let formatter: Intl.NumberFormat;
+  try {
+    formatter =
+      displayCurrency && currency
+        ? new Intl.NumberFormat("en-US", { style: "currency", currency, ...opts })
+        : new Intl.NumberFormat("en-US", opts);
+  } catch {
+    formatter = new Intl.NumberFormat("en-US", opts);
+  }
+  precisionFormatterCache.set(key, formatter);
+  return formatter;
+};
 
 const getCurrencyFormatter = (currency: string) => {
   const normalizedCurrency = currency?.toUpperCase?.() ?? "USD";
@@ -304,6 +329,7 @@ export function formatAmount(
   amount: number | string | null | undefined,
   currency: string,
   displayCurrency = true,
+  precision?: number,
 ) {
   if (amount == null) return "-";
   const numericAmount = typeof amount === "string" ? Number(amount) : amount;
@@ -311,6 +337,13 @@ export function formatAmount(
   const displayAmount = Math.abs(numericAmount) < 0.005 ? 0 : numericAmount;
   const rawCurrency = currency ?? "USD";
   const isPenceCurrency = rawCurrency === "GBp" || rawCurrency === "GBX";
+
+  if (precision != null && precision !== DISPLAY_DECIMAL_PRECISION) {
+    const showCcy = displayCurrency && !isPenceCurrency;
+    const formatter = getPrecisionFormatter(precision, showCcy ? rawCurrency : undefined, showCcy);
+    const result = formatter.format(numericAmount);
+    return isPenceCurrency && displayCurrency ? `${result}p` : result;
+  }
 
   if (isPenceCurrency) {
     const formattedNumber = decimalFormatter.format(displayAmount);
--- a/apps/frontend/src/pages/asset/asset-edit-sheet.tsx
+++ b/apps/frontend/src/pages/asset/asset-edit-sheet.tsx
@@ -1020,7 +1020,7 @@ export function AssetEditSheet({
                                 latestQuote.close,
                                 latestQuote.currency,
                                 true,
-                                asset?.kind === "BOND" ? 4 : undefined,
+                                asset?.instrumentType === "BOND" ? 4 : undefined,
                               )}
                             </p>
                             <p className="text-muted-foreground text-xs">Latest price</p>
--- a/apps/frontend/src/pages/asset/asset-edit-sheet.tsx
+++ b/apps/frontend/src/pages/asset/asset-edit-sheet.tsx
@@ -1016,7 +1016,12 @@ export function AssetEditSheet({
                         <div className="grid grid-cols-3 gap-4 text-center">
                           <div>
                             <p className="text-xl font-semibold">
-                              {formatAmount(latestQuote.close, latestQuote.currency)}
+                              {formatAmount(
+                                latestQuote.close,
+                                latestQuote.currency,
+                                true,
+                                asset?.kind === "BOND" ? 4 : undefined,
+                              )}
                             </p>
                             <p className="text-muted-foreground text-xs">Latest price</p>
                           </div>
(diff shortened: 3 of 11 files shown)


marcoscale98:
@afadil
A simple change, but it's really helpful for avoiding loss of precision with Italian bonds.

afadil:
Thanks for tracking this down.  I'd like to suggest narrowing the frontend scope before merging.

### Summary

  - The Rust change is good as-is (with one suggested test).
  - The formatAmount global change has a blast radius far beyond bond prices and will subtly alter display for unrelated monetary values. I'd recommend making the extra
  precision opt-in instead.
  - The getDecimalPrecision default widens save precision for all non-FX kinds; an explicit BOND case would be more surgical.

  Concern: packages/ui/src/lib/utils.ts has a wide blast radius

  formatAmount is called from ~80+ sites across the app — not just price cells. It formats portfolio totals, contribution limits, goal target amounts, account deposits, cost
  basis, acquisition fees, AI-assistant amounts, holdings totals, chart tooltips, etc.

  With min=2, max=4, any value carrying >2 meaningful decimals will now render more digits than before:

  - goal.targetAmount = 10000.125 → was $10,000.13, now $10,000.125
  - Contribution limit 1234.5678 → was $1,234.57, now $1,234.5678
  - FP aggregation residuals on portfolio totals will now leak into the UI instead of rounding cleanly

  Users expect stable 2-decimal formatting for monetary totals; only prices need the extra digits. The bond problem is localized, but the current fix is global.

### Suggested shape

1.   Rust — keep, add one test

  Only the latest-quote path is covered. Add a historical-bars test so the map(|d| d / divisor) chain is locked down:

```
  #[test]
  fn historical_bond_close_preserves_precision() {
      let raw = 91.92_f64;
      let divisor = Decimal::from(100);
      let d = Decimal::try_from(raw).unwrap() / divisor;
      assert_eq!(d.to_string(), "0.9192");
  }
```

  2. formatAmount — revert the global default, add an opt-in arg

  ```
const DECIMAL_FORMAT_OPTIONS: Intl.NumberFormatOptions = {
    minimumFractionDigits: DISPLAY_DECIMAL_PRECISION, // 2
    maximumFractionDigits: DISPLAY_DECIMAL_PRECISION, // 2 — restore
  };

  export function formatAmount(
    amount: number | string | null | undefined,
    currency: string,
    displayCurrency = true,
    maxDecimals?: number, // new: opt-in for price cells
  ) {
    // ... existing null / NaN / pence handling ...

    const formatter =
      maxDecimals != null && maxDecimals !== DISPLAY_DECIMAL_PRECISION
        ? new Intl.NumberFormat("en-US", {
            ...(displayCurrency ? { style: "currency", currency: rawCurrency } : {}),
            minimumFractionDigits: DISPLAY_DECIMAL_PRECISION,
            maximumFractionDigits: maxDecimals,
          })
            minimumFractionDigits: DISPLAY_DECIMAL_PRECISION,
            maximumFractionDigits: maxDecimals,
            minimumFractionDigits: DISPLAY_DECIMAL_PRECISION,
            maximumFractionDigits: maxDecimals,
          })
        : displayCurrency
          ? getCurrencyFormatter(rawCurrency)
          : decimalFormatter;

    return formatter.format(numericAmount);
  }

```
  Intl.NumberFormat already trims trailing zeros when min < max, so 0.92 still renders as 0.92 while 0.9192 renders in full.

3.   Pass 4 only at price-render sites

  Audit specifically the price cells:

  - apps/frontend/src/pages/asset/quote-history-table.tsx:200,218,235,252 (OHLC)
  - apps/frontend/src/pages/asset/assets-table.tsx:235 (quote.close)
  - apps/frontend/src/pages/asset/assets-table-mobile.tsx:252 (price cell)
  - apps/frontend/src/pages/asset/asset-edit-sheet.tsx:953 (latestQuote.close)
  - apps/frontend/src/pages/asset/asset-lots-table.tsx:59 (acquisitionPrice — fees/costBasis at :62/65 should stay 2)

  Everything else — goals, contribution limits, deposits, AI amounts, holdings totals — keeps 2/2. No regression.

4.   quote-history-data-grid.tsx — make the bond case explicit

 ```
 const getDecimalPrecision = (assetKind?: AssetKind | null): number => {
    switch (assetKind) {
      case "FX":   return 6;
      case "BOND": return 4;
      default:     return 2;
    }
  };
```

  Narrower than the blanket 4, matches the actual issue, and leaves stock/ETF save behavior untouched.

5.   Frontend test

  Table test for formatAmount to lock the contract:

  - (1234, "USD") → "$1,234.00"
  - (1234.5678, "USD") → "$1,234.57" (default 2)
  - (0.9192, "USD", false, 4) → "0.9192"
  - (0.92, "USD", false, 4) → "0.92" (min=2, trailing zeros trimmed)

  Why this shape

  - Fixes the root cause in one place (bond price rendering) without touching 80+ unrelated call sites.
  - formatAmount(price, ccy, true, 4) reads as "this is a price, up to 4 decimals" — grepping the 4 callers gives you the exact bond-affected surface.
  - Extensible: later we can pass 8 for crypto prices without changing the global default.

marcoscale98:
Thanks for the detailed review @afadil — the narrowing makes sense. One concern before I implement, on the proposed `formatAmount(amount, ccy, displayCurrency, maxDecimals)` shape with `min=DISPLAY_DECIMAL_PRECISION (2)`, `max=maxDecimals (4)`:

**Column alignment in price tables.** `Intl.NumberFormat` with `min=2, max=4` trims trailing zeros, so within the same price column rows end up with different decimal widths. Concrete example — `quote-history-data-grid` (saved daily quotes) and the lots/transactions tables for a bond:

```
Date         Close
2026-04-22   0.9200   ← was, now renders as 0.92
2026-04-23   0.9192
2026-04-24   0.9180   ← renders as 0.918
2026-04-25   0.9200   ← renders as 0.92
```

The decimal points and digits no longer line up vertically. Today every monetary column is uniform 2/2 so this is a visual regression specifically on the price grids the change is meant to fix. Stocks/ETFs aren't affected (always 2 decimals already); the FX `precision: 6` in `quote-history-data-grid.tsx` only governs input `step`/`roundToDecimals` — display still goes through `formatAmount` at 2/2, so there's no precedent in the app for variable-width decimals.

Two options, both narrower than my original PR:

**Option A — fixed precision per call.** Change the new arg semantics so it sets both min and max:
```ts
formatAmount(amount, currency, displayCurrency = true, precision?: number)
// when precision != null → min = max = precision
// otherwise → min = max = DISPLAY_DECIMAL_PRECISION (2)  ← global default unchanged
```
Bond price cells pass `4` → all rows render as `0.9200` / `0.9192`, columns align. Same call-site surface as your snippet, just predictable padding instead of trimming. Extends naturally to crypto (`8`) later.

**Option B — two bond precisions based on quote convention.** 4 decimals only for bonds quoted in % of face value (Boerse Frankfurt → divisor `100`, so 100 → 1.0000), 2 for bonds quoted directly. More semantically accurate but requires propagating a "quoted-in-percent" flag from the provider through to the asset record — bigger change.

Open to other approaches you have in mind too — happy to follow whichever direction you prefer.

afadil:
I prefer option A : fixed precision (`min = max = precision`).
Column alignment on price grids matters more than zero-trimming, and there's no precedent in the app for variable-width decimal columns. 

Option B (propagating a quoted-in-percent flag from provider → asset record) is a bigger refactor with the same practical outcome for bonds today not worth it.

marcoscale98:
Hi @afadil, I've addressed all the review feedback. Here's a summary of what changed:

**Changes made:**

1. **`packages/ui/src/lib/utils.ts`** — Reverted `DECIMAL_FORMAT_OPTIONS` to `min=max=2` (default unchanged). Added opt-in `precision?: number` param to `formatAmount` with a dedicated cached `getPrecisionFormatter` (fixed `min=max=precision` → columns always aligned, no trailing-zero trimming).

2. **`apps/frontend/src/lib/utils.ts`** — Mirrored the same `precision?` param and `getPrecisionFormatter` logic so calls from frontend-local imports also type-check correctly.

3. **Price-render sites only** — `precision=4` is passed explicitly at:
   - `assets-table.tsx` — `quote.close`
   - `assets-table-mobile.tsx` — `quote.close`
   - `asset-edit-sheet.tsx` — `latestQuote.close`
   - `asset-lots-table.tsx` — `acquisitionPrice` only (fees and cost basis stay at 2)
   - `quote-history-data-grid.tsx` — OHLC cells (via `getDecimalPrecision` with explicit `if (instrumentType === "BOND") return 4`)

4. **Bond detection** — Used `instrumentType === "BOND"` (string field on `Asset`) instead of `AssetKind`, since `"BOND"` is not part of the `AssetKind` union type in the frontend.

5. **`quote-history-table.tsx`** — Updated for consistency, but note: this component appears to be dead code (not rendered anywhere in the current codebase). Happy to revert those changes if you'd prefer to keep it untouched.

6. **Rust test** — Added `historical_bond_close_preserves_precision` in `crates/market-data/src/provider/boerse_frankfurt/mod.rs`.

7. **Frontend vitest test** — Added `apps/frontend/src/lib/format-amount.test.ts` covering default 2-decimal behavior, rounding, and fixed 4-decimal precision with zero-padding.

All checks pass (cargo fmt, clippy, tests, pnpm type-check, build).

afadil:
Thanks for the contribution! This is already fixed in v3.8, closing as superseded.

PR #828, opened 2026-04-08
fix(frontend): download SQLite backup/export in web mode (#688)

### LINKED ISSUE #688: [Bug?] Self-hosted data export only saves on server
The app offers both Backups and Exports, but it looks like both are just creating files on the server, with neither offering a download. This makes using the backups/exports really difficult.

Using self hosted 3.0.4 with Docker.

AtomicTrxn:
## Description

Fixes self-hosted web backup/export flow for SQLite so users receive a browser download instead of only a server-side filename.

Root cause:
- Web `backup_database` already returned `{ filename, data }` from the backend.
- Frontend web paths (`useExportData` and `useBackupRestore`) ignored `data` and treated the result as server-only save.

Fix summary:
- Web SQLite paths now call `openFileSaveDialog(data, filename)` and treat success as a local save.
- Updated web backup/export copy to match actual behavior.
- Added regression tests for both hooks covering web SQLite download behavior.

Notes:
- Backend/API unchanged (`POST /utilities/database/backup` already returns data and filename).
- This PR replaces closed PR #703, which could not be reopened after the source branch was temporarily deleted.

## Test Evidence

- `pnpm run build:types` ✅
- `pnpm format:check` ✅
- `pnpm lint` ✅
- `pnpm type-check` ✅
- `pnpm test` ✅
- `pnpm build` ✅
- `cargo fmt --all -- --check` ✅
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` ✅
- `CONNECT_API_URL=http://test.local cargo test --workspace` ✅
- `cargo build -p wealthfolio-server --release` ✅

Fixes #688

## Checklist

### DIFF
--- a/apps/frontend/src/pages/settings/exports/backup-restore-form.tsx
+++ b/apps/frontend/src/pages/settings/exports/backup-restore-form.tsx
@@ -17,8 +17,8 @@ const desktopNotes = [
 ] as const;
 
 const webNotes = [
-  "Backups include WAL and SHM files and are stored in the server data directory.",
-  "Download or copy backup files directly from the host environment when needed.",
+  "Backups include WAL and SHM files and are downloaded directly to your browser.",
+  "Store downloaded backups somewhere safe so they can be restored later in desktop or iOS.",
   "Restores are only available in the desktop application.",
   "Create backups regularly, especially before bulk imports or migrations.",
 ] as const;
@@ -132,7 +132,7 @@ const WebBackupPanel = ({ performBackup, isBackingUp }: WebPanelProps) => {
 
       <BackupCard
         title="Create Backup"
-        description="Create a complete backup with WAL and SHM files stored automatically in the server data directory for safekeeping."
+        description="Create a complete backup with WAL and SHM files and download it directly to your device."
         isLoading={isBackingUp}
         disabled={isBackingUp}
         actionLabel="Backup Database"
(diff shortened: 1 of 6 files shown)


afadil:
Thanks for the contribution.

I think the web backup flow needs a different approach. Here's some context and what the intended solution looks like:
 **Current intended behavior:**
  - Web backups are server-side operations. The backup is saved to the server's data directory, and the user is told the filename.
  - This is by design for self-hosted users — backups live on the host where they can be picked up by server-level backup scripts (cron, rsync, etc.).
  - The toast message "Backup created on the server as {filename}" reflects this intent.

  **Current bug:**
  - The POST /utilities/database/backup endpoint base64-encodes the entire database and returns it in the JSON response (data_b64), even though the frontend only
  uses filename and discards the data.
  - **On large databases (hundreds of MB+), this wastes bandwidth and memory, and can crash the browser tab.**

 ## Intended solution:
I was thinking about something like the following:

  **Backend:**
  - POST /utilities/database/backup — keep as-is but stop returning data_b64. Only return { filename }.
  - GET /utilities/database/backups — new endpoint to list available backup files in the server data directory.
  - GET /utilities/database/backups/:filename — new endpoint that **streams the backup** file as application/octet-stream with Content-Disposition: attachment. The
  browser handles the download natively, no JS memory overhead.

  **Frontend (web mode):**
  - Backup page shows a list of existing server-side backups.
  - "Create Backup" button triggers the backup and refreshes the list.
  - Each backup row has a "Download" button that points to the streaming endpoint.

  Why this approach:
  - Self-hosted users expect backups to live on the server, next to their data. A browser download as the primary flow doesn't fit this use case.
  - Listing + streaming gives users full visibility and control over their backups without the memory risk.
  - Streaming avoids holding the full file in JS memory, so it works regardless of DB size.
  
  Curious what do you think? heavy users can have 1Go db file

afadil:
Implemented in latest version

PR #829, opened 2026-04-08
feat(ai): add Claude Subscription provider via local Claude Code

obouafif:
## Description

Add a new AI provider — **Claude (Subscription)** — that lets users access
Claude through their local Claude Code (Pro/Max) subscription instead of an
Anthropic API key. The existing API-key path is unchanged; this is purely
additive.

In the UI, the new provider is presented as an embedded sub-mode of the
existing Anthropic provider card (one Anthropic entry, two connection
methods), so the settings page stays clean.

### Why

Users on a Claude Pro / Max plan already pay a flat monthly fee. Today the
only way to use Claude in Wealthfolio is to also create an Anthropic API key
and pay per token on top of that subscription. This change lets those users
re-use the Claude they already pay for.

### How it works

```
ChatService ─► subscription_chat_stream() ─► Node sidecar (bridge.mjs)
                                                    │
                                                    ▼
                                          @anthropic-ai/claude-agent-sdk
                                                    │
                                                    ▼
                                          local Claude Code (authed)
```

- The Rust backend spawns a small Node sidecar at
  `apps/tauri/sidecars/claude-agent-bridge` that wraps
  `@anthropic-ai/claude-agent-sdk`. The SDK transparently uses the local
  Claude Code login as the auth source.
- The Rust ↔ Node bridge speaks NDJSON over stdio:
  `start_query` / `tool_call` / `tool_result` / `done`.
- Wealthfolio's existing portfolio tools (`get_holdings`, `get_performance`,
  …) are exposed to Claude via an in-process MCP server in the bridge.
  When Claude calls a tool, the bridge emits a `tool_call` event; the Rust
  side dispatches it through the same `rig::tool::ToolDyn` registry the
  rig-core path already uses, then sends the result back through stdin.
  **No tool logic is duplicated** — both code paths go through a new shared
  helper `ToolSet::into_allowed_tools()`.
- `spawn_chat_stream()` short-circuits to the new path when the provider id
  is `claude-subscription`. The rig-core path is otherwise untouched.

### Security model

- `ANTHROPIC_API_KEY` is stripped from the sidecar's environment so the SDK
  can never accidentally fall back to API-key billing.
- All default Claude Code tools that touch the filesystem or shell
  (`Bash`, `Read`, `Write`, `Edit`, `Glob`, `Grep`, `Agent`,
  `AskUserQuestion`) are dropped via `tools: []`.
- Only **`WebSearch`** and **`WebFetch`** are whitelisted as built-in tools
  (read-only network access for stock prices, news, ticker info).
- Wealthfolio's per-provider tools allowlist is honored end-to-end: any tool
  not in the allowlist is rejected with `AiError::ToolNotAllowed` before
  reaching the SDK.
- Per-thread token cap (100 k default) cancels runaway loops.
- Settings are loaded from neither `~/.claude` nor `CLAUDE.md` files
  (`settingSources: []`), so the sidecar can't be influenced by external
  user instructions.

### Code-quality notes

- The 11-arm tool-builder ladder previously inlined inside a macro in
  `chat.rs` is now a single `ToolSet::into_allowed_tools()` helper used by
  both code paths. Adding a new tool only needs one site.
- `dispatch_tool` returns `Result<Value, AiError>`, so tool failures
  surface with their proper error code (`TOOL_NOT_ALLOWED`,
  `TOOL_EXECUTION_FAILED`) in stream events.
- `subscription_chat_stream` always emits a terminal `Done` event even on
  failure, so the chat UI never hangs.
- The dev-only workspace walk-up in `resolve_bridge_path` is gated behind
  `#[cfg(debug_assertions)]`.
- The cached bridge path uses a `OnceCell<PathBuf>` so we don't re-stat
  on every spawn.

### Known follow-up (not in this PR)

The sidecar is currently spawned **per chat message** (~300–700 ms of Node
startup per turn). The bridge already supports multiplexing queries by
`threadId`, so the natural next step is to hold a single shared
`Arc<SidecarHandle>` in `ChatService` and reuse it across messages, with
cancellation propagated through the existing `SidecarCommand::Cancel`.
Documented inline at `SidecarHandle::spawn` with a `NOTE(perf)` comment.
Happy to land that as a follow-up PR if the maintainers prefer a smaller
first change.

### Testing

- **Sidecar smoke test** (`apps/tauri/sidecars/claude-agent-bridge/smoke-test.mjs`):
  spawns the bridge as a child process and exercises a full
  ping → start_query → tool_call → tool_result → done round-trip with a
  fake bridged tool. No Tauri or Wealthfolio app required:
  ```bash
  cd apps/tauri/sidecars/claude-agent-bridge
  npm install
  node smoke-test.mjs
  ```
  Expected output: `[smoke] PASS — bridge protocol works end-to-end with
  tool round-trip.`

- **Manual end-to-end** in the desktop app:
  1. Install Claude Code and `claude login` with a Pro/Max account.
  2. `pnpm tauri dev`
  3. Settings → AI Providers → Anthropic → expand the card → enable
     **Claude Code Subscription**
  4. Open the AI chat, pick **Claude (Subscription)** in the provider
     picker, send "Show me my current portfolio holdings"
  5. Expected: streamed text → `get_holdings` tool call → real numbers
     from the local DB → final assistant message using those numbers.
     No `ANTHROPIC_API_KEY` required.

### Files changed

**Rust (backend)**
- `crates/ai/src/claude_subscription.rs` — new module: sidecar process
  management, NDJSON protocol, streaming loop, tool dispatch, prompt
  rendering (~660 LOC).
- `crates/ai/src/chat.rs` — early branch in `spawn_chat_stream` routes
  the new provider to the sidecar path. Existing rig-core paths refactored
  to use the new shared `ToolSet::into_allowed_tools()` helper instead of
  the inline 11-arm ladder.
- `crates/ai/src/tools/mod.rs` — new `ToolSet::into_allowed_tools()`
  helper, used by both chat code paths.
- `crates/ai/src/lib.rs` — `pub mod claude_subscription;`
- `crates/ai/src/provider_service.rs` — `list_models` short-circuit for
  the subscription provider (returns the catalog entry directly instead
  of calling Anthropic's API, which would 401 without a key).
- `crates/ai/src/ai_providers.json` — new `claude-subscription` provider
  entry with `type: "subscription"`.
- `crates/ai/Cargo.toml` — enable `tokio` features `process` and
  `io-util`.

**Node sidecar** (`apps/tauri/sidecars/claude-agent-bridge/`)
- `bridge.mjs` — long-running NDJSON server. Wraps the Agent SDK's
  `query()`, exposes Wealthfolio tools through an in-process MCP server,
  forwards `tool_call`s to the Rust parent.
- `smoke-test.mjs` — standalone integration test for the bridge protocol.
- `package.json` / `package-lock.json` — pins
  `@anthropic-ai/claude-agent-sdk`.

**Frontend**
- `apps/frontend/src/features/ai-assistant/components/provider-settings-card.tsx`
  — embedded "Claude Code Subscription" sub-mode UI inside the Anthropic
  card; new optional `subscriptionProvider` / `onToggleSubscription` props.
- `apps/frontend/src/pages/settings/ai-providers/ai-providers-page.tsx`
  — filters `claude-subscription` out of the standalone list, surfaces it
  to the Anthropic card via props. Hoisted memos above early returns to
  comply with the Rules of Hooks.
- `apps/frontend/src/features/ai-assistant/hooks/use-chat-model.ts` and
  `use-provider-picker.ts` — allow `type: "subscription"` providers
  through the enabled-provider filter.

## Checklist

### DIFF
--- a/apps/frontend/src/features/ai-assistant/components/model-picker.tsx
+++ b/apps/frontend/src/features/ai-assistant/components/model-picker.tsx
@@ -30,6 +30,12 @@ export const ModelPicker: FC = () => {
     const favoriteIds = currentProvider.favoriteModels ?? [];
     const allModels: MergedModel[] = currentProvider.models ?? [];
 
+    // Subscription providers (e.g. Claude via Claude Code) have a small fixed
+    // model set — always show all of them regardless of favorites.
+    if (currentProvider.type === "subscription") {
+      return allModels;
+    }
+
     // If no favorites set, fall back to all catalog models
     if (favoriteIds.length === 0) {
       return allModels;
--- a/apps/frontend/src/features/ai-assistant/components/provider-settings-card.tsx
+++ b/apps/frontend/src/features/ai-assistant/components/provider-settings-card.tsx
@@ -39,6 +39,12 @@ interface ProviderSettingsCardProps {
   onSetCapabilityOverride?: (modelId: string, overrides: ModelCapabilityOverrides | null) => void;
   onToolsAllowlistChange?: (tools: string[] | null) => void;
   isLast?: boolean;
+  /**
+   * For the Anthropic provider card, the related "Claude (Subscription)"
+   * provider state, surfaced as an embedded sub-mode toggle.
+   */
+  subscriptionProvider?: MergedProvider | null;
+  onToggleSubscription?: (enabled: boolean) => void;
   // Model fetching props (controlled by parent via React Query)
   modelComboboxOpen?: boolean;
   onModelComboboxOpenChange?: (open: boolean) => void;
@@ -85,6 +91,9 @@ export function ProviderSettingsCard({
   onSetCapabilityOverride,
   onToolsAllowlistChange,
   isLast = false,
+  // Embedded sub-provider (Claude Subscription inside Anthropic)
+  subscriptionProvider,
+  onToggleSubscription,
   // Model fetching props
   modelComboboxOpen: controlledComboboxOpen,
   onModelComboboxOpenChange,
@@ -390,6 +399,61 @@ export function ProviderSettingsCard({
                         Save
                       </Button>
                     </div>
+                  </div>
+                </div>
+              )}
+
+              {/* Claude Code Subscription — embedded sub-mode of Anthropic.
+                  Lets the user authenticate Claude through their local Claude Code
+                  (Pro/Max) install instead of, or alongside, an API key. */}
+              {provider.id === "anthropic" && subscriptionProvider && (
+                <div className="border-primary/15 bg-primary/5 rounded-lg border border-dashed p-4">
+                  <div className="flex items-start justify-between gap-3">
+                    <div className="min-w-0 flex-1 space-y-1.5">
+                      <div className="flex flex-wrap items-center gap-2">
+                        <Label
+                          htmlFor="claude-subscription-toggle"
+                          className="text-sm font-medium"
+                        >
+                          Claude Code Subscription
+                        </Label>
+                        <Badge
+                          variant="outline"
+                          className="h-5 px-1.5 text-[10px] font-normal"
+                        >
+                          Alternative to API key
+                        </Badge>
+                        {subscriptionProvider.enabled && (
+                          <Badge
+                            variant="outline"
+                            className="border-success/30 bg-success/10 text-success h-5 px-1.5 text-[10px] font-normal"
+                          >
+                            Active
+                          </Badge>
+                        )}
+                      </div>
+                      <p className="text-muted-foreground text-xs leading-relaxed">
+                        Use Claude through your local Claude Code (Pro / Max)
+                        subscription. No API key required, no per-token billing —
+                        usage is covered by your existing Claude plan. Requires
+                        Claude Code to be installed and logged in on this machine.
+                      </p>
+                      {subscriptionProvider.enabled && (
+                        <p className="text-muted-foreground flex items-center gap-1.5 pt-1 text-xs">
+                          <Icons.Sparkles className="h-3 w-3" />
+                          Selectable in chat as{" "}
+                          <span className="text-foreground font-medium">
+                            Claude (Subscription)
+                          </span>
+                        </p>
+                      )}
+                    </div>
+                    <Switch
+                      id="claude-subscription-toggle"
+                      checked={subscriptionProvider.enabled}
+                      onCheckedChange={(enabled) => onToggleSubscription?.(enabled)}
+                      className="data-[state=checked]:bg-success mt-1 shrink-0"
+                    />
                   </div>
                 </div>
               )}
--- a/apps/frontend/src/features/ai-assistant/hooks/use-chat-model.ts
+++ b/apps/frontend/src/features/ai-assistant/hooks/use-chat-model.ts
@@ -109,8 +109,11 @@ export function useChatModel(): ChatModelState {
   // Ensure resolved model is accessible in the provider's favorites.
   // ModelPicker filters by favorites — if the resolved model isn't among them,
   // the picker shows "No model" while the hook sends a different model to the backend.
+  // Subscription providers show all catalog models (not filtered by favorites),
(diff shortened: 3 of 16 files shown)


Jonjon-prog:
Hi @obouafif , 

I tested this PR and overall it works well and is surprisingly fast! 
Great work on the security model and the bridge architecture.

A few things I noticed:

**Bug — zod peer dependency conflict**
`npm install` fails out of the box due to a conflict between `zod@^3.x` 
(pinned in `package.json`) and `@anthropic-ai/claude-agent-sdk@0.2.92` 
which requires `zod@^4.0.0`. Had to use `npm install --legacy-peer-deps` 
to get it working.

**Missing chat history persistence**
Conversations are not saved, unlike other providers (e.g. Ollama). 
Not sure if this is intentional for the first iteration, but worth noting.

**No model selection**
The provider always uses Claude Code's default model — there's no way 
to pick a specific model (e.g. claude-opus vs claude-sonnet). 
Could be a follow-up improvement.

Overall a solid contribution, the value proposition for Pro/Max subscribers 
is clear. Thanks!

skoppisetty:
Checkout https://github.com/skoppisetty/cli-agents-rs 

It would be nice to just wrap all local CLIs  like claude, codex and gemini under one tool. I use this for my personal tool and it works great.

obouafif:
> Checkout https://github.com/skoppisetty/cli-agents-rs
> 
> It would be nice to just wrap all local CLIs like claude, codex and gemini under one tool. I use this for my personal tool and it works great.

   Thanks for sharing @skoppisetty — cool project! 🔥
  
   For this PR we went with the Node sidecar + @anthropic-ai/claude-agent-sdk approach to get deep integration with Claude's tool system and MCP server
   bridging. But wrapping multiple local CLIs (Codex, Gemini CLI, etc.) under a unified interface is a great idea for expanding subscription-based provider support.
 
   Would make a solid follow-up PR if there's interest — feel free to open an issue to track it!

afadil:
I think this might violate Anthropic's Consumer ToS. Don't they prohibit using subscription credentials for automated/third-party access?

obouafif:
> I think this might violate Anthropic's Consumer ToS. Don't they prohibit using subscription credentials for automated/third-party access?

   You're absolutely right, and thanks for flagging this — I went and dug into it properly.
  
   This was in a grey area until recently, but Anthropic explicitly closed the door:
  
   - Feb 2026 — Anthropic updated the Consumer Terms + Claude Code docs to state that using OAuth tokens from Free/Pro/Max accounts in any other product,
   tool, or service — including via the Agent SDK — is not permitted. (https://code.claude.com/docs/en/legal-and-compliance,
   https://platform.claude.com/docs/en/agent-sdk/overview)
   - April 4, 2026 — Anthropic began enforcing this: Claude Pro/Max subscription limits no longer cover third-party harnesses like OpenClaw. Users must
   switch to an API key or a separately-billed "extra usage" bundle. (https://www.theregister.com/2026/04/06/anthropic_closes_door_on_subscription/,
   https://techcrunch.com/2026/04/04/anthropic-says-claude-code-subscribers-will-need-to-pay-extra-for-openclaw-support/)
  
   Direct quote from Anthropic's docs:
    "Anthropic does not allow third party developers to offer claude.ai login or rate limits for their products, including agents built on the Claude
   Agent SDK."
  
   The architecture in this PR (Node sidecar → @anthropic-ai/claude-agent-sdk using the user's local Claude Code login) is exactly the pattern that was
   just banned, even though credentials never leave the user's machine. A whole ecosystem of tools (OpenClaw, NanoClaw, Cline, Aider, Crush, meridian,
   claude-code-gateway, etc.) is currently dealing with the same fallout.
  
   Given that, I don't want to ship something that could get Wealthfolio users' Claude accounts flagged or banned. I'm closing this PR. The existing
   Anthropic API-key provider already covers the legitimate path, and we can revisit if Anthropic ever publishes an approved route for OSS desktop apps.
  
   Thanks again for catching this before it merged. 🙏

PR #839, opened 2026-04-11
fix(accounts): clean up snapshots and valuations on account deletion

triantos:
## Summary

- `holdings_snapshots` and `daily_account_valuation` have no FK CASCADE
  on `account_id`, so deleting an account left orphaned rows in these
  tables.
- Add explicit deletes for both tables in `delete_account` before the
  account row is removed.
- This also enables a faster wipe-and-reimport workflow: delete the
  account (instant CASCADE of activities + lots + snapshots +
  valuations) then recreate and reimport, instead of deleting activities
  one-by-one.

## Test plan
- [x] `cargo fmt`, `cargo clippy`, `cargo test` all pass
- [x] Verified `holdings_snapshots` and `daily_account_valuation` have
  no FK constraint on `account_id` (confirmed via `sqlite_master` schema)
- [x] Both `delete_snapshots_by_account_ids` and
  `delete_valuations_for_account` are existing repository methods with
  their own test coverage


afadil:
Superseded by #1745, which adds transactional cleanup for local and synced account deletions and repairs existing orphan data.

PR #831, opened 2026-04-09
Phase B: valuation consistency, lots hardening, UI polish

triantos:
## Summary

Phase B of the data-model refactor. Depends on #792 (Phase A — materialized
lots table).

### Valuation & net worth
- Route negative asset categories to liabilities section
- Track alternative asset value separately for the investments tab
- Replace TOTAL pseudo-account valuations with a dedicated
  `daily_portfolio_valuation` table
- Replace "TOTAL" string literals with a `PORTFOLIO_TOTAL_ACCOUNT_ID` constant
- Staleness badge when portfolio data is > 1 day old

### Lots / holdings
- Rich lot viewer with account grouping and open/closed status
- Persist fully-consumed lots (closed lots were previously dropped during
  full recalc when a buy and sell were processed in the same pass)
- Remove the legacy `lots` display field from `Holding` — frontend reads
  `lot_details` exclusively
- Set `lots.open_activity_id` so FK CASCADE clears lots when an activity is
  deleted (the column was always declared but never populated)
- Refuse to wipe lots on empty `sync_lots_for_account` inputs (safety net
  against calculator bugs that produce empty state)

### Schema / backend
- Add nullable `account_id` FK to `assets` table (for alternative assets
  linked to accounts)
- Optional account link UI for alternative assets
- Valuation service reads asset metadata from the asset repo instead of
  snapshot positions
- Eliminate positions JSON from snapshot DB reads in the calculator
  (positions are now rehydrated from the lots table at query time;
  `holdings_snapshots.positions` remains as the HOLDINGS-mode storage column)

### Tests
- New `scripts/test-lots-migration.sh regression` mode with 5 mutation
  scenarios covering incremental recalc, activity delete isolation,
  HOLDINGS snapshot delete, recalc idempotency, and
  holdings↔valuations consistency. Each scenario asserts the
  sum-of-lots invariant (lots table matches `/holdings`).

## Test plan
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace` (with `CONNECT_API_URL=http://test.local`)
- [x] `cargo build -p wealthfolio-server --release` (with and without Connect env)
- [x] Frontend: `pnpm format:check && pnpm lint && pnpm type-check && pnpm test && pnpm build`
- [x] `scripts/test-lots-migration.sh regression` — 28/28 assertions pass on a fresh DB
[...]

PR #842, opened 2026-04-13
feat: replace snapshot positions JSON with relational table

triantos:
> **Depends on #831 (Phase B)** — review only the last commit until Phase B merges.

## Summary

- Creates `snapshot_positions` table as the relational source of truth for HOLDINGS-mode position data, replacing the `positions` JSON blob in `holdings_snapshots`
- Migration populates the new table from existing JSON, then clears the JSON column to `'{}'`
- All read paths now load positions from the relational table; the JSON column is vestigial (always `'{}'`, never read)
- Fixes pre-existing bug where `refresh_lots_from_latest_snapshot` didn't handle `Synthetic` snapshots

## Detail

**Schema:** `snapshot_positions` with integer autoincrement PK, unique `(snapshot_id, asset_id)`, FK CASCADE to `holdings_snapshots` and `assets`. Migration uses `printf('%.20f')` to avoid SQLite scientific notation when extracting decimal values from JSON.

**Write path:** Repository `save_snapshots` and `save_or_update_snapshot` capture positions before the `From` conversion (which now writes `'{}'`) and insert them into `snapshot_positions` directly.

**Read path:** `From<AccountStateSnapshotDB>` sets `positions: HashMap::new()`. Call sites that need positions enrich from the table:
- TOTAL portfolio aggregation (batch load via `get_snapshot_positions_batch`)
- `save_manual_snapshot` content comparison
- `ensure_holdings_history` synthetic backfill (both snapshot_service.rs and broker/service.rs copies)
- Broker service legacy fallback path

**JSON column kept for sync compatibility:** The `positions` column remains in the schema (always `'{}'`) because the sync replay system rejects unknown payload columns. Dropping it would require a coordinated multi-device upgrade. The column is inert and costs negligible storage.

## Test plan

- [ ] All 894 existing tests pass (798 core + 84 storage + 12 property)
- [ ] `cargo fmt`, `cargo clippy`, `prettier --check` all clean
- [ ] Deployed to test server: migration populated 45 position rows from 20 HOLDINGS snapshots, holdings API returns correct data
- [ ] Verified: no non-empty positions JSON remains after migration


triantos:
Superseded by #856 — the three open data-model PRs (#792, #831, #842) have been consolidated into a single branch (`refactor_data_model`) with a single migration and a clean commit layout for easier review.

PR #843, opened 2026-04-13
feat: symbol mapping validation with real-time feedback and E2E tests

### LINKED ISSUE #833: UX: Improve symbol mapping input with autocomplete and validation feedback
## Problem

The Symbol Mapping screen (in the asset's Market Data tab) lacks feedback and guidance, making it hard to use correctly.

Image
(LYMS does not retrieve data, LYMS.DE does it instead. But no feedback is received)  

**Current issues:**

1. **No autocomplete** — The symbol input is a plain text field. Users must know the exact ticker format for each provider (e.g. `LYMS.DE` for Yahoo Finance) with no hints or suggestions.

2. **No format guidance** — There's no indication of what format the symbol should be in for the selected provider. Users don't know if they should enter `AAPL`, `AAPL.US`, `NASDAQ:AAPL`, etc.

3. **No real-time validation** — While typing, there's no feedback on whether the symbol is recognized by the provider or whether price data is available for it.

4. **No validation on save** — When pressing "Save Changes", the symbol is saved silently even if it's unknown to the provider or will return no prices. There's no confirmation or warning that the mapping will work.

## Expected Behavior

- Autocomplete or search suggestions while typing a symbol (filtered by the selected provider with effective data).
- A format hint or placeholder text showing the expected format per provider (e.g. `e.g. AAPL, LYMS.DE`).
- Real-time validation: show a checkmark/error after the user stops typing to indicate if the symbol is found and has price data.
- On save: warn the user if the symbol could not be validated (no prices found), rather than saving silently.

## Impact

Users end up with broken symbol mappings without realizing it, causing assets to show stale or missing prices silently.

marcoscale98:
## Summary

- **Real-time symbol validation** in the Market Data tab: each provider override row now validates the typed symbol against the selected provider (Yahoo Finance, Börse Frankfurt, etc.) with an 800ms debounce, showing a spinner → green check or red error icon
- **Provider-specific validation**: threaded `preferred_provider` through the full stack (Rust `resolve_symbol_quote` trait → Tauri command → Axum route → TypeScript adapter) so validation uses the exact provider selected in each row, not the default
- **Save warning**: a non-blocking toast warns when saving with invalid symbol mappings
- **E2E tests**: 4 Playwright tests covering Yahoo Finance (AAPL valid, invalid ticker) and Börse Frankfurt (DE0007164600 ISIN valid, invalid ticker), with `data-testid` attributes on validation icons for reliable selection

## Test Plan

- [x] Open an asset's Market Data tab → enable Automatic pricing → click Add → type `AAPL` with Yahoo Finance selected → green check appears within ~2s
- [x] Same flow with `INVALID_TICKER_XYZ_E2E` → red error icon appears
- [x] Change provider to Börse Frankfurt → type `DE0007164600` → green check
- [x] Change provider to Börse Frankfurt → type `INVALID_TICKER_XYZ_E2E` → red error icon appears
- [x] E2E suite passes: `npx playwright test e2e/01-happy-path.spec.ts e2e/10-symbol-mapping-validation.spec.ts` (01 must run first to complete onboarding; requires fresh DB from `node scripts/prep-e2e.mjs`)

Closes #833

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.claude/skills/run-e2e-tests/SKILL.md
+++ b/.claude/skills/run-e2e-tests/SKILL.md
@@ -1,0 +1,14 @@
+---
+name: run-e2e-tests
+description: >
+  How to run Wealthfolio E2E tests correctly. Use this skill whenever you are
+  about to run, execute, or trigger Playwright E2E tests — including running a
+  single spec file, the full test suite, or checking whether E2E tests pass.
+  Also use it when setting up the environment for E2E tests or troubleshooting
+  E2E test failures. Do NOT skip this skill just because the task seems simple
+  — the environment setup is mandatory and skipping it leads to silent failures.
+---
+
+# Running Wealthfolio E2E Tests
+
+Read `e2e/README.md` and follow its instructions **exactly and in order** before running any E2E test command. Do not skip any step.
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ db/
 apps/db/
 apps/tauri/db/
 tasks
+**/*.local.md
--- a/apps/frontend/src/adapters/shared/market-data.ts
+++ b/apps/frontend/src/adapters/shared/market-data.ts
@@ -187,12 +187,14 @@ export const resolveSymbolQuote = async (
   symbol: string,
   exchangeMic?: string,
   instrumentType?: string,
+  providerId?: string,
 ): Promise<ResolvedQuote | null> => {
   try {
     return await invoke<ResolvedQuote>("resolve_symbol_quote", {
       symbol,
       exchangeMic,
       instrumentType,
+      providerId,
     });
   } catch (_error) {
     logger.error("Error resolving symbol quote.");
--- a/apps/frontend/src/adapters/web/core.ts
+++ b/apps/frontend/src/adapters/web/core.ts
@@ -711,15 +711,17 @@ export const invoke = async <T>(command: string, payload?: Record<string, unknow
       break;
     }
     case "resolve_symbol_quote": {
-      const { symbol, exchangeMic, instrumentType } = payload as {
+      const { symbol, exchangeMic, instrumentType, providerId } = payload as {
         symbol: string;
         exchangeMic?: string;
         instrumentType?: string;
+        providerId?: string;
       };
       const params = new URLSearchParams();
       params.set("symbol", symbol);
       if (exchangeMic) params.set("exchangeMic", exchangeMic);
       if (instrumentType) params.set("instrumentType", instrumentType);
+      if (providerId) params.set("provider_id", providerId);
       url += `?${params.toString()}`;
       break;
     }
--- a/apps/frontend/src/adapters/web/core.ts
+++ b/apps/frontend/src/adapters/web/core.ts
@@ -721,7 +721,7 @@ export const invoke = async <T>(command: string, payload?: Record<string, unknow
       params.set("symbol", symbol);
       if (exchangeMic) params.set("exchangeMic", exchangeMic);
       if (instrumentType) params.set("instrumentType", instrumentType);
-      if (providerId) params.set("provider_id", providerId);
+      if (providerId) params.set("providerId", providerId);
       url += `?${params.toString()}`;
       break;
     }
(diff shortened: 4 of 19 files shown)


afadil (commented):
## Review

### Bug (blocking)

**Web adapter sends wrong query parameter name** — `apps/frontend/src/adapters/web/core.ts`

The Axum struct `ResolveSymbolQuoteQuery` uses `#[serde(rename_all = "camelCase")]` (`apps/server/src/api/market_data.rs:241`), which means the Rust field `provider_id` expects the query parameter name `providerId` (camelCase). But the web adapter sends `provider_id` (snake_case):

```ts
// PR adds:
if (providerId) params.set("provider_id", providerId);  // ← snake_case
```

Should be:
```ts
if (providerId) params.set("providerId", providerId);
```

As-is, **provider-specific validation silently falls back to default provider resolution in web mode**. It works in Tauri mode (which uses IPC, not HTTP query params), so this could easily be missed during testing.

---

### Medium issues

**1. Validation fires on mount for pre-existing mappings**

When the sheet opens with already-saved symbol mappings, `useWatch` picks up those values and the `useEffect` triggers validation API calls for every row. Opening the edit sheet for an asset with 3 saved mappings = 3 network requests just to re-validate what's already persisted.

Consider initializing `validationStatus` to `"valid"` for rows loaded from the database, or skipping validation until the value actually changes from its initial value.

**2. `as any` type casts on form field names**

The original inline code uses typed field names without casts. The extracted `SymbolMappingRow` adds `as any` to every field name:

```tsx
name={`providerConfig.${index}.symbol` as any}
```

This loses type safety that existed before. Worth fixing the type rather than escaping with `as any`.

---

### Minor observations

**3. Price === 0 treated as invalid**

```tsx
const status: SymbolValidationStatus = result?.price ? "valid" : "invalid";
```

JS truthiness: `0` is falsy. `result?.price != null` would be more precise.

**4. `symbolValidations` not reset on asset change**

It's reset when the sheet opens, but not when `asset` changes while the sheet is open. Low risk but worth noting.

**5. E2E `waitForTimeout` calls**

~10 `waitForTimeout` calls (300–1000ms) as padding. Playwright best practice is proper waiters (`waitForSelector`, `expect(...).toBeVisible()`). Not blocking, but worth cleaning up if tests prove flaky in CI.

**6. Unrelated lockfile change**

The -52 lines in `pnpm-lock.yaml` don't seem related to this feature. Worth confirming this is intentional.

---

### What looks good

- Rust trait change is clean — `preferred_provider: Option<&str>` added consistently across trait, impl, and all call sites.
- `provider_config` JSON construction in `service.rs` correctly matches how `Asset::preferred_provider()` parses it.
- Extracting `SymbolMappingRow` as a component is the right call — keeps validation state scoped per row.
- `onValidationChange` is stable (`useCallback` with `[]` deps), no re-render loops.
- Non-blocking save with toast warning is the right UX.

---

**Verdict:** The web adapter param name bug needs a fix before merge — it silently breaks provider-specific validation in web mode. The rest ranges from "nice to fix" to "fine as-is."

marcoscale98:
Thanks for the thorough review! 

The review comments have not been addressed yet — we're tracking them and will follow up in a subsequent commit. The latest push adds robustness improvements to the E2E tests (retry loop for Radix dropdown race, cleanup of leftover mapping rows between tests).

marcoscale98:
All issues flagged in the review have been addressed:

**Blocking**
- `provider_id` → `providerId` (camelCase) in the web adapter query param — provider-specific validation now works correctly in web mode too.

**Medium**
- Validation no longer fires on mount for pre-existing mappings: `validationStatus` initialises to `"valid"` for rows loaded from the DB, and a `useRef` guard skips the first effect run when the symbol hasn't changed.
- `as any` casts replaced with `as Path<AssetFormValues>` throughout `SymbolMappingRow`.

**Minor**
- `result?.price != null` instead of `result?.price` — price=0 now correctly treated as valid.
- `symbolValidations` reset when `asset?.id` changes while the sheet is open.
- `waitForTimeout` calls in spec 10 replaced with DOM-state waiters (`toHaveCount`, `not.toHaveCount`, `toBeVisible`).

All 5 symbol-mapping E2E tests pass consistently.

PR #864, opened 2026-04-17
fix(activity): securities transfers show qty × price, never a stored amount

loreboldo:
First of all, a huge thank you for Wealthfolio - it's a joy to use and the local-first, privacy-respecting philosophy is exactly what this space needed. I've been running it self-hosted on a spare phone over local network / Tailscale and it is becoming my daily portfolio tracker.

Disclosure: I had Claude assist me in investigating the bug and drafting this change. Every line has been reviewed and tested by me before submitting.

## Summary
Securities `TRANSFER_IN` / `TRANSFER_OUT` were being rendered in the activity view from the stored `amount` column, which is redundant with `quantity × unit_price` and in the wild can hold corrupted values (I had two rows where `amount = qty² × unit_price`, producing absurd totals like €31M for 2,078 shares at €7.29). The account page was correct because it already derived value from `qty × price` — only the activity table was affected.

The fix is in two layers:

- **Display** — new `isSecuritiesTransfer` helper; `calculateActivityValue`, the activity table, and the mobile activity card all now derive securities-transfer value from `qty × unit_price` and ignore `amount`. Cash transfers are unchanged (still use `amount`).
- **Write path (defensive)** — so a bad `amount` can't be persisted in the first place:
  - `bulk-holdings-modal` no longer sends a precomputed `amount` for the `TRANSFER_IN` it creates.
  - `ActivityService::prepare_new_activity`, `prepare_update_activity`, and `prepare_activities_internal` force `amount = None` whenever a transfer's resolved asset is non-cash. This mirrors how the dividend-in-kind compiler already leaves `transfer_in_leg.amount = None`, making `qty × unit_price` the single source of truth for securities transfers.

## Test plan
- [x] `pnpm test` — all 505 frontend tests pass, including a new regression in `activity-utils.test.ts` that feeds a deliberately bogus `amount` and asserts the value still comes from `qty × unit_price`.
- [x] Backend unit tests for the new `is_securities_transfer` classifier (cash vs. security vs. non-transfer).
- [x] Verified on my instance: after nulling the corrupted `amount` column, the two rows now show correct totals.
- [ ] Reviewer: confirm cash transfers (`CASH:USD`, `$CASH-EUR`, …) still show `amount` unchanged in the activity view.
- [ ] Reviewer: confirm a new *Add Existing Holdings* entry persists with `amount IS NULL` in SQLite.

### DIFF
--- a/apps/frontend/src/lib/activity-utils.test.ts
+++ b/apps/frontend/src/lib/activity-utils.test.ts
@@ -211,6 +211,50 @@ describe("Activity Utilities", () => {
       expect(calculateActivityValue(transferOut)).toBe(1010);
     });
 
+    it("treats blank-asset transfers as cash and uses amount", () => {
+      const transferIn = createActivity({
+        activityType: ActivityType.TRANSFER_IN,
+        assetSymbol: "",
+        assetId: "",
+        quantity: "0",
+        unitPrice: "0",
+        amount: "500",
+        fee: "0",
+      });
+
+      expect(calculateActivityValue(transferIn)).toBe(500);
+    });
+
+    it("treats broker cash placeholders ($CASH-EUR, CASH-GBP, CASH_GBP) as cash and uses amount", () => {
+      const placeholders = ["$CASH-EUR", "CASH-GBP", "CASH_GBP", "$CASH_CAD"];
+      for (const symbol of placeholders) {
+        const transferIn = createActivity({
+          activityType: ActivityType.TRANSFER_IN,
+          assetSymbol: symbol,
+          assetId: symbol,
+          quantity: "0",
+          unitPrice: "0",
+          amount: "750",
+          fee: "0",
+        });
+        expect(calculateActivityValue(transferIn)).toBe(750);
+      }
+    });
+
+    it("preserves amount for securities transfers missing unitPrice (legacy imports)", () => {
+      const transferIn = createActivity({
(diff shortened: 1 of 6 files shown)


afadil:
I'm glad you're enjoying the app.

Thanks for the fix. I think this needs a couple of adjustments before merge.

1. `isSecuritiesTransfer()` currently treats any transfer that is not `isCashTransfer()` as a securities transfer. Normal cash transfers in this codebase often have no asset/symbol, and the backend also recognizes cash placeholders like `$CASH-EUR`, `CASH-GBP`, and `CASH_GBP`. Those rows now skip the amount branch and render as `quantity * unitPrice`, usually `0`. Please treat blank/no-asset transfers and all supported cash placeholders as cash, or pass asset identity into the classifier.

2. The backend now clears `amount` for every resolved non-cash transfer. Existing/imported transfer rows can have a real symbol plus quantity and amount but no unitPrice; import classification accepts that shape when quantity is present. Clearing amount there loses the only monetary value, and the read fallback cannot reconstruct it without unitPrice. Please preserve/fallback to amount when unitPrice is missing, or derive unitPrice from amount / quantity.

Regression tests that would cover this:
- cash `TRANSFER_IN/OUT` with empty assetSymbol uses amount
- `$CASH-EUR` / `CASH-GBP` cash transfer uses amount
- securities transfer with quantity + amount + no unitPrice preserves/displays amount

PR #868, opened 2026-04-19
chore(dev): add Claude Code hooks to run CI checks before commit/push

marcoscale98:
## Summary

Adds Claude Code pre-commit and pre-push hooks that mirror the CI pipeline locally, so failures are caught before reaching GitHub Actions.

- **`.claude/settings.json`**: registers two `PreToolUse` hooks on `git commit` and `git push`
- **`scripts/ci-check.sh`**: the hook script, mirrors `.github/workflows/pr-check.yml`

### Hook behavior

| Trigger | Checks run |
|---------|-----------|
| `git commit` | `cargo fmt`, `cargo clippy -D warnings`, `pnpm build:types`, `pnpm format:check`, `pnpm lint`, `pnpm type-check` |
| `git push` | All of the above + `cargo test --workspace` + `pnpm test` |

Checks are scoped: Rust checks only run if `.rs` files changed, frontend checks only if `.ts/.tsx/.js/.jsx/.css` files changed. If no relevant files changed, the hook exits immediately.

## Motivation

PR #843 failed CI due to a `clippy::unnecessary_sort_by` warning introduced by Rust 1.85. The local Rust version was older and didn't catch it. These hooks prevent that class of CI failure.

## Test plan

- [x] Stage a `.rs` file → commit blocked if `cargo clippy` fails
- [ ] Stage a `.ts` file → commit blocked if `pnpm lint` or `pnpm type-check` fails
- [ ] Stage only a `.md` file → no checks run, commit passes immediately
- [ ] `git push` → full checks including tests run before push

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,5 +1,31 @@
 {
   "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash scripts/ci-check.sh",
+            "if": "Bash(git commit*)",
+            "timeout": 120,
+            "statusMessage": "Running pre-commit CI checks..."
+          }
+        ]
+      },
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash scripts/ci-check.sh --full",
+            "if": "Bash(git push*)",
+            "timeout": 300,
+            "statusMessage": "Running full CI checks before push..."
+          }
+        ]
+      }
+    ],
     "PostToolUse": [
       {
         "matcher": "Edit|Write",
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,31 +1,5 @@
 {
   "hooks": {
-    "PreToolUse": [
-      {
-        "matcher": "Bash",
-        "hooks": [
-          {
-            "type": "command",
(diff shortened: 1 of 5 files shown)


afadil:
I pushed `9e3cd7b5` to adjust the approach here.

The main change is replacing the Claude-specific `PreToolUse` commit/push hooks with a tracked Git `pre-push` hook plus the shared `scripts/ci-check.sh` script.

Why this is simpler and more broadly useful:

- Git hooks use normal process exit codes, so we avoid Claude-specific JSON/blocking behavior.
- The hook is assistant-agnostic: it works the same whether the changes came from Claude, Codex, another assistant, an IDE, or manual edits.
- Once a contributor opts in with `pnpm hooks:install`, it runs for normal `git push` from any Git client that honors hooks, not only for Bash commands issued inside Claude Code.
- `scripts/ci-check.sh --full` now includes the CI build steps that were missing: `pnpm build` and `cargo build -p wealthfolio-server --release`.
- Full pre-push checks now run for any non-doc change, so build-affecting files like `Cargo.toml`, `Cargo.lock`, `package.json`, `pnpm-lock.yaml`, and workspace/config files do not slip through unchecked.

The only tradeoff is that Git intentionally does not auto-enable committed hooks for security reasons, so contributors need the one-time opt-in command documented in `CONTRIBUTING.md`.

PR #886, opened 2026-04-25
Fix CSV import asset review timeout for large Options imports

### LINKED ISSUE #884: CSV Import - Review asset timeout
v.3.2.1

When trying import a bunch of Options transactions at once, it seems there isn't enough time for the "Review Asset" process to do it's thing, and it times out and cancels. 

Can this timeout be increased please? Or be made configurable in settings?

afadil:
## Description

Fixes #884.

The "Review Assets" step in the CSV import wizard calls `preview_import_assets`, which resolves unique symbols against the market-data provider. For Options, each symbol generates 3–4 candidate variants (bare, currency-suffixed, exchange-suffixed) and backend symbol-resolution concurrency caps at 10. Large Options imports legitimately exceed the **global 2-minute Tauri invoke timeout** (`apps/frontend/src/adapters/tauri/core.ts`) and the 5-minute web `AbortSignal` timeout (`apps/frontend/src/adapters/web/core.ts`), causing the wizard to fail with "Asset preview failed".

### Root cause

- Tauri adapter wrapped every IPC call in `Promise.race(..., setTimeout(reject, 120_000))`.
- Web adapter used a single `AbortSignal.timeout(300_000)` for every HTTP call.
- Both safety nets fired on perfectly valid long-running batch operations, not just hung commands.

### Fix

Per-command timeout override map in both adapters. The 2 min / 5 min defaults remain in place as a safety net for unrelated commands; the two commands that legitimately do batched provider lookups get 10 minutes:

- `preview_import_assets`
- `check_activities_import` (runs the same `resolve_symbols_batch` per account)

`import_activities` itself does not call `resolve_symbols_batch` (assets are already resolved upstream by the preview step) and is left on the default. The AI-assistant import path uses the same shared adapter and inherits the override automatically.

### Why not "configurable in settings"?

The issue suggested making the timeout configurable. Rejected as over-engineering — a correctly-sized timeout shouldn't be a user preference, it just shouldn't trip on legitimate work.

### Backend left unchanged

`SYMBOL_RESOLVE_CONCURRENCY: usize = 10` in `crates/core/src/activities/activities_service.rs:707` is left as-is. Bumping it would help large imports finish faster but risks Yahoo Finance rate-limiting (HTTP 429) and warrants its own validation; tracking as a follow-up.

## Files changed

- `apps/frontend/src/adapters/tauri/core.ts` — default + override map applied to `Promise.race` setTimeout.
- `apps/frontend/src/adapters/web/core.ts` — same override map applied to `AbortSignal.timeout`.

## Test plan

- [x] `pnpm type-check` passes.
- [x] `pnpm lint` — no new warnings on edited files (the two warnings reported are pre-existing in unchanged code).
- [ ] Manual: import a CSV containing ~150+ unique Options transactions on desktop (`pnpm tauri dev`). The "Review Assets" step should finish within 10 minutes instead of failing at 2.
- [ ] Manual (web mode): same scenario via `pnpm run dev:web`. Browser should not abort the request.
- [ ] Smoke: an unrelated command that hangs should still reject at the 2 min (Tauri) / 5 min (web) default — the safety net is preserved.

## Checklist

### DIFF
--- a/apps/frontend/src/adapters/tauri/core.ts
+++ b/apps/frontend/src/adapters/tauri/core.ts
@@ -26,17 +26,26 @@ export const logger: Logger = {
   },
 };
 
-const INVOKE_TIMEOUT_MS = 120_000;
+const DEFAULT_INVOKE_TIMEOUT_MS = 120_000;
+
+// Commands that legitimately do batched network I/O over many symbols (Yahoo
+// Finance lookups during CSV import). Larger imports — especially Options —
+// can exceed the default 2-minute safety net. See issue #884.
+const INVOKE_TIMEOUT_OVERRIDES_MS: Record<string, number> = {
+  preview_import_assets: 600_000,
+  check_activities_import: 600_000,
+};
 
 /**
  * Invoke a Tauri command (internal - use typed adapter functions instead)
  */
 export const invoke = async <T>(command: string, payload?: Record<string, unknown>): Promise<T> => {
+  const timeoutMs = INVOKE_TIMEOUT_OVERRIDES_MS[command] ?? DEFAULT_INVOKE_TIMEOUT_MS;
   try {
     const result = await Promise.race([
       tauriInvoke<T>(command, payload),
       new Promise<never>((_, reject) =>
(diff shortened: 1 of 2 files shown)


afadil:
@codex review

PR #892, opened 2026-04-26
fix(activity): activity grid exchange-only asset edits not rebinding transactions

### LINKED ISSUE #891: Bug: Activity grid edit mode does not persist exchange-only symbol changes
On the Activity page in grid edit mode, changing a transaction’s symbol from one exchange-specific variant to another with the same base ticker does not persist correctly.

### Example:
- Original: `FBTC.NE`
- Edited to: `FBTC.TO`

### Observed behavior:

- Save reports success
- The transaction remains bound to the old asset (`FBTC` on `XNEO`)
- Reopening edit still shows the old exchange
- Downstream quote sync continues using the old Yahoo symbol / exchange mapping

### Expected behavior:

If the user changes exchange while keeping the same base symbol, the activity should be rebound to the correct asset for the new market identity

### Root cause:

- In Activity grid edit mode, change detection only compared the normalized `assetSymbol`
- `FBTC.NE` and `FBTC.TO` both normalize to `FBTC`
- Because original exchange was not included in asset identity change detection, the grid treated the asset as unchanged and sent the old asset ID in the update payload

### Notes:

- Single-transaction form edit did not reproduce this issue
- The bug was specific to the bulk/grid edit path

### Impact:

- Exchange-only edits in grid mode can silently keep the old asset binding
- Save UX is misleading because the update appears successful but does not reflect the intended asset switch

hao2hu:
Close #891 

## Description

This fixes an Activity page grid edit bug where exchange-only symbol updates were treated as unchanged asset identities.

### Example:

`FBTC.NE` -> `FBTC.TO`

### What changed:

- Grid edit state now tracks original exchange MIC alongside original symbol
- Grid save payload logic now treats exchange changes as asset identity changes, not just symbol text changes
- When symbol stays the same but exchange changes, the grid sends `symbol + exchangeMic` instead of reusing the old asset ID
- Added backend hardening so conflicting old asset ID vs submitted market identity is re-resolved safely
- Added regression tests for:
- frontend grid payload generation
- backend activity rebind behavior

### Why:

- The app normalizes provider tickers like `FBTC.NE` and `FBTC.TO` into:
   - symbol: `FBTC`
   - exchange: `XNEO` / `XTSE`
- Grid edit mode previously only compared the normalized symbol field, so exchange-only edits were missed

### Result:

- Activity grid edit mode now correctly persists same-symbol/different-exchange updates
- Transactions are rebound to the intended asset instead of silently staying on the old one

## Checklist

### DIFF
--- a/apps/frontend/src/pages/activity/components/activity-data-grid/activity-utils.test.ts
+++ b/apps/frontend/src/pages/activity/components/activity-data-grid/activity-utils.test.ts
@@ -369,6 +369,8 @@ describe("activity-utils", () => {
           isNew: false,
           assetSymbol: "VWRPL.XC",
           _originalAssetSymbol: "VWRPL.XC",
+          exchangeMic: "XLON",
+          _originalExchangeMic: "XLON",
           _originalAssetId: "asset-1",
           pendingQuoteCcy: "GBP",
           pendingInstrumentType: "EQUITY",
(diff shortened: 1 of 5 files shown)


hao2hu:
Close the PR as #907 covered the changes.

PR #939, opened 2026-05-05
Add dev:web:log script to tee server output to file

marcoscale98:
## Summary

- Adds `--file-log` flag to `scripts/dev-web.mjs`: derives log filename from `WF_DB_PATH` (or `DATABASE_URL` fallback), creates `logs/<db>.log`, and tees stdout+stderr of all child processes to it
- Adds `dev:web:log` npm script as convenience wrapper
- Excludes all hidden directories (`.*`) from Prettier checks in `.prettierignore`

## Test plan

- [x] Run `pnpm dev:web:log` and verify `logs/app-dev.log` (or matching your `WF_DB_PATH`) is created
- [x] Verify output appears both on screen and in the log file
- [x] Verify `pnpm dev:web` still works unchanged

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,10 +1,8 @@
 # Claude worktrees
 .claude/
 
-# Local tooling
-.air/
-.agents/
-.agent/
+# Local hidden directories
+.*
 
 # Build outputs
 dist/
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,11 @@
 # Claude worktrees
 .claude/
 
+# Local tooling
+.air/
+.agents/
+.agent/
+
 # Build outputs
 dist/
 coverage/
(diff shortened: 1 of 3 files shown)


afadil:
Wouldn't this do the trick simply?: `mkdir -p logs && pnpm dev:web 2>&1 | tee logs/app-dev.log`

is it for Windows?

marcoscale98:
The one-liner works great on Unix! The main reasons for the script approach:

1. **Dynamic filename** — the log name is derived from `WF_DB_PATH`, so when running tests with a specific database (e.g. `app-testing-20260506T123456Z.db`), the log is automatically named `logs/app-testing-20260506T123456Z.log` — making it easy to correlate logs with test runs. This isn't wired up in `pnpm test:e2e` yet, but it could be a natural next step.

2. **Cross-platform** — `tee` and `2>&1` aren't available on Windows `cmd.exe`.

Happy to simplify to the one-liner approach if Windows support isn't a concern for dev scripts, or to evolve this toward `test:e2e` integration if that direction makes sense to you!

PR #948, opened 2026-05-07
fix(connect): normalize OCC option symbols from broker activity sync

triantos:
## Summary

SnapTrade/Wealthfolio Connect occasionally returns option contract tickers
in the space-padded 21-character OCC form (e.g. `BA    260116C00200000`)
rather than the compact form (`BA260116C00200000`). The position-sync
path in `broker/service.rs:795` already calls `normalize_option_symbol()`
to collapse both representations into the canonical compact form, but the
activity-sync mapper in `broker/mapping.rs:381` passes
`option_symbol.ticker` through unchanged.

The result: the same option contract held in the same account can end up
split across two distinct asset records — one created by position sync
(compact), one by activity sync (padded) — fragmenting holdings, lots,
and per-asset performance.

## Changes

1. **`broker/mapping.rs`** — call existing `normalize_option_symbol`
   on `option_symbol.ticker` so both ingestion paths produce the
   canonical compact OCC form.

2. **New migration** (`2026-05-07-000001_dedupe_padded_option_symbols`)
   that cleans up pre-existing duplicates in two safe cases:
   - Padded option assets with no activities and no positions
     referencing them are deleted (the common orphan case, where
     activity-sync writes got deduplicated against earlier
     compact-symbol activities, leaving a dangling asset row).
   - Padded option assets with no compact equivalent are renamed
     in-place to compact form (the GENERATED `instrument_key` column
     auto-recomputes).

   The rare "both forms have separate activities" case is left untouched
   — merging activity history is error-prone enough to warrant a
   per-user manual repair.

## Test plan

- [x] \`cargo build -p wealthfolio-connect\` succeeds
- [x] \`cargo build -p wealthfolio-storage-sqlite\` succeeds
- [ ] After deploy, brokers that previously emitted padded OCC route
  activities to the same compact-symbol asset that position sync uses
- [ ] No regression on brokers already emitting compact OCC
  (\`normalize_option_symbol\` falls through with \`unwrap_or(t)\`)
- [ ] Migration: orphan padded option asset rows are removed; padded
  rows without a compact equivalent get their \`instrument_symbol\`
  rewritten in compact form

### DIFF
--- a/crates/connect/src/broker/mapping.rs
+++ b/crates/connect/src/broker/mapping.rs
@@ -374,12 +374,15 @@ pub fn map_broker_activity(
             })
     };
 
-    // Also get option symbol if present
+    // Also get option symbol if present. SnapTrade/Connect sometimes returns
+    // OCC tickers in space-padded form ("BA    260116C00200000"); normalize
+    // to compact form so we don't fragment asset identity per-broker.
     let option_symbol = activity
         .option_symbol
         .as_ref()
         .and_then(|s| s.ticker.clone())
-        .filter(|t| !t.trim().is_empty());
+        .filter(|t| !t.trim().is_empty())
+        .map(|t| wealthfolio_core::utils::occ_symbol::normalize_option_symbol(&t).unwrap_or(t));
     let is_option_activity = option_symbol.is_some() || option_leg_type.is_some();
     // Option contracts are uniquely identified by OCC ticker; adding underlying MIC can fragment identity.
     let exchange_mic = if is_option_activity {
--- a/crates/core/src/utils/occ_symbol.rs
+++ b/crates/core/src/utils/occ_symbol.rs
@@ -484,6 +484,14 @@ mod tests {
     }
 
     #[test]
+    fn test_normalize_padded_five_char_root() {
+        // Five-character roots leave only a single space of padding, so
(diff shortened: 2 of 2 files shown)


afadil:
Thanks for spotting this.

I'll add normalization on the backend Connect side as well, so SnapTrade padded OCC option tickers are normalized before reaching the app. I’m also keeping the app-side normalization as a final ingestion guard, since it is cheap, idempotent, and protects older Connect responses or future provider drift.

The safest scope for this PR is:

- normalize OCC symbols in backend Connect output
- keep app-side mapper normalization before asset lookup/create
- add/keep focused tests for padded OCC activity symbols, including 1-space roots like `GOOGL 260116C00200000`
- remove the automatic SQLite cleanup migration from this PR

The migration is riskier than the forward fix for two concrete reasons:

- The delete step runs before the rename step and does not require a compact duplicate to exist. A sole padded option asset with no activities or snapshot positions would be deleted instead of normalized, which can also cascade-delete related quotes/taxonomy data.
- The padded-symbol filter uses `LIKE '%  %'`, which misses valid 21-character OCC symbols with five-character roots and exactly one padding space, for example `GOOGL 260116C00200000`.

Given those risks, I’d keep this PR focused on preventing new bad writes and leave historical data repair for a separate, explicitly tested repair flow if we need it.

PR #952, opened 2026-05-08
fix(import): preserve TRANSFER_IN external flag through CSV import (#927)

### LINKED ISSUE #927: Transfer In "External" bit not preserved during import process
When importing a CSV containing `TRANSFER_IN` transactions, on the "Review Activities" page the "External" checkbox is correctly displayed as checked. However, after the import is completed, the "External" checkbox will be unchecked in the Activity page.

#### Steps To Reproduce
1. Import a CSV containing stock purchases
2. Map the transaction type for the purchases to `TRANSFER_IN` (assuming these are RSU grants, so shares are directly deposited)
3. On the "Review Assets" page, ensure "External" is checked by default.
4. Complete the import.
5. Go to the Activity page and view the new transactions.

#### Expected Behavior
New transactions should have the "External" bit set to true on the Activity page (matching what was shown on Step 3 above).

#### Actual Behavior
New transactions have the "External" bit set to false.

#### Environment
Version: Wealthfolio 3.3.0
OS: Windows 11 Pro (25H2)

afadil:
## Summary

Fixes #927 — the "External" checkbox on the Review Activities grid was lost during CSV import for `TRANSFER_IN` / `TRANSFER_OUT` rows. After completing the import, the flag was silently reset on the Activity page.

The flag was being dropped in three places in the request → persistence chain:

1. **Frontend** — `draftToActivityImport()` did not copy `isExternal` into the `ActivityImport` payload sent to Tauri.
2. **Backend struct** — `ActivityImport` had no `is_external` field; serde silently discarded the value even after the frontend started sending it.
3. **Backend conversion** — `From<ActivityImport> for NewActivity` hard-coded `metadata: None`, so the flag had nowhere to land.

This PR wires the flag end-to-end and persists it as `metadata.flow.is_external` on the activity, matching the path the manual transfer form already uses (`activity-form-config.ts:422`). Net contribution and flow classification now see CSV-imported external transfers the same way as manually-entered ones.

## Test plan
- [x] `cargo test -p wealthfolio-core --lib activities::` — 218 passed, 0 failed (incl. 2 new regression tests)
- [x] `cargo check --workspace --tests` — clean
- [x] `vitest run draft-utils.test.ts schemas.test.ts` — 18 passed
- [ ] Manual: import a CSV with TRANSFER_IN rows, confirm checkbox stays checked on the Activity page after import

### DIFF
--- a/apps/frontend/src/lib/schemas.ts
+++ b/apps/frontend/src/lib/schemas.ts
@@ -195,6 +195,9 @@ export const importActivitySchema = z
     fxRate: decimalLikeSchema.nullable().optional(),
     subtype: z.string().optional(),
     forceImport: z.boolean().default(false),
+    /** True when a TRANSFER_IN/OUT crosses the tracked-account boundary
+     * (e.g. RSU grant deposit). Persisted as `metadata.flow.is_external` on the activity. */
+    isExternal: z.boolean().optional(),
   })
   .refine(
     (data) => {
--- a/apps/frontend/src/pages/activity/import/utils/draft-utils.test.ts
+++ b/apps/frontend/src/pages/activity/import/utils/draft-utils.test.ts
@@ -112,6 +112,20 @@ describe("createDraftActivities explicit activity mapping", () => {
     expect(draft.amount).toBe("250.00");
   });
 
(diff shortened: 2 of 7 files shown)


afadil:
@codex review

PR #964, opened 2026-05-11
feat(portfolios): clean portfolio reporting scopes with relational schema

Jonjon-prog:
## Context

This is a clean restart of PR #938, following @afadil's "Portfolio Views and Account Groups Design" spec shared in that thread.

PR #938 is treated as a prototype. This PR addresses all P1 blockers identified in the review:
- ❌ JSON `account_ids` column → ✅ relational `portfolio_accounts` join table with FK constraints
- ❌ `MULTI:id,id` composite strings → ✅ typed `AccountFilter` enum resolved once at service boundary
- ❌ No device sync → ✅ both tables registered in `APP_SYNC_TABLES` with outbox events
- ❌ Web CRUD not wired → ✅ full Axum routes + web adapter parity
- ❌ Single-account portfolio loses identity → ✅ `{ type: "portfolio" }` preserved through UI and backend

## Schema

```sql
portfolios (id, name, description, sort_order, created_at, updated_at)
portfolio_accounts (id "pfm_{pfid}_{accid}", portfolio_id FK CASCADE, account_id FK CASCADE, sort_order)
```

## AccountFilter

```rust
enum AccountFilter { All, Account(String), Portfolio(String) }
```

Resolved once at the service boundary → repositories receive `&[String]` account IDs only.

## What ships

- Portfolio CRUD (settings page, Tauri commands, Axum routes, device sync)
- `AccountFilterSelector` component (All / Account / Portfolio)
- Income, Holdings, Insights pages: `AccountFilterSelector` replaces `AccountSelector` + `PORTFOLIO_ACCOUNT_ID` sentinel
- `accounts.group` untouched — kept as account organization metadata per spec
- Dashboard unchanged — no portfolio rows inserted into account group hierarchy

## Deferred (follow-up PR)

- Per-portfolio holdings/income aggregation: requires base-currency columns on `daily_account_valuation` as described in the "Currency Semantics" section of the design doc
- Performance page: TWR + transfer reclassification relative to portfolio member set
- Activity page: AccountFilter wiring (existing multi-select already handles account IDs)

## Migration note

The `up.sql` starts with `DROP TABLE IF EXISTS portfolio_accounts; DROP TABLE IF EXISTS portfolios;` as a dev-environment guard (for machines that ran PR #938 locally). On a fresh install these are no-ops. Happy to remove before merge if preferred.

## Test plan

- [ ] Create portfolio with 2+ accounts in Settings → Portfolios
- [ ] Edit portfolio (rename, change accounts)
- [ ] Delete portfolio → memberships cascade
- [ ] AccountFilterSelector visible on Income and Holdings pages
- [ ] Select specific account → filters correctly
- [ ] Web mode: all portfolio CRUD routes respond correctly
- [ ] Device sync: portfolio create/update/delete appears in outbox

## Checklist


afadil:
@codex review

afadil:
Thanks for the clean restart, the architecture is aligned with the target design now.

A few things to address before merge:

### Must fix

**1. Remove `DROP TABLE IF EXISTS` from `up.sql`**

```sql
-- crates/storage-sqlite/migrations/2026-05-11-000001_portfolios/up.sql
DROP TABLE IF EXISTS portfolio_accounts;
DROP TABLE IF EXISTS portfolios;
```

You already flagged this — please drop it. Migrations run once per machine; defensive drops in committed `up.sql` are an anti-pattern and will silently wipe data if any future migration or fork ever reuses these names. Local dev machines that ran #938 can clean up manually.

**2. The portfolio filter is a no-op — the picker silently lies**

In `holdings-page.tsx`, `income-page.tsx`, and `portfolio-insights.tsx`:

```ts
const accountId =
  accountFilter.type === "account" ? accountFilter.accountId : PORTFOLIO_ACCOUNT_ID;
```

When a user picks a portfolio, the result is identical to "All Accounts". I get that aggregation is deferred, but the UX has no signal. Before merge, either:
- disable portfolio rows in `AccountFilterSelector` with a "Coming soon" hint, or
- show a banner/empty-state on filtered pages when a portfolio is active.

Otherwise expect confused user reports.

**3. Duplicate-name error → generic toast**

The `idx_portfolios_name_unique` index rejects dupes with a SQLite UniqueViolation, which `usePortfolioMutations` surfaces as `"Failed to create portfolio."`. Either pre-check in the service and return a typed `ValidationError`, or map the constraint error to a friendly message.

### Should fix

**4. No tests.** The PR has manual-only verification. At minimum:
- service: name trim, empty-name reject, duplicate-account reject, missing-account reject, `resolve_account_filter` for each variant
- repository: regression around `update()` removing-then-reinserting memberships inside a tx

**5. Axum/Tauri parity gap.** `get_portfolio` exists as an Axum route but has no Tauri command. Either drop the route (UI only uses list + CRUD) or add the Tauri command.

**6. `useAccounts` filtering is inconsistent.** Settings page passes `filterActive: false` while `AccountFilterSelector` passes `filterActive: true`. If a portfolio references an inactive account, the selector hides it but the portfolio still has it as a member. Pick one.

### Nits

- `Portfolio` TS interface in `types.ts` is defined but unused — only `PortfolioWithAccounts` is consumed.
- `PortfolioDialog` relies on `key={editing?.id ?? "new"}` to remount/reset state. Works, but add a code comment so the key isn't accidentally removed.
- `PortfolioRepository::list()` filters memberships in O(P·M); a `HashMap<portfolio_id, Vec<_>>` is one line and avoids the quadratic. Not a real concern at expected scale.
- Unrelated Prettier reflow in `README.md`, `addons/community-addons.md`, `activity-delete-modal.tsx`, `alert-dialog.tsx` — worth isolating or reverting.

Jonjon-prog:
**Review addressed — summary**

Hi @afadil, thanks for the detailed review. Here's what was fixed in this pass:

**Must**
> - Removed `DROP TABLE IF EXISTS` from migration (`up.sql` is now clean — dev DB compat was a local workaround only)
> - `AccountFilterSelector`: portfolio items are now disabled with a "Coming soon" label — no silent no-op possible
> - Service maps `UniqueViolation` → `ValidationError("Portfolio name already exists")` with a user-friendly message

**Should**
> - Added `get_portfolio` Tauri command (parity with `GET /portfolios/{id}` on Axum)
> - `AccountFilterSelector` now uses `filterActive: false` (consistent with settings page)
> - Added 9 service unit tests covering: name trim, empty name, empty account list, duplicate account IDs, non-existent account, `UniqueViolation` mapping (create + update), `resolve_account_filter` (All + Account)

 **Nit**
> - `list()` uses `HashMap` grouping instead of O(P·M) nested filter
> - `key` prop on `PortfolioDialog` has a comment explaining the reset intent
> - Removed unused `Portfolio` TS type (everything uses `PortfolioWithAccounts`)
>

 **On the Prettier nit**: the 4 pre-existing files were already fixed in a prior commit — reverting them would break `pnpm format:check` in CI. Happy to isolate if you prefer a specific approach.

**Not in this PR (planned follow-up)**: wiring `AccountFilterSelector` + `AccountFilter` into Holdings / Income / Activity / Allocations pages — that's Step 14 and touches enough commands + pages to warrant its own PR. The "Coming soon" on portfolio items is intentional until that lands.

PR #967, opened 2026-05-12
refactor(asset-resolution): preserve provider-backed import identities

afadil:
## Summary
- preserve provider-backed asset identity metadata through symbol search, activity forms, holdings import, CSV asset preview, and backend asset resolution
- separate symbol search quote mode from provider data source so market assets without preferred provider metadata do not flip to manual pricing
- avoid duplicate provider re-resolution for reviewed import assets and pass provider refs into quote resolution
- restore currency-hinted unsuffixed import lookup by trying the primary Yahoo suffix before falling back to the raw symbol
- allow ISIN-only import rows to resolve through existing assets or provider search instead of short-circuiting

## Behavior Notes
- Plain MARKET assets no longer persist `preferred_provider: "YAHOO"` by default. Quote sync still falls back to Yahoo when no provider is stored.
- Region auto-classification no longer derives region from exchange MIC; provider country/countries data is required.
- Unsuffixed import symbols now use the row/account currency as a lookup hint for the primary exchange suffix, then fall back to the raw provider search symbol.

## Root Cause
Symbol search results were using `dataSource` as both provider provenance and quote-mode signal. Existing market assets without a preferred provider could therefore surface as manual, and reviewed import/provider metadata could be dropped before later resolution steps.

## Validation
- cargo test -p wealthfolio-core test_market_asset_without_provider_stays_market_in_search_summary
- cargo test -p wealthfolio-core test_preview_import_assets_preserves_provider_refs
- cargo test -p wealthfolio-core test_resolve_import_asset_inputs_
- pnpm --filter frontend exec vitest run src/pages/activity/components/forms/fields/__tests__/symbol-search.test.tsx
- pnpm --filter frontend exec vitest run src/pages/activity/import/utils/import-asset-rules.test.ts
- pnpm --filter frontend type-check
- pnpm --filter @wealthfolio/ui type-check
- cargo fmt --check
- git diff --check


afadil:
@codex review

PR #972, opened 2026-05-14
Data model refactor: materialized lots, unified valuation, relational snapshot positions

triantos:
24 commits, 89 files, +10.8k / -3.5k. Rebased onto v3.4.0 main.

## What this does

Three structural changes:

1. **Materialized `lots` table** — position state stored relationally
   instead of recomputed from activities on every read. BUY/SELL/SPLIT/
   TRANSFER maintain lot rows in FIFO order; closed lots persist for
   tax-lot reporting. Lots store `as-acquired quantity + split_ratio`
   so post-split lots reconstruct pre-split history correctly.

2. **Unified valuation pipeline** — `daily_portfolio_valuation`
   replaces the `TOTAL` pseudo-account rows in `daily_account_valuation`,
   and `alternative_market_value` is its own column so the Investments
   page can filter alt assets consistently. Net Worth routes
   negative-value categories to the liabilities section. Historical
   valuation replays lots to the as-of date.

3. **Relational `snapshot_positions`** — replaces the positions JSON
   blob in `holdings_snapshots` with a proper table. The JSON column
   stays additive (dual-write + JSON-fallback read) so older sync
   peers keep working.

Plus: liabilities are first-class \`LIABILITY\` assets, options link
to their underlying equity, a \`POST /assets/enrich\` admin endpoint,
a \`LotView\` UI, and assorted activity-import and metal-price-API
fixes.

## Design doc

\`docs/architecture/data_model.md\` (full narrative + diagrams).
\`docs/architecture/tax-engine-proposal.md\` sketches Phase D tax
support — out of scope for this PR.

## Migrations

Single consolidated migration: \`2026-04-14-000001_refactor_data_model\`.
Creates the new tables, adds two new columns to \`daily_account_valuation\`,
deletes \`TOTAL\` rows from it, and backfills \`snapshot_positions\` from
the legacy JSON. \`down.sql\` drops the schema additions but doesn't
restore data. The originally separate \`lots_split_ratio\` migration is
squashed in.

## Verification

All CI checks pass locally against \`upstream/main\`:

- \`cargo fmt --check\`, \`cargo clippy --all-targets --all-features -- -D warnings\`,
  \`cargo test --workspace\` (1659 tests), \`cargo check -p wealthfolio-server --release\`.
- \`pnpm install --frozen-lockfile\`, \`pnpm build:types\`, \`pnpm lint\`,
  \`pnpm type-check\`, \`pnpm test\` (607 tests), \`pnpm build\`.

[...]

PR #995, opened 2026-05-18
feat(portfolio-filters): wire AccountFilter into Holdings, Income, Allocations, Activity

Jonjon-prog:
## Summary

Hey @afadil , 

Here is the PR for the step 14 of portfolio view design. 

It wires the typed `AccountFilter` discriminant union into all relevant pages so users can filter by saved portfolio, individual account, or all accounts.

- **Holdings / Allocations**: replace `accountId: string` state with `AccountFilter`; use `AccountFilterSelector` component; backend resolves portfolio → account IDs and aggregates holdings in base currency
- **Income**: replace `accountId` filter with `AccountFilter`; backend resolves portfolio members before querying
- **Activity**: add portfolio preset dropdown (sets `selectedAccounts` from portfolio members — agreed design, full `AccountFilter` replacement deferred per Q6)
- **Mobile filter sheet**: add portfolios section, fix "All Portfolio" label → "All Accounts"
- **`DrillableAccountChart`**: now respects current filter (account/portfolio) instead of always showing all accounts
- **Rust IPC fix**: serde internally-tagged enums fail deserialization in Tauri v2 IPC; replaced `AccountFilter` enum as command parameter with `AccountFilterInput` flat struct at the IPC boundary only — internal logic still uses `AccountFilter` enum

## Backend changes

- `get_holdings`, `get_portfolio_allocations`, `get_holdings_by_allocation`, `get_income_summary` — now accept `AccountFilterInput` (Tauri) / `Json<FilterBody>` (Axum); resolve filter → account IDs via `portfolio_service.resolve_account_filter`
- `get_holdings_for_accounts` — new aggregation path: merges holdings by asset across accounts, sums `MonetaryValue { local, base }`, sets `account_id = ""`
- `get_portfolio_allocations_for_accounts`, `get_holdings_by_allocation_for_accounts` — same aggregation pattern for allocations

## Known issues / open questions for Afadil

1. `account_id` on aggregated holdings**

When Holdings are fetched for a portfolio (multi-account), the results are merged by `asset_id` across member accounts. Each merged holding needs an `account_id` but no single account owns it. Currently set to `""` (empty string). Two options:
- **Option A (current):** `account_id = ""` — neutral, signals "aggregated"
- **Option B:** `account_id = portfolio_id` — makes the source traceable in the UI

Keeping Option A for now. 
Would like your recommendation before changing, since this touches how the frontend displays or routes holdings.

**2. GET → POST on existing Axum routes**

To carry `AccountFilter` in the request body, four routes changed from GET to POST:
- `GET /holdings` → `POST /holdings`
- `GET /allocations` → `POST /allocations`
- `GET /allocations/holdings` → `POST /allocations/holdings`
- `GET /income/summary` → `POST /income/summary`

This is a breaking change for any existing web mode client. Flagging in case a cleaner approach (e.g. a separate `/account-filters/resolve` endpoint as mentioned in the design doc, or query params) is preferred.

**3. Regions allocation widget — values can exceed portfolio total**

Pre-existing bug, not introduced by this PR. When a holding is a global fund (e.g. World ETF), its value is distributed across multiple regions based on underlying exposure weights. The sum of all region values can therefore exceed the total portfolio value. Root cause is in the taxonomy allocation calculation, not in the AccountFilter wiring. 
Flagging for visibility.

## Test plan

- [x] `cargo test` — 1054 passed
- [x] `pnpm type-check` — no new errors
- [x] `pnpm test` — 624 passed (updated adapter parity test for POST)
- [x] Manual: Holdings All accounts ✓, individual account ✓, portfolio ✓
- [x] Manual: Income with portfolio filter ✓
- [x] Manual: Mobile filter sheet shows portfolios ✓
- [x] Manual: `DrillableAccountChart` respects filter ✓
- [x] Dashboard unchanged ✓
- [ ] Web mode (Axum) — not testable locally without `WF_SECRET_KEY` config

## Checklist

### DIFF
--- a/apps/frontend/src/adapters/adapter-command-parity.test.ts
+++ b/apps/frontend/src/adapters/adapter-command-parity.test.ts
@@ -97,12 +97,218 @@ describe("adapter command parity", () => {
     });
 
     const [url, init] = fetchMock.mock.calls[0] as [string, RequestInit];
-    expect(url).toBe("/api/v1/allocations/holdings");
+    expect(url).toBe("/api/v1/allocations/holdings/query");
     expect(init.method).toBe("POST");
     expect(JSON.parse(init.body as string)).toEqual({
       filter: { type: "all" },
       taxonomyId: "asset_classes",
       categoryId: "EQUITY",
+    });
+  });
+});
+
+// ─── Scope routing coverage ───────────────────────────────────────────────────
+
+function stubFetch(body: unknown = []) {
+  const res = new Response(JSON.stringify(body), {
+    status: 200,
+    headers: { "Content-Type": "application/json" },
+  });
+  const mock = vi.fn<typeof fetch>().mockResolvedValue(res);
+  vi.stubGlobal("fetch", mock);
+  return mock;
+}
+
+function lastCall(mock: ReturnType<typeof stubFetch>) {
+  const [url, init] = mock.mock.calls[0] as [string, RequestInit];
+  return { url, method: (init as RequestInit & { method: string }).method, body: init.body };
+}
+
+describe("scope-based routing — get_holdings", () => {
+  it("all → POST /holdings/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings", { filter: { type: "all" } });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/holdings/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({ filter: { type: "all" } });
+  });
+
+  it("account → GET /holdings?accountId=...", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings", { filter: { type: "account", accountId: "acc_1" } });
+    const { url, method } = lastCall(mock);
+    expect(url).toBe("/api/v1/holdings?accountId=acc_1");
+    expect(method).toBe("GET");
+  });
+
+  it("portfolio → POST /holdings/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings", { filter: { type: "portfolio", portfolioId: "pf_1" } });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/holdings/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({
+      filter: { type: "portfolio", portfolioId: "pf_1" },
+    });
+  });
+
+  it("accounts → POST /holdings/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings", {
+      filter: { type: "accounts", accountIds: ["acc_1", "acc_2"] },
+    });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/holdings/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({
+      filter: { type: "accounts", accountIds: ["acc_1", "acc_2"] },
+    });
+  });
+});
+
+describe("scope-based routing — get_portfolio_allocations", () => {
+  it("all → POST /allocations/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_portfolio_allocations", { filter: { type: "all" } });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/allocations/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({ filter: { type: "all" } });
+  });
+
+  it("account → GET /allocations?accountId=...", async () => {
+    const mock = stubFetch();
+    await invoke("get_portfolio_allocations", {
+      filter: { type: "account", accountId: "acc_1" },
+    });
+    const { url, method } = lastCall(mock);
+    expect(url).toBe("/api/v1/allocations?accountId=acc_1");
+    expect(method).toBe("GET");
+  });
+
+  it("portfolio → POST /allocations/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_portfolio_allocations", {
+      filter: { type: "portfolio", portfolioId: "pf_1" },
+    });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/allocations/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({
+      filter: { type: "portfolio", portfolioId: "pf_1" },
+    });
+  });
+
+  it("accounts → POST /allocations/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_portfolio_allocations", {
+      filter: { type: "accounts", accountIds: ["acc_1", "acc_2"] },
+    });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/allocations/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({
+      filter: { type: "accounts", accountIds: ["acc_1", "acc_2"] },
+    });
+  });
+});
+
+describe("scope-based routing — get_holdings_by_allocation", () => {
+  const drilldown = { taxonomyId: "asset_classes", categoryId: "EQUITY" };
+
+  it("all → POST /allocations/holdings/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings_by_allocation", { filter: { type: "all" }, ...drilldown });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/allocations/holdings/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({ filter: { type: "all" }, ...drilldown });
+  });
+
+  it("account → GET /allocations/holdings?...", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings_by_allocation", {
+      filter: { type: "account", accountId: "acc_1" },
+      ...drilldown,
+    });
+    const { url, method } = lastCall(mock);
+    expect(url).toBe(
+      "/api/v1/allocations/holdings?accountId=acc_1&taxonomyId=asset_classes&categoryId=EQUITY",
+    );
+    expect(method).toBe("GET");
+  });
+
+  it("portfolio → POST /allocations/holdings/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings_by_allocation", {
+      filter: { type: "portfolio", portfolioId: "pf_1" },
+      ...drilldown,
+    });
+    const { url, method, body } = lastCall(mock);
+    expect(url).toBe("/api/v1/allocations/holdings/query");
+    expect(method).toBe("POST");
+    expect(JSON.parse(body as string)).toEqual({
+      filter: { type: "portfolio", portfolioId: "pf_1" },
+      ...drilldown,
+    });
+  });
+
+  it("accounts → POST /allocations/holdings/query", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings_by_allocation", {
+      filter: { type: "accounts", accountIds: ["acc_1", "acc_2"] },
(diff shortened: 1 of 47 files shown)


afadil:
@Jonjon-prog Thanks for the PR.

A few issues i spotted with the help of codex:

1. **Allocation drilldown loses the selected portfolio filter**

   `AllocationDetailSheet` still calls `getHoldingsByAllocation({ type: "account", accountId }, ...)`, while `HoldingsInsightsPage` only passes `accountId` into the sheet. In a saved-portfolio view, the chart can be scoped to the portfolio, but the drilldown can fetch holdings for `TOTAL` / all accounts instead.

   Please pass the active `AccountFilter` into `AllocationDetailSheet`, include it in the query key, and call:

   ```ts
   getHoldingsByAllocation(accountFilter, taxonomyId, categoryId)
   ```

2. **Aggregated holdings keep percentage fields from the first account**

   `get_holdings_for_accounts` correctly sums quantity, market value, cost basis, gains, day change, and previous close value, but it does not recompute:

   - `unrealized_gain_pct`
   - `total_gain_pct`
   - `day_change_pct`

   If the same asset exists in multiple accounts, the aggregated row can show the first account’s percentages instead of portfolio-level percentages. After merging, please recompute percentages from the summed base values.

3. **Activity portfolio filter has split source-of-truth**

   Activity persists both `selectedPortfolioId` and expanded `selectedAccounts`. That can go stale if portfolio membership changes after selection. Prefer carrying the same `AccountFilter`/scope shape through Activity too, or derive account IDs from the selected portfolio at render/query time instead of persisting both.

4. **Tauri IPC filter parsing should fail closed**

   `AccountFilterInput::into_account_filter` currently maps unknown `type` values to `All`, and missing account/portfolio IDs to empty strings. That can silently broaden a malformed request to all accounts.

   Please return `Result<AccountFilter, String>` and reject:
   - unknown filter types
   - missing `accountId` for `account`
   - missing `portfolioId` for `portfolio`
   - missing/empty `accountIds` for ad-hoc/multi-account filters

## Design direction

The target design should be: **typed report queries with backend-owned scope resolution**.

`AccountFilter` should become the single reporting-scope contract from UI → adapter → backend. Longer term, I’d suggest naming this concept `AccountScope`, because it is not just a UI filter; it defines the reporting universe.

Example shape:

```ts
type AccountScope =
  | { type: "all" }
  | { type: "account"; accountId: string }
  | { type: "portfolio"; portfolioId: string }
  | { type: "accounts"; accountIds: string[] };
```

I’d avoid `adHoc` as a public/persisted API name because it describes where the selection came from, not what it means. `accounts` or `accountList` is clearer.

On the backend, the resolved form should avoid leaking fake account IDs like `"TOTAL"` through account paths:

```rust
enum ResolvedAccountScope {
    TotalSnapshot,
    Account(String),
    Accounts(Vec<String>),
}
```

That makes “use the precomputed total snapshot” explicit instead of treating `"TOTAL"` as if it were a real account ID.

For the `account_id` question: please don’t put `portfolio_id` into `Holding.account_id`. That overloads an account field with a non-account identity. Short term, empty string is less misleading. Better long term is explicit aggregate metadata, for example:

```ts
scope: AccountScope
sourceAccountIds: string[]
```

Then `accountId` stays meaningful only for single-account holdings.

## POST vs GET direction

I’m fine with POST for these complex report queries, but I would not replace the existing GET routes outright.

Best target:

```txt
GET  /api/v1/holdings?accountId=...                  legacy/simple
POST /api/v1/holdings/query                          typed scope query

GET  /api/v1/allocations?accountId=...                legacy/simple
POST /api/v1/allocations/query                       typed scope query

GET  /api/v1/allocations/holdings?...                 legacy/simple
POST /api/v1/allocations/holdings/query              typed scope query

GET  /api/v1/income/summary?accountId=...             legacy/simple
POST /api/v1/income/summary/query                    typed scope query
```

Use GET for simple identity/resource reads:

```txt
GET /holdings/item?accountId=...&assetId=...
GET /holdings/by-asset?assetId=...
GET /valuations/history?accountId=...&startDate=...&endDate=...
GET /snapshots?accountId=...
```

Use POST for computed report/search/query bodies:

```txt
POST /holdings/query
POST /allocations/query
POST /allocations/holdings/query
POST /income/summary/query
POST /performance/accounts/simple
POST /activities/search
```

This gives us room for future query fields without brittle query-param encoding:

```json
{
  "scope": { "type": "portfolio", "portfolioId": "pf_123" },
  "asOfDate": "2026-05-18",
  "includeCash": true,
  "baseCurrency": "USD"
}
```

Please also add adapter/server coverage for `all`, `account`, `portfolio`, and multi-account/ad-hoc scopes, especially allocation drilldown.

Jonjon-prog:
All review items addressed in the latest push. One remaining known issue to flag:

**Regions allocation widget — values can exceed portfolio total**

In the Regions/Country allocation chart, when viewing an aggregated scope (portfolio or multi-account), some category percentages appear to exceed 100% of the portfolio. This seems to be a pre-existing issue with how region weights are computed across aggregated holdings — likely the weights are summed from individual account snapshots rather than recomputed against the merged portfolio total.

Not introduced by this PR, but visible when using portfolio scope. Happy to investigate further if you can point to the relevant allocation weight computation.

afadil:
**Design Direction**

The target design is still: `AccountFilter` / `AccountScope` is the single reporting scope contract from UI to adapter to backend. Keep portfolio membership resolution backend-owned. Frontend expansion is acceptable only for display widgets that truly need account ids, and those widgets must not widen to all while loading.

For this PR, I would scope the merge to holdings, allocations, allocation drilldown, activity filtering, and income once the bugs bellow are fixed. 

Do not claim portfolio-scoped performance/valuation support until the backend has a proper scoped valuation and cashflow engine. The lots/snapshot-position work from PR #972 is a better long-term foundation, but this PR should either stay within the current safe scope.


**Findings To Fix**

1. **[P1] Web POST bodies for portfolio and multi-account scopes do not deserialize.**  
   Frontend sends camelCase fields like `portfolioId` / `accountIds`, but the server POST DTOs deserialize directly into core `AccountScope`, which expects `portfolio_id` / `account_ids`. This means web POST filtering fails before reaching the service layer. Fix with a web DTO equivalent to the Tauri input DTO, or add serde aliases/renames consistently. Cover holdings, allocations, allocation drilldown, and income.

2. **[P1] Scope-aware performance and valuation history are not implemented.**  
   The PR wires holdings/allocations/income, but historical valuations and performance still use single `accountId` / `TOTAL` style paths. For portfolios or multi-account scopes, correct performance requires backend-owned scoped daily valuations plus cashflow boundary classification: transfers inside the scope are internal, transfers crossing the scope boundary are deposits/withdrawals.

3. **[P2] Aggregated holdings still write portfolio id into `accountId`.**  
   Portfolio aggregates currently put the portfolio id into `Holding.account_id`. That overloads an account field with a non-account identity. Short term, use an empty string for aggregate rows. Longer term, add explicit scope metadata such as `scope: AccountFilter` or `sourceAccountIds`; keep `accountId` meaningful only for single-account holdings.

4. **[P2] Legacy web `GET /income/summary` now requires `accountId`.**  
   The old no-`accountId` all-account behavior appears broken because the query DTO requires a string. New frontend may use POST for all-scope, but existing clients/addons could regress. Make `accountId` optional and preserve all-account behavior when absent.

5. **[P2] Mobile income filter still disables portfolio selection.**  
   Desktop income supports the new account scope selector, but the mobile filter sheet still marks portfolios as “Coming soon”. That leaves inconsistent behavior across responsive layouts.

6. **[P2] Frontend portfolio resolution can temporarily widen scope to all.**  
   Some display/filter paths resolve portfolio account ids in the frontend. When portfolios are still loading or missing, the derived account list can become `undefined`, which downstream code treats as no filter. Gate rendering/loading or return an empty closed scope until portfolio membership is known.

7. **[P3] Remove leftover old scope naming if the contract is now `accounts`.**  
   The Tauri parser still accepts the old multi-account alias and the error text still mentions it. If the public contract is `accounts`, keep the compatibility path only if intentionally needed, but remove stale wording from errors/tests/docs.

8. **[P3] Income repository builds manual SQL for account-id lists.**  
   The current string-built `IN (...)` clause is escaped, but it is still a regression from bound query patterns and is easy to maintain incorrectly. Prefer Diesel `eq_any`, a chunked bound query, or a shared helper for list binding.

9. **[P3] Test coverage needs to hit the actual server boundary and aggregate math.**  
   Adapter parity tests only verify URLs and request bodies. Add server/handler tests with real JSON bodies for all, single account, portfolio, and multi-account scopes. Add core tests for aggregate holdings: same asset across accounts, cash by currency, zero cost basis, missing quote/cost basis, `sourceAccountIds`, and aggregate `accountId` behavior.

afadil:
Also, I double-checked the Regions allocation issue against the PR base, and I agree this looks pre-existing rather than introduced by this PR.

One clarification: I don’t think the root cause is summed account allocation snapshots. The portfolio/multi-account path merges holdings first, then reuses the existing `aggregate_by_taxonomy` logic. The risky part is the existing taxonomy math: asset-level region assignment weights are applied directly to holding market value, but the service does not normalize or reject cases where an asset’s assignments for a taxonomy sum above 100%, or where overlapping parent/child region assignments are both counted.

So I’m fine tracking this as follow-up, but let’s describe it accurately as allocation weight normalization / validation debt, not an AccountScope aggregation bug.

Suggested follow-up:
- normalize or validate assignment weights per asset/taxonomy;
- guard against overlapping parent/child region assignments being double-counted;
- add a regression test where one asset has region weights above 100% and verify portfolio percentages do not exceed the scoped total.

Jonjon-prog:
Hey @afadil 

Thanks for the thorough review. 


---

**P1a — Fixed.**
`AccountScope` had `rename_all = "camelCase"` on the enum container, which renames variant names but not struct variant fields. Added `#[serde(rename_all = "camelCase")]` to each struct variant (`Account`, `Portfolio`, `Accounts`) so `portfolioId`/`accountIds`/`accountId` deserialize correctly in all web POST bodies (holdings, allocations, allocation drilldown, income).

**P1b — Confirmed out of scope, no code claiming support.**
Performance and valuation history are not wired to `AccountScope` in this PR. The performance page uses its own `TrackedItem` system, the holdings insights page has no valuation history chart, and the dashboard doesn't use account scope for valuations. No code in this PR claims portfolio-scoped performance support — agreed this belongs to a future PR with proper scoped daily valuations and cashflow boundary classification (the PR #972 foundation you mention).

**P2a — Fixed.**
`aggregated_id()` in the server handler was returning `portfolio_id` for portfolio scope, writing a non-account identity into `Holding.account_id`. Fixed to always return `""` for aggregated rows. Same fix applied to all three Tauri handlers (holdings, allocations, allocation drilldown).

**P2b — Fixed.**
`accountId` is now `Option<String>` in the `GET /income/summary` query DTO. When absent, falls back to all-accounts behavior, preserving the original semantics for existing clients and addons.

**P2c — Fixed.**
Removed the "Coming soon" placeholder in the mobile income filter sheet. Portfolio rows are now clickable with active state, matching desktop behavior.

**P2d — Fixed.**
`filteredAccountIds` in `HoldingsInsightsPage` now returns `[]` instead of `undefined` when portfolio scope is selected but membership is not yet resolved (portfolios still loading or portfolio not found). Display widgets receive a closed empty scope rather than widening to all accounts.

**P3a — Fixed.**
Error message in the Tauri parser for the `adHoc | accounts` branch now cites the canonical name `"accounts"`. The `"adHoc"` alias is kept intentionally for backwards compatibility.

**P3b — Acknowledged, tracked as follow-up.**
The manual `IN (...)` SQL in the income repository is a known pattern regression. Refactoring requires rewriting a complex raw query (multi-join, CASE-based income calculation) into diesel's query builder. The current escaping (`replace('\'', "''")`) is correct since account IDs are internal UUIDs, so this is safe to track separately without blocking this PR.

**Allocation weight normalization — agreed on description.**
Updated understanding: the issue is allocation weight normalization/validation debt in `aggregate_by_taxonomy` (weights not normalized when an asset's assignments for a taxonomy sum above 100%, overlapping parent/child region assignments double-counted), not an `AccountScope` aggregation bug. Tracking as follow-up with the three guard points you suggested.

PR #856, opened 2026-04-16
Data model refactor: materialized lots, unified valuation, relational snapshot positions

triantos:
Consolidates the three previously-open data model PRs (#792, #831, #842)
into a single branch for easier review and merge. Closing those PRs in
favor of this one.

## What this does

Three structural changes plus the associated rewiring:

1. **Materialized `lots` table** — position state is now stored relationally
   instead of recomputed from activities on every read. BUY/SELL/SPLIT/
   TRANSFER maintain lot rows in FIFO order; closed lots persist for
   tax-lot reporting; historical point-in-time quantities are derived via
   activity replay against each lot's `original_quantity`.

2. **Unified valuation pipeline** — `daily_portfolio_valuation` replaces
   the `TOTAL` pseudo-account rows in `daily_account_valuation`, and
   `alternative_market_value` is surfaced as its own column so the
   Investments page can filter alt assets consistently between header
   and account cards. Net Worth routes negative-value categories to the
   liabilities section instead of silently dropping them.

3. **Relational `snapshot_positions`** — replaces the positions JSON blob
   in `holdings_snapshots` with a proper table. The JSON column stays in
   place as vestigial `'{}'` for sync-compatibility; nothing reads it.

Plus: liabilities are first-class `LIABILITY` assets with an optional
account link; the rich `LotView` display type (open/closed badges,
per-account grouping); a mutex serializing concurrent portfolio recalcs;
and the Tauri `get_asset_lots` command registration.

## Design doc

`docs/architecture/data_model.md` — narrates all of the above, with four
SVG diagrams (before/after/with-tax/rules-engine). Also proposes a
multi-jurisdiction tax-tagging surface (Phase D) that's out of scope for
this PR but explains where the refactor is heading.

## Migrations

Everything goes through a single migration:
`2026-04-14-000001_refactor_data_model`. Creates the new tables, adds
the two new columns, deletes `TOTAL` rows from `daily_account_valuation`,
and backfills `snapshot_positions` from the legacy JSON (one-shot atomic
INSERT). The `down.sql` drops the schema additions but does not restore
data.

## Commit layout

11 thematic commits, each compilable and passing CI on its own:

- `docs: add data model refactor design and diagrams`
- `migration: consolidated data model refactor`
- `feat(lots): materialize tax lots with FIFO consumption`
- `feat(valuation): unified portfolio valuation and balance-sheet routing`
- `fix(lots): robustness fixes in the lot-writing path`
- `refactor: stop reading positions JSON from holdings_snapshots`
[...]

PR #1002, opened 2026-05-20
Materialize lots and snapshot_positions tables (additive groundwork)

afadil:
## Summary

Extracts the `lots` + `snapshot_positions` groundwork from [#972](https://github.com/wealthfolio/wealthfolio/pull/972), with the simplified direction from review:

- **`lots`** stores only real transaction-derived tax lots. One row per acquisition or transferred sub-lot, updated as shares are disposed. Quantities/costs stay in as-acquired units; `split_ratio` carries post-acquisition split effects.
- **`snapshot_positions`** materializes per-snapshot position rows beside the existing `holdings_snapshots.positions` JSON.
- **`AssetLotView`** is the UI/read model for the asset Lots tab. It combines persisted transaction lots with latest HOLDINGS-mode snapshot positions as aggregate rows when requested.

HOLDINGS-mode accounts are snapshot tracking, not tax-lot tracking, so this PR no longer synthesizes HOLDINGS rows into the `lots` table and no longer needs first-launch backfill logic.

## Current Design

- TRANSACTIONS-mode accounts write `lots` during normal holdings recalculation from the calculator's final in-memory lot state.
- HOLDINGS-mode accounts stay out of `lots`; the asset lot endpoint can append their latest snapshot positions as `SnapshotPosition` rows.
- The frontend asset Lots tab loads the combined view with `includeSnapshotPositions = true`.
- `snapshot_positions` is dual-written with snapshot JSON, and repository reads have JSON fallback for older snapshots.
- The migration creates the two relational tables, preserves non-calculated/manual snapshots, clears derived `CALCULATED` snapshots and `daily_account_valuation`, and lets the normal portfolio calculation rebuild derived state.
- The first post-migration rebuild uses the existing startup update path: `useGlobalEventListener` calls `updatePortfolio()` after listeners mount; desktop emits `portfolio:trigger-update`, web posts `/portfolio/update`, and both enqueue the normal portfolio job. Because the migration clears derived rows, that job rebuilds calculated snapshots, transaction lots, and valuations without a separate backfill hook.

## Cleanup From Review

- Removed the stale first-launch/HOLDINGS-lot backfill direction from implementation and docs.
- Removed unused staged Phase B code (`replay_lots_to_date`) and its ignored known-bug test block.
- Removed unused split-ratio backfill helper.
- Removed stray `[LOT-DEBUG]` logging in FIFO reduction.
- Parameterized the sync replay `snapshot_positions` delete instead of interpolating SQL.
- Simplified `includeSnapshotPositions` query handling to a defaulted bool.
- Aligned Diesel `lots` schema/model column order with the migration.
- Added a service-level regression test proving lot sync failures are surfaced as errors.
- Adjusted SPLIT lot eligibility to compare user-local calendar dates consistently.

## Verification

- `cargo test -p wealthfolio-core lots`
- `cargo test -p wealthfolio-core test_recalculate_holdings_snapshots_surfaces_lot_sync_errors`
- `cargo test -p wealthfolio-core test_split_uses_user_local_calendar_date_for_lot_eligibility`
- `cargo test -p wealthfolio-core test_since_date_recalc_clears_snapshots_and_lots_when_last_activity_deleted`
- `cargo test -p wealthfolio-storage-sqlite lots`
- `cargo test -p wealthfolio-storage-sqlite replay_snapshot_clears_stale_snapshot_positions`
- `cargo check -p wealthfolio-server`
- `cargo check -p wealthfolio-app`
- `pnpm format:check`
- `git diff --check`

## Deferred

- Switching valuation/holdings read paths fully off JSON snapshots.
- Tax overlays such as disposal method, wash sale, and holding period.
- Larger data-model items from #972 that are unrelated to this additive lot/snapshot-position groundwork.


afadil:
Thanks @chatgpt-codex-connector — verified against my test DB: 1 orphan `assetId` in snapshot JSON did produce 2 orphan `snapshot_positions` rows on the original migration. The actual failure mode is slightly different from what the bot stated (FK is OFF during diesel migrations, so the INSERT didn't abort), but the underlying concern is right: orphans silently land in the relational table at migration time, then break the **runtime** dual-write path at `write_snapshot_positions`' DELETE+INSERT pair when the user later edits that snapshot.

Fixed in [f5ec2ee49](https://github.com/wealthfolio/wealthfolio/commit/f5ec2ee49) at three layers:

1. **Migration backfill** — added `AND EXISTS (SELECT 1 FROM assets WHERE id = ...)` to the INSERT.
2. **`SnapshotRepository::write_snapshot_positions`** — pre-query existing `asset_id`s, drop missing-asset positions with a warn log.
3. **`LotsRepository::replace_lots_for_account` + `sync_lots_for_account`** — same filter for the lots write paths (HOLDINGS-mode derivation has the same risk).

Verified against the test DB: drift count stays 0, all workspace tests pass.

afadil:
All three P2 findings addressed in [5aefd9061](https://github.com/wealthfolio/wealthfolio/commit/5aefd9061):

**1. Full recalcs not populating `snapshot_positions`** — confirmed. `overwrite_all_snapshots_for_account` used raw `diesel::replace_into` and never called `write_snapshot_positions`. The other two overwrite methods go through `save_snapshots` which was already fixed. Mirrored `save_snapshots`' capture-then-write pattern in `overwrite_all`.

**2. Closure `split_ratio = "1"`** — confirmed. Added `split_ratio: String` to `LotClosure`, `record_lot_closure` now captures `lot.effective_split_ratio()`, and `sync_lots_for_account` writes `closure.split_ratio` instead of the hardcoded `"1"`.

**3. Delete returning Ok on lot-refresh failure** — confirmed. `refresh_lots_from_latest_snapshot` now propagates errors via `.map_err(|e| { error!(...); e })?` — keeps the diagnostic log line but bubbles the failure up to `delete_snapshot_for_account` and on to the Tauri/server handler. Same fix applied to `save_manual_snapshot`' lot write, which had the same log-and-swallow pattern.

Verification: 1062 core / 248 storage / 121 connect / 50 device-sync tests pass. `cargo clippy --tests -- -D warnings` clean. The two tests you referenced (`replay_split_updates_split_ratio_not_quantity` exists at `crates/core/src/lots/mod.rs:1362` and is in the `tests` module filter; `sync_with_empty_inputs_does_not_wipe_existing_lots` at `crates/storage-sqlite/src/lots.rs:661`) both pass.

afadil:
All four findings addressed in [042b42640](https://github.com/wealthfolio/wealthfolio/commit/042b42640):

**P1 — original_cost_basis / fee_allocated corrupted after partial sells.** Confirmed. `reduce_lots_fifo` mutates `Lot.acquisition_fees` to track remaining fee, but `extract_lot_records` and `record_lot_closure` treated it as immutable. A $10-fee, half-sold lot persisted `fee_allocated=5`, `original_cost_basis = price*qty + 5`. Fixed by adding an immutable `Lot.original_acquisition_fees` field paired with the existing `original_quantity`, populated at every lot construction site (add_lot, add_lot_values, transfer-in, removed-lot copies). `Lot::original_fees()` accessor falls back to `acquisition_fees` for pre-this-field snapshots. Regression test `extract_lot_records_preserves_original_fees_after_partial_sell` locks the contract.

**P1 — DRIP/staking synthetic activity IDs.** Confirmed. `DefaultActivityCompiler` emits `{parent}:buy` and `{parent}:{income_suffix}` for asset-income expansions; `handle_buy` stored these as `source_activity_id` and the FK on `lots.open_activity_id` rejected them at runtime. Storage-layer fix per your suggestion: `filter_and_normalize_lots` validates both `open_activity_id` and `close_activity_id` against `activities.id` and NULLs them out with a warn log when missing. New storage test `synthetic_open_activity_id_is_nulled_out` exercises the `drip-1:buy` case and asserts the lot persists with `open_activity_id IS NULL`.

**P2 — Full recalc + no activities left stale lots.** Confirmed. Now clears lots for every account in `ids_to_delete` inside the early-return branch and **propagates** the error per your suggestion — if we just deleted snapshots and can't keep lots in sync, the caller must know rather than warn-and-pretend-Ok.

**P2 (latent) — `replay_lots_to_date` and transferred split-adjusted lots.** Confirmed but not called by any read path in this PR. Documented as a `# Known limitation` block on the function doc with an example (inherited 2 × in-account 3 = 6, but the binary reset_split produces 1 or 3 depending on as-of). Locked in with `#[ignore]`'d test `replay_lots_to_date_transferred_split_adjusted_lot_known_bug` — runnable via `cargo test ... -- --ignored`. Verified it actually fails against the current implementation ("6" returned, "2" expected at as_of=2022).

Verification:
- `cargo test --workspace`: 1063 core (+1 new) / 122 storage (+1 new) / 248 connect / 50 device-sync / others — all pass
- `cargo clippy -p wealthfolio-core -p wealthfolio-storage-sqlite -p wealthfolio-server --tests -- -D warnings` clean
- `cargo fmt --all` applied

Tests you suggested for verification — `lots::tests::replay_split_updates_split_ratio_not_quantity` (`crates/core/src/lots/mod.rs:1438`) and `lots::tests::sync_with_empty_inputs_does_not_wipe_existing_lots` (`crates/storage-sqlite/src/lots.rs:751`) — both still pass.

afadil:
All three findings addressed in [95a8c5ce3](https://github.com/wealthfolio/wealthfolio/commit/95a8c5ce3).

**P1 — Normal activity changes don't update lots.** Confirmed. Planner emits `SnapshotRecalcMode::SinceDate` for every activity add/edit/delete; my hook's `skip_for_transactions` gate then blocked the lot write on every non-Full mode for TRANSACTIONS accounts. Dropped the skip entirely. The original concern (pre-PR snapshots with baked-in split quantities) is handled by the startup `backfill_lots_if_needed` which does a Full recalc on first launch — after that every snapshot is in the new semantics, so incremental modes seed from known-good state.

**P1 — Lot upserts leave origin fields stale after edits.** Confirmed. Open-lot UPSERT was missing `asset_id`, `account_id`, `open_date`, `open_activity_id`, `fee_allocated`. Closure UPSERT was missing all of those plus `original_quantity`, `cost_per_unit`, `original_cost_basis`, `split_ratio`. Now both SET clauses refresh every replay-derived column; only `id` (PK) and `created_at` (audit anchor) are preserved across conflicts. Added two regression tests:
  - `sync_upsert_refreshes_origin_fields_on_open_lot_edit` — edits asset, date, activity_id, fee, price; verifies all propagate
  - `sync_upsert_refreshes_origin_fields_on_closure` — same on a fully-consumed lot

**P2 — App sync makes snapshot_positions stale.** Confirmed. Generic `INSERT ON CONFLICT DO UPDATE` updates the JSON column in place without firing the FK CASCADE; the receiving device's relational rows kept their pre-sync state. `get_snapshot_positions` then returned those stale rows because it only falls back to JSON when no rows exist at all. Added a hook in `apply_remote_event_lww_tx` that runs `DELETE FROM snapshot_positions WHERE snapshot_id = ?` after a successful `SyncEntity::Snapshot` UPSERT. The next read falls back to the freshly-synced JSON. Heavier "parse JSON and rebuild relational rows here" is left for the Phase B read-path switchover. Regression test `replay_snapshot_clears_stale_snapshot_positions` proves the hook fires.

Verification: 1063 core / 125 storage (+3 new) / 248 connect / 50 device-sync / others — all pass. `cargo clippy -p wealthfolio-core -p wealthfolio-storage-sqlite -p wealthfolio-server --tests -- -D warnings` clean. `cargo fmt --all` applied.

afadil:
Heads up: your reference says "Latest PR head is still 042b4264..." — but the head is now [95a8c5ce3](https://github.com/wealthfolio/wealthfolio/commit/95a8c5ce3) (and [1a1477a20](https://github.com/wealthfolio/wealthfolio/commit/1a1477a20) after this latest commit). The three P1/P1/P2 you listed are all already addressed there:

| Finding | Status at 95a8c5ce3 |
|---|---|
| P1 — normal activity changes don't update lots | **Fixed.** `skip_for_transactions` gate removed in `snapshot_service.rs`. `grep skip_for_transactions` on head returns nothing. |
| P1 — lot upserts preserve stale facts | **Fixed.** Both open-lot and closure UPSERTs now refresh `account_id`, `asset_id`, `open_date`, `open_activity_id`, `fee_allocated` (plus split_ratio, original_quantity, cost_per_unit, original_cost_basis on closures). Verified by `grep account_id.eq.*excluded crates/storage-sqlite/src/lots.rs` — appears in both UPSERT branches. Regression tests: `sync_upsert_refreshes_origin_fields_on_open_lot_edit`, `sync_upsert_refreshes_origin_fields_on_closure`. |
| P2 — app sync makes snapshot_positions stale | **Fixed.** `apply_remote_event_lww_tx` runs `DELETE FROM snapshot_positions WHERE snapshot_id = ?` after the generic UPSERT for `SyncEntity::Snapshot`. Regression test: `replay_snapshot_clears_stale_snapshot_positions`. |

I'd suggest re-pulling and re-reviewing at the current head.

---

The fourth concern is genuinely new and now fixed in [1a1477a20](https://github.com/wealthfolio/wealthfolio/commit/1a1477a20):

**P2 edge — Full recalc with zero-output for an account.** Confirmed. When a Full recalc produces empty `open_lots` + empty `closures` + empty `snapshot.positions` for an account whose activities are all non-lot-producing (DEPOSIT / DIVIDEND / INTEREST), the calculator is authoritatively saying "this account holds nothing" — but `sync_lots_for_account`'s safety guard preserves the stale lots from a since-deleted BUY.

Fix at the snapshot_service hook: when (Full mode AND open_lots empty AND closures empty AND positions empty), call `replace_lots_for_account(&[])` explicitly instead of `sync_lots_for_account`. The wipe is gated on Full mode because incremental modes seed from a prior snapshot's positions and might legitimately produce no NEW lots while carrying historical ones — the safety stays for those paths.

`cargo test --workspace` all passing. `cargo clippy --tests -- -D warnings` clean.

afadil:
All four findings addressed in [2ca45a039](https://github.com/wealthfolio/wealthfolio/commit/2ca45a039):

**P1 — HOLDINGS-mode saves wiping lots.** Confirmed and the worst of the bunch. `manual_snapshot_service` builds Positions with `lots: VecDeque::new()`; the lot hook called `extract_lot_records` which iterated those (empty) in-memory lots and returned `[]`; then `replace_lots_for_account(account_id, &[])` wiped every synthetic HOLDINGS lot. Every manual / CSV / broker save on the latest snapshot was erasing the lots.

Fix: extracted the synthetic-lot derivation (one HOLDINGS-{acct}-{asset} per position, copying aggregate qty/cost/cost_basis) into `synthetic_holdings_lot_records`. Used by all three call sites — backfill, refresh, **and the save_manual_snapshot lot hook**. PR-972's design intent was for the third site to also use synthetic derivation; it just wasn't wired correctly.

**P2 — Sync replay leaves HOLDINGS lots stale.** Confirmed (same gap exists in PR-972). In `apply_remote_event_lww_tx` for `SyncEntity::Snapshot`:
- On Update: parse `accountId` from payload, then `DELETE FROM lots WHERE account_id = ? AND open_activity_id IS NULL`
- On Delete: pre-query the snapshot's account_id BEFORE the DELETE, then run the same lot DELETE after

Lots get repopulated on the next local save_manual_snapshot. A full re-derive from the new latest snapshot is deferred to Phase B's read-path switchover (TODO comments mark both sites). Regression tests:
- `replay_snapshot_update_clears_stale_holdings_lots`
- `replay_snapshot_delete_clears_stale_holdings_lots`

**P2 — Lot sync errors silent.** Confirmed. Now accumulate per-account error messages into a `Vec<String>` during the recalc loop. After the loop, return `Err` with the concatenated list if any account failed. Trade-off chosen: complete the loop so every other account still gets its snapshots saved, but surface the divergence to the caller. The caller can decide to log, retry, or trigger a Full recalc.

**P3 — `get_open_position_quantities` undercounted.** Confirmed. Selected only `remaining_quantity` and summed it raw, but quantities are stored in as-acquired (pre-split) units. Now selects `split_ratio` too and sums `remaining_quantity * split_ratio`, treating a zero / unparseable ratio as 1.0 to match `Lot::effective_split_ratio()`. No production caller in this PR yet — Phase B will use it — but locking the contract in now. Regression test: `get_open_position_quantities_applies_split_ratio`.

Verification:
- `cargo test --workspace`: 1063 core / 128 storage (+3 new) / 248 connect / 50 device-sync / others — all pass
- `cargo clippy -p wealthfolio-core -p wealthfolio-storage-sqlite -p wealthfolio-server --tests -- -D warnings` clean
- `cargo fmt --all` applied

afadil:
Bundled a pre-existing P1 bug fix into this PR: [fe96f0e7b](https://github.com/wealthfolio/wealthfolio/commit/fe96f0e7b).

**Bug:** `prepare_activities_for_import` was defined in `activities_service.rs` but **never called** in production code. The save path correctly calls `prepare_activities_for_save`, the sync path correctly calls `prepare_activities_for_sync` (via `broker/service.rs:347`), but the `POST /api/v1/activities/import` path was missing the equivalent call. Every imported activity landed with `asset_id = NULL`, silently breaking lots / positions / holdings derivation for the entire imported batch.

Bug exists in current `main` independent of this PR's work — caught by Nick (`@triantos`) in PR #856 (commit [87b373ae](https://github.com/wealthfolio/wealthfolio/pull/972/commits/87b373aefa11f5777ad7760d6433ff0dd09dc589)) and patched by the rebase follow-up [a2357a76](https://github.com/wealthfolio/wealthfolio/pull/972/commits/a2357a763d32b57c0b685beea579dd0ae4dca7b4) to preserve `idempotency_key` against force-import's deliberate `None`. Ported both into this PR.

**Why bundling rather than separate PR:** the bug silently routes broken data through the new `lots` table that this PR introduces — natural to fix where reviewers are already looking at lot-correctness. Tiny (~68 lines).

**Verified:** `test_import_force_import_bypasses_*` (the canary tests for the idempotency_key preservation) still pass. All workspace tests green. `cargo clippy --tests -- -D warnings` clean.

afadil:
All four findings addressed in [cfdd87eb6](https://github.com/wealthfolio/wealthfolio/commit/cfdd87eb6).

**F1 (P1) — Closed synthetic lots deleted on incremental recalcs.** Confirmed via the migration schema. The orphan sweep's comment claimed it handled `FK ON DELETE SET NULL` on `open_activity_id`, but the schema is `ON DELETE CASCADE` — a deleted opening activity removes the whole lot row, never leaving a NULL FK. The sweep was guarding against a case that physically can't happen, and was actively destroying real DRIP/staking closed-lot history (which legitimately has NULL `open_activity_id` after F1's synthetic-ID normalization) on every incremental recalc whose `known_ids` didn't include the originating activity. Removed the closed-NULL sweep entirely. The open-lots sweep handles legitimate orphans.

**F2 (P2) — Sync wiped HOLDINGS lots for historical-snapshot events.** Confirmed. Added `is_latest_non_calculated` / `was_latest_non_calculated` helpers that compute `max(snapshot_date)` for the account's non-CALCULATED snapshots. Update wipes only when the synced row equals the new latest. Delete wipes only when the pre-deletion row was the latest. Two new regression tests `replay_snapshot_{update,delete}_preserves_lots_for_historical_snapshot` lock the contract.

**F3 (P2) — Import preparation errors.** Confirmed and partly fixed, partly deferred. Tried two approaches:
1. Drop all `prep.errors` entries — broke `test_import_accepts_cash_dividend_without_symbol` (cash flows like DIVIDEND-without-symbol legitimately surface in prep.errors but should still import with asset_id=NULL).
2. Gate drop on `requires_symbol(act_type)` — same break (DIVIDEND is in that list since broker-exported DIVIDENDs do need symbol resolution when one's provided).

Distinguishing "legitimate cash-only error" from "real symbol-resolution failure" requires reproducing `classify_import_activity`'s logic at the drop point. Deferred with a comment block documenting the trade-off at the decision site. The main bug (no resolution at all on `/activities/import`) remains fixed.

**F4 (P2) — SinceDate cleanup lot errors swallowed.** Confirmed. Now pushes the error into the same `lot_sync_errors` Vec the main loop uses, so the function surfaces a real Err at the end instead of returning Ok with the lots table out of sync.

Verified: 1063 core / 130 storage (+2 new sync tests) / 248 connect / 50 device-sync / others all pass. `cargo clippy -p wealthfolio-core -p wealthfolio-storage-sqlite -p wealthfolio-server --tests -- -D warnings` clean.

Frontend Check failure on `custom-provider-form.test.tsx:57` is unrelated to this PR — that file isn't in my diff and is an existing flake on main.

afadil:
Updated direction after the latest review:

- `lots` stays a transaction-derived tax-lot table only. HOLDINGS-mode aggregate snapshots are no longer written as synthetic `lots` rows.
- The Asset Lots tab now uses a backend read model: transaction lots from `lots`, plus latest HOLDINGS-mode snapshot positions as aggregate rows with `source = SNAPSHOT_POSITION` when requested.
- First-launch lot backfill and the synthetic HOLDINGS lot refresh/sync cleanup paths were removed. The normal portfolio calculation refills transaction-derived lots after migration.
- The migration now creates `lots` and `snapshot_positions`, preserves user-entered/imported HOLDINGS snapshots, clears derived `CALCULATED` snapshots, and clears `daily_account_valuation` so valuations rebuild from the regenerated snapshots.
- Existing HOLDINGS snapshots are not SQL-backfilled into `snapshot_positions`; read paths use relational positions when present and fall back to legacy `holdings_snapshots.positions` JSON for older snapshots.
- Added regression coverage for stale `snapshot_positions` cleanup when a snapshot is updated to an empty position map.

Verification run locally:
- `cargo test -p wealthfolio-storage-sqlite test_save_or_update_snapshot_clears_positions_when_empty`
- `cargo test -p wealthfolio-storage-sqlite test_save_snapshots_clears_positions_when_empty`
- `cargo test -p wealthfolio-storage-sqlite replay_snapshot_clears_stale_snapshot_positions`
- `cargo test -p wealthfolio-storage-sqlite lots`
- `cargo test -p wealthfolio-storage-sqlite portfolio::snapshot::repository::tests`
- `cargo check -p wealthfolio-server`
- `cargo check -p wealthfolio-app`
- `git diff --check`

Reviewer note: if you ran an earlier version of this PR locally, your dev DB may already contain synthetic `HOLDINGS-*` rows in `lots`. Reset the migration/dev DB or clear those rows before testing the new read-model behavior to avoid seeing duplicate Transaction + Snapshot rows.

afadil:
Updated direction after review cleanup:

- `lots` now stays transaction-only. HOLDINGS-mode accounts are snapshot tracking, so their asset-lot rows are exposed as aggregate `SnapshotPosition` rows in `AssetLotView` instead of synthetic rows in the `lots` table.
- Removed the first-launch/HOLDINGS backfill direction, unused staged Phase B replay helper, unused split-ratio backfill helper, and stray lot debug logging.
- Kept the additive migration shape: create `lots` + `snapshot_positions`, preserve legacy JSON, clear derived calculated snapshots/valuations so normal recalculation repopulates transaction lots.
- Tightened nits from review: parameterized the sync delete, defaulted the include-snapshot query bool, aligned Diesel schema/model ordering with the migration, and added focused regression coverage for lot-sync error propagation plus split calendar-date eligibility.

Verification is listed in the updated PR body.

PR #1008, opened 2026-05-20
fix(allocation): normalize weights >100% and guard parent/child double-counting in regions

Jonjon-prog:
## Summary

Fixes two bugs in `aggregate_by_taxonomy` that caused region/country allocation percentages to exceed the portfolio total. Pre-existing debt flagged during PR #995 review.

**Bug 1 — Weights above 100% not normalized**
If an asset had taxonomy assignments summing above 10000 bps (100%), each weight was applied directly, so the asset could contribute more than its full market value to the totals. Fix: divide by `max(total_weight, 10000)` so contributions are always proportional and never exceed market value.

**Bug 2 — Parent/child region assignments double-counted on rollup**
In rollup mode (`rollup_to_top_level = true`), if an asset was assigned to both a parent category (Americas) and a child (United States), rolling up United States to Americas while also counting the direct Americas assignment would sum both. Fix: leaf-wins — skip a top-level assignment when a child of that top-level is already assigned for the same asset.

## Tests

Two regression tests added directly to `allocation_service.rs`:
- `weights_above_100_pct_are_normalized` — 60%+60% AAPL scenario, total % must not exceed 100%
- `parent_child_region_not_double_counted_on_rollup` — Americas+United_States scenario, Americas value must equal leaf-only contribution (400, not 1000)

## Related

Requested as follow-up in [#995 review](https://github.com/wealthfolio/wealthfolio/pull/995#issuecomment-0) by @afadil:
> normalize or validate assignment weights per asset/taxonomy; guard against overlapping parent/child region assignments being double-counted; add a regression test where one asset has region weights above 100% and verify portfolio percentages do not exceed the scoped total.

## Checklist

### DIFF
--- a/crates/core/src/portfolio/allocation/allocation_service.rs
+++ b/crates/core/src/portfolio/allocation/allocation_service.rs
@@ -156,12 +156,6 @@ impl AllocationService {
                         .entry("__UNKNOWN__".to_string())
                         .or_insert(Decimal::ZERO) += market_value;
                 } else {
-                    // Normalize weights: if total exceeds 10000 bps (100%), scale down
-                    // proportionally so a single asset never contributes more than its
-                    // full market value to the portfolio total.
-                    let total_weight: i32 = taxonomy_assignments.iter().map(|(_, _, w)| w).sum();
-                    let weight_divisor = Decimal::from(total_weight.max(10000));
-
                     // In rollup mode: skip top-level category assignments when a child of
                     // that category is also assigned for this asset (leaf-wins principle).
                     // Without this guard, Americas + United States both roll up to Americas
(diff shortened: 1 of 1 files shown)


afadil:
Could you please check these:

**Findings**

1. **P1: Drill-down still double-counts and skips normalization**  
   [allocation_service.rs](/private/tmp/wealthfolio-pr-1008-review/crates/core/src/portfolio/allocation/allocation_service.rs:556) still sums raw assignment weights for matching parent + child categories, then [uses `/ 10000`](/private/tmp/wealthfolio-pr-1008-review/crates/core/src/portfolio/allocation/allocation_service.rs:610). In the PR’s own Americas + United_States example, summary allocation shows `Americas = 400`, but clicking Americas in the detail sheet can return `1000` because the drill-down path adds `6000 + 4000`. The same mismatch affects the AI drill-down tool. This should share the same normalized/leaf-wins logic as `aggregate_by_taxonomy`.

2. **P2: skipped parent weights still dilute kept leaf weights**  
   [allocation_service.rs](/private/tmp/wealthfolio-pr-1008-review/crates/core/src/portfolio/allocation/allocation_service.rs:163) computes `total_weight` before [skipping parent assignments](/private/tmp/wealthfolio-pr-1008-review/crates/core/src/portfolio/allocation/allocation_service.rs:197). If an asset has `Americas=100%` and `United_States=100%`, the parent is skipped but the child is divided by `20000`, so the kept leaf contributes only 50%. If “leaf-wins” means parent assignments are ignored, build the active assignment set first, then normalize over only those active assignments.

Jonjon-prog:
Both fixed in the latest push.

**P2 — weight dilution**
 was computed over all assignments (including parents about to be skipped), so a skipped  was still inflating the divisor for the kept  leaf, yielding 50% instead of 100%. 
 Fix: build  after the leaf-wins filter first, then compute over only those active assignments before normalizing.

**P1 — drill-down double-counting**
`compute_holdings_by_allocation_from_holdings` was accumulating weights for all matching categories (parent + children) into a single `asset_to_weight` map via `+=`, so Americas=6000 + United_States=4000 summed to 10000. 
Fix: collect direct (target category) and child weights separately, then apply leaf-wins
Child weights take precedence, direct weight used only when no child exists for that asset. 
Also caps each asset's effective weight to 10000 bps.

PR #1009, opened 2026-05-21
Add private assets manual-entry ledger

SCVtx:
## Summary

Adds a dedicated manual-entry ledger for private assets behind a default-off feature flag. When enabled, users can track fund managers, private assets, sub-assets, and valuation/cash-flow statements; those values then appear in holdings, dashboard, net worth, and assistant/tool surfaces. When disabled, private-asset routes/actions are hidden and backend APIs/commands reject access.

This keeps private assets separate from public securities because they have different source data, valuation cadence, and audit needs. Public holdings are transaction/market-price driven; private assets are statement-driven and often manually curated. A dedicated ledger keeps that behavior explicit without forcing private statements into the public-securities activity model.

## Data Model / Migrations

- Adds SQLite migrations for private-assets foundation tables, direct/manager check constraints, and statement cash-flow type.
- Adds storage repositories, core models/services, projection logic, web API handlers, Tauri commands, and AI tools for private assets.
- Extends settings with `privateAssetsEnabled`, defaulting to `false`.
- Net-worth/history services include private assets only when the setting is enabled.
- Preserves upstream portfolio-scope tables, AI environment split, and recent dashboard/net-worth changes while adding the private-assets schema.

## UI / API Surfaces

- Adds settings pages for fund managers, private assets, sub-assets, and valuation statements.
- Adds holdings/dashboard/net-worth integration when the feature is enabled.
- Adds feature toggle under General settings.
- Hides private-assets launcher/search/navigation/assistant data-access affordances while disabled.
- Documents the architecture in `docs/architecture/private-assets.md` and links it from `README.md`.

## Non-Scope

- No PDF parsing/import.
- No IRR/performance analytics for private funds.
- No event-ledger generalization.
- No addon `.npmrc`, local handoff docs, temporary files, or local tooling residue.

## Validation

Passed:

- `cargo test -p wealthfolio-core private_assets --lib`
- `cargo test -p wealthfolio-storage-sqlite private_assets --lib`
- `cargo test -p wealthfolio-ai tool_names_are_exactly_the_strings_used_by_allowlist --lib`
- `cargo check -p wealthfolio-server`
- `cargo check -p wealthfolio-app --lib`
- Targeted frontend Vitest for touched private-assets/holdings/dashboard files
- `git diff --check`
- `bash scripts/ci-check.sh`
- `bash scripts/ci-check.sh --full` after rebasing onto latest `upstream/main`
- `pnpm test:e2e` with a local ignored `.env.web`: 88 passed

Notes:

- `bash scripts/ci-check.sh --full` completed with the repo's existing warning-only lint output and normal Vite bundle/source-map warnings.
- `pnpm test:e2e` needed local setup from `.env.web.example`; the ignored `.env.web` was adjusted to bind the backend on `8088` and use a local dummy 32-byte test secret. This file is not tracked and is not included in the PR.
- Native desktop GUI smoke was not rerun in this final rebase pass. Command registration/startup coverage was checked through Rust compile/check paths.

## Mobile QA

Previously checked responsive web at `390x844` and `430x932` with a throwaway database:

- Default-off state: `privateAssetsEnabled=false`, private-assets API returns 403, `/settings/private-assets` redirects to General settings.
- Enabled state: `/settings/private-assets`, private-asset detail route, statement modal, `/holdings?tab=private-assets`, dashboard, and net worth.
- Verified modal scroll/reachability for Save statement and Cancel.
- Verified visible settings/detail surfaces had no text/control overflow.
- Verified holdings/dashboard/net-worth private-assets totals/lists were visible when enabled.
- No console errors observed.
- Fixed a route-gate loading issue found during mobile QA where direct private-assets links redirected before settings had loaded.

## Scrub Notes

- Re-fetched latest `upstream/main`; this branch is one commit on top of `3af0637f7`.
- Scrubbed the PR diff for `.env`, `.npmrc`, `tmp/`, `.playwright-mcp`, handoff docs, stale `history-chart.tsx`, local paths, and common secret patterns.
- No secrets or local-only files are included.

## CLA

I am comfortable submitting this contribution under the upstream project CLA and will complete any GitHub CLA prompt/check if requested.


afadil:
this is a huge PR, could you post some screenshots?
also i feel this is something that should fit or extend the existing personal assets .

SCVtx:
Thanks. I generated a screenshot walkthrough from the PR head with seeded demo data: five private assets, public market holdings, and existing personal/alternative assets so the new surfaces can be compared side by side.

Screenshot source: `SCVtx/wealthfolio:codex/private-assets-upstream-pr-v2` at `faa408bf3c28e890381daf98d4436bee37a16b92`. I pushed the images to a separate screenshot-only branch so the code PR diff stays clean: https://github.com/SCVtx/wealthfolio/tree/codex/pr1009-screenshots-20260611/pr1009/screenshots

<details>
<summary>Screenshots</summary>

### 1. General settings feature toggle
General settings feature toggle

### 2. Private assets list with five seeded assets
Private assets list

### 3. Private equity fund detail
Private equity fund detail

### 4. Direct/co-invest detail
Direct private asset detail

### 5. Dashboard investments tab
Dashboard investments tab

### 6. Dashboard net worth tab
Dashboard net worth tab

### 7. Holdings public investments tab
Holdings public investments tab

### 8. Existing personal/alternative assets tab
Existing personal alternative assets tab

### 9. Holdings private-assets tab
Holdings private assets tab

### 10. Holdings insights private-assets allocation
Holdings insights private assets allocation

### 11. App launcher private-assets search
App launcher private assets search

### 12. Feature disabled route behavior
Disabled route redirect

</details>

On fitting this into the existing personal/alternative assets model: I agree the feature should live near that vocabulary in the product, but I do not think the private-investment records should be implemented by extending the same lightweight record used for cars, watches, collectibles, property, and similar assets.

The current personal/alternative asset shape is good for value-only assets: the user has an object, maybe a cost basis, maybe a latest valuation, and Wealthfolio can answer “what is this worth today?” That model stays intentionally simple.

Private funds and direct private investments need different accounting facts:

- fund manager or sponsor
- vehicle kind, such as fund vs direct investment
- strategy, such as private equity, venture, private credit, real estate, infrastructure, or fund of funds
- committed capital
- contributed / called capital
- distributed capital
- current reported value
- statement date and value source
- stale-statement handling
- optional look-through rows when the user wants to track underlying portfolio companies, fund sleeves, or co-invest positions

Those are not just extra display fields. They affect how the asset is created, validated, updated, summarized, hidden, and eventually analyzed.

If we extend the existing personal asset structure to hold these fields, we end up with a few practical problems:

1. The personal asset table becomes sparse and confusing. Most rows, like a car or a watch, would have null manager, null commitment, null contributed amount, null distributions, null statement basis, and null look-through state. The generic table starts carrying semantics that only apply to one narrow class of asset.

2. The validation rules diverge. A direct private investment should not require a fund manager. A fund vehicle usually should. A watch should not know what a commitment amount is. A statement mark should have an as-of date and value source. Those rules become hard to express cleanly if every personal asset row shares one generic shape.

3. History becomes ambiguous. Existing personal assets mostly need valuation history. Private funds need statement history: value plus contributed capital, distributed capital, cash-flow basis, date, and source. Using only the existing value history loses the capital-account story that users need for private investments.

4. Reporting becomes noisier. Holdings, dashboard, net worth, AI context, and exports need to know whether they are looking at a market security, a simple personal asset, or a private investment with capital-account detail. If private investments are hidden inside generic personal assets, every reporting surface needs special-case logic anyway.

5. Look-through does not fit the simple asset record. If the user tracks a fund and optional underlying positions, that is a parent/child relationship with different reporting basis and ownership metadata. Folding that into a generic asset record either drops the detail or forces nested private-investment concepts into a model meant for standalone objects.

6. Future work becomes harder. PDF or statement ingestion, IRR/performance analytics, stale-mark warnings, capital-call workflows, and manager-level rollups all depend on private-investment-specific facts. Keeping a dedicated ledger gives those features a clear place to grow without making every personal asset carry that complexity.

The intent here is not to create a completely unrelated asset universe. The UI still places the feature in the same broad area as holdings and personal/alternative assets, and the aggregate values flow into dashboard, holdings, net worth, and assistant/tool surfaces. The separation is mainly at the persistence and service layer so the accounting model stays explicit.

This is also why the feature is default-off. If a user needs private-asset tracking, they generally need the whole set of related detail: managers, commitments, statement marks, distributions, and sometimes look-through. If they do not need that, the feature should stay invisible and the existing personal/alternative asset experience should remain simple.

Process note: I have not refreshed this PR branch onto the newest `main` again yet. Since the remaining question is the model shape, I did not want to churn the diff before confirming direction. Once you are comfortable with this direction, I can update the PR branch against current `main`, rerun checks, refresh screenshots if needed, and keep this PR moving in place.

PR #1022, opened 2026-05-27
fix(valuation): retain full quote-gap days

Jonjon-prog:
## Problem

Valuation calculation could omit a date when none of the account's current quotable positions had a usable quote on or before that date. If the account had valuation rows both before and after the omission while another account had that date, scoped aggregation rejected the history as an interior gap and the dashboard/performance request failed.

## Fix

- Emit a typed valuation row for a full quote-gap date instead of dropping the date.
- Calculate the row from that date's holdings snapshot, preserving current cash, net contribution, external flows, and FX.
- Let the valuation calculator mark missing coverage as `PartialUnpriced` or `Unavailable`.
- Keep scoped history validation unchanged so genuinely missing persisted rows are still rejected.

Current `main` already carries the latest available quote forward for ordinary market holidays and provider gaps. This rebased implementation handles the remaining no-prior-quote transition safely without cloning a stale prior-day valuation.

## Validation

- `cargo test -p wealthfolio-core --lib` — 1,722 passed
- `cargo check -p wealthfolio-core`

## Checklist

### DIFF
--- a/crates/core/src/portfolio/valuation/valuation_service.rs
+++ b/crates/core/src/portfolio/valuation/valuation_service.rs
@@ -2347,7 +2347,10 @@ impl ValuationService {
                 if !quotable_positions.is_empty()
                     && missing_quotes.len() == quotable_positions.len()
                 {
-                    return None;
+                    debug!(
+                        "No quotes available on or before {} for any quotable position in account '{}'; emitting an unavailable valuation row.",
+                        day.date, account.account_id
+                    );
                 }
 
                 let fx_today = facts.fx_rates_for_date(&account.required_fx_pairs, day.date);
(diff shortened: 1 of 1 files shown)


Jonjon-prog:
@afadil 

I rebased this PR on latest main. 

Single conflict in the `aggregate_scoped_valuations` signature (new `internal_transfer_flow_adjustments_by_date` param + `DailyFlowAmounts` type), resolved cleanly.

All 1442 tests in `crates/core` pass. The fix itself is unchanged: `fill_interior_gaps()` forward-fills missing interior dates before the validator runs, so multi-account aggregation no longer errors when an account skips a day due to absent quotes.

PR #1011, opened 2026-05-23
Implement scoped lots valuation and portfolio aggregation

afadil:
## Summary

Implements `docs/lots-valuation.md` phases 1-6 as a production path for real-account scoped valuation, performance, and saved portfolio aggregation.

### Design direction

- Removes the runtime dependency on persisted `account_id = "TOTAL"` valuation/snapshot rows.
- Resolves `AccountScope` into concrete real account IDs for all/account/accounts/saved portfolio requests.
- Aggregates scoped holdings, allocation, income, valuation history, latest valuation, performance, goals, AI tools, addon SDK APIs, and frontend selectors from real account data.
- Recreates `daily_account_valuation` as unsynced derived data with account/base-currency columns: cash, investment market value, total value, cost basis, net contribution, external inflow/outflow, and performance-eligible value.
- Keeps alternative assets net-worth-only by excluding them from investment valuation and performance-eligible values.
- Uses Modified Dietz fields for money-weighted scoped performance and preserves TWR/simple-return behavior for holdings-mode and mixed tracking-mode scopes.
- Aggregates saved portfolio performance from scoped real accounts instead of pseudo-account snapshots.

### Refactoring and improvements

- Adds scoped valuation repository/service APIs and base-currency aggregation paths.
- Reworks server and Tauri portfolio update jobs to recalculate real account snapshots/valuations and reconcile quote sync from latest account snapshots.
- Updates web/Tauri adapters, account selectors, dashboard, holdings, allocation, income, performance, account, asset, AI tool UI, and retirement-goal surfaces to carry account scopes consistently.
- Adds adapter parity coverage for the new scoped APIs.
- Adds import review rule coverage without adding `prepare_activities_for_import` to the import apply path; the frontend still prepares and creates assets before import by design.
- Updates snapshot history UI to display total cash value instead of the misleading count of cash-currency buckets.

### Fixes included

- Migration cleanup removes/ignores orphan legacy aggregate snapshot/valuation rows while avoiding new `TOTAL` dependencies.
- Migration now normalizes legacy `HOLDINGS` + `CALCULATED` snapshots to `MANUAL_ENTRY` before deleting rebuildable calculated snapshots, preserving cash-only holdings-mode history like TD Chequing.
- Sync restore ignores orphan snapshot rows and avoids restoring legacy aggregate rows.
- Cash base valuation now uses the calculated cash balance instead of stale cached snapshot totals.
- Dashboard top holdings uses total gain for total-return display and no longer shares the holdings-page display preference key.
[...]

PR #902, opened 2026-04-28
feat(allocations): portfolio allocation targets & rebalancing advisor

Jonjon-prog:
## Summary

Hi 👋

This PR adds a full **Allocation Targets** system with a **Rebalancing Advisor** to Wealthfolio — letting users define target allocations per asset class and get actionable trade suggestions to reach those targets.

- Define target % per asset class (Equity, Bonds, Cash…) with optional per-holding breakdowns within each class
- Visual health strip showing portfolio alignment at a glance (on target / drift / amount to deploy)
- Rebalancing advisor generates a prioritized trade list given available cash, in **Buy Only** or **Buy & Sell** mode
- CSV export in importable format (compatible with the activity import flow, includes ISIN for European users)
- Allocation Strategy settings page to configure rebalancing mode per account

## How it works

**Overview tab**
- Health strip: portfolio value, classes on target, worst drift, amount to deploy to reach targets
- Donut chart + deviation table (actual % vs target %, drift colored green/yellow/red)
- Click on a class → drilldown view opens inline: flat list of holdings on the left, target editor + holding-level targets on the right

screenshot 2026-04-28 à 22 30 37

</br>

screenshot 2026-04-28 à 22 30 46

</br>

**Rebalancing tab**
- Enter available cash → one click generates a trade plan
- **Buy Only**: deploys cash into underweight positions proportionally to their shortfall
- **Buy & Sell**: sells overweight positions first, redeploys proceeds + new cash
- Trade table with sortable columns (class, asset, action, shares, price, amount)
- BUDGET rows when a class has budget but no holding-level targets configured (with warning tooltip)
- Warning icons when holding targets are partial or missing
- CSV export ready to import as activities (symbol, activityType, quantity, unitPrice, isin…)

screenshot 2026-04-28 à 22 36 18

</br>

screenshot 2026-04-28 à 22 59 39

</br>

## Technical changes

**Rust / Backend**
- New `portfolio/targets` module: `PortfolioTarget`, `TargetAllocation`, `HoldingTarget` with full CRUD and SQLite repository
- New `portfolio/rebalancing` module: buy-only and buy-and-sell algorithms with 13+ unit tests
- `has_partial_targets` detection: compares portfolio holdings vs configured holding targets
- `isin` field propagated through `Instrument` → `HoldingSummary` → `TradeRecommendation` from `asset.metadata`
- 3 new migrations: `portfolio_targets`, `holding_targets`, `rebalance_mode`

**Frontend**
[...]

PR #1025, opened 2026-05-28
feat(allocation-targets): Allocation Targets — Milestone 1 + full UI

Jonjon-prog:
Hello @afadil 

This is the first big  PR for the allocation and target feature. 
I will let you the time to review it, I tried to document it as much as possible. 
I started it before the new spending tracker merge, so I will do a rebase. 

For now I put the allocation page in the menu tab for simplicity of use but after if you want to move it, I can dot it. 
Just as it will be a bigger feature as before, maybe it's better to have it living alone for UI simplicity. 

## Summary

Introduces the **Allocation Targets** module: define target weights per taxonomy, monitor live drift, and decide when to act. This PR covers backend + full UI through the "Targets & Drift" phase. 
The Rebalance Planner is the next PR.

This is the first step of the rollout defined in `docs/features/allocations/v1-spec.md` and the SOTA phase plan in `docs/features/allocations/sota-target-model-spec.md`.

---

### V1 Spec — what's covered

| V1 Requirement | Status |
|---|---|
| One active profile per scope (unique constraint) | ✅ Done |
| Scope: whole portfolio, single account | ✅ Done (`all`, `portfolio`, `account`) |
| Primary taxonomy target, default `asset_classes` | ✅ Done |
| Target categories with zero current holdings | ✅ Done |
| Targets stored in basis points | ✅ Done |
| Drift calculation (DriftService) | ✅ Done |
| Tolerance band trigger | ✅ Done |
| Save draft only on explicit user action | ✅ Done |
| Desktop and web adapter parity | ✅ Done |
| Manual rebalance planning (RebalanceService) | ❌ Next PR |
| Cash-flow-only planning | ❌ Next PR |
| Optional sell-to-rebalance mode | ❌ Next PR |
| Export suggested trades as CSV | ❌ Next PR |
| Minimum trade amount / whole-share rounding | ⚠️ Fields exist in model & DB, not used yet (no RebalanceService) |

**V1 Milestone 1 (Targets and Drift): complete.**
V1 Milestones 2–4 (Cash-Flow Planner, Sell Mode, Integration Hooks): next PR.

### SOTA Spec — Phase 1 coverage

| SOTA Entity | V1 decision |
|---|---|
| `TargetProfile` | ✅ Simplified — no `objective`, `model_type`, `version`, `effective_from` |
| `TargetAllocationNode` | ✅ Simplified — no `parent_node_id`, per-node `min_bps`/`max_bps`, `rebalance_priority` |
| `RebalancePolicy` | ✅ Inline fields on `target_profiles` (not a separate table) |
| `TargetGuardrail` | ❌ Not in V1 |
| `HoldingTarget` | ❌ Not in V1 |
| `FundingPolicy` | ❌ Not in V1 |
| `ExecutionPolicy` | ⚠️ Inline simple fields (`allow_sells`, `min_trade_amount`, etc.) |
| `RebalanceRun` / `TradeDraft` | ❌ `rebalance_drafts` table scaffolded, not written to yet |

---

### Profile management
- **CRUD** — create, edit, activate, archive, delete target profiles.
- **Scope** — `all` (whole portfolio), `portfolio` (saved portfolio), `account`.
  Unique-active constraint enforced at DB + service level: activating a profile
  auto-archives the previous active profile for the same scope.
- **Trigger types** — `threshold` (drift exceeds band), `calendar` (fixed
  schedule — monthly / quarterly / semi-annual / annual), `combined`
  (whichever fires first), `manual`. V1 spec defined manual + threshold;
  calendar/combined added now because SOTA spec calls for them and SQLite CHECK
  constraint recreation makes later migration expensive.
- **Drift tolerance** — configurable band 0.5–10 % (stored as bps).

### Target weights editor
- Inline bps editor with proportional auto-adjust across unlocked rows.
- Per-row lock to pin a weight and redistribute among the rest.
- Total-weight indicator gating Save/Activate buttons.

### Taxonomy support (beyond V1 default)
V1 spec defaults to `asset_classes`. We support 5 taxonomies with model presets (3 built-in + "current allocation" each):

| Taxonomy | Presets |
|---|---|
| Asset classes | Three-Fund, 60/40, All Weather |
| Industries (GICS) | S&P 500, Equal Weight, Defensive |
| Regions | Global Cap, Ex-US, Equal Weight |
| Instrument type | Passive, Core-Satellite, Diversified |
| Risk category | Conservative, Balanced, Aggressive |

Current allocation mapped per-taxonomy from existing `PortfolioAllocations` service (no new backend needed). Switching taxonomy resets weights; switching back to the saved taxonomy restores persisted weights (no false dirty-flag).

### Drift engine
- `DriftReport` per profile: current/target bps, drift bps, value delta,
  `DriftStatus` (in_band / underweight / overweight / not_targeted) per row.
- **Drift bar** — visual current vs. target breakdown.
- **Drift drivers** (`DriftDriversCard`) — see open question below.

### UI highlights
- Onboarding path when no profile exists.
- Unsaved-changes guard (navigate away → Save / Discard / Cancel).
- Taxonomy picker uses `DropdownMenu`, not native `<select>`.
- Scope-aware current allocation feeds the "current allocation" preset for each active taxonomy.

---

## Migrations

| Migration | Effect |
|---|---|
| `2026-05-25-000002_allocation_targets` | Creates `target_profiles`, `target_allocation_nodes`, `rebalance_drafts` |
| `2026-05-28-000001_target_profile_calendar_trigger` | Recreates `target_profiles` with updated `trigger_type` CHECK + `review_frequency` / `next_review_date` columns |

Both have full `down.sql` reversions.

---

## What remains before V1-complete (next PR)

1. **RebalanceService** — cash-flow-only planning (V1 Milestone 2)
2. **Available cash input** + suggested manual buys
3. **Minimum trade amount** + whole-share rounding (fields already persisted)
4. **CSV export** of suggested trades
5. **Sell-to-rebalance** scenario (V1 Milestone 3, gated on `allow_sells`)
6. **Save rebalance draft** (table exists, not written to yet)
7. **Rebalance tab** (currently placeholder "Coming soon")

After that, remaining SOTA phases:
- Phase 2: sell-to-rebalance + hybrid scenario + rebalance runs + trade drafts
- Phase 3: funding policies (recurring contributions, value averaging), guardrails
- Phase 4: tax-aware lot-level planning
- Phase 5: model marketplace + addon SDK

---

## Screenshots

screenshot 2026-05-28 à 14 17 34
</br>

screenshot 2026-05-28 à 14 17 40
</br>

screenshot 2026-05-28 à 14 18 02
</br>

screenshot 2026-05-28 à 14 18 14
</br>

screenshot 2026-05-28 à 14 18 31
</br>

screenshot 2026-05-28 à 14 19 49
</br>

screenshot 2026-05-28 à 14 33 41
</br>

screenshot 2026-05-28 à 14 33 56
</br>

---
### 1. "What moved off target" — DriftDriversCard

`DriftDriversCard` explains why sleeves drifted using Option A (rule-based, already implemented). 
Two alternatives:

**Option A** (current) — infer attribution from current snapshot vs. target: market drift, cash inflow, new positions. Fast, no storage required.

**Option B** — temporal comparison: compare current snapshot to the prior
snapshot (last rebalance date or last month-end). Requires persisting an
`allocation_snapshot` per profile on a schedule. More accurate but adds
a scheduled job and storage.

**Option C** — AI-powered: send drift report + recent activity to an LLM, return a natural-language explanation. Readable but requires an AI assistant configured and incurs LLM cost per view.

_Recommend aligning on B or C before implementing storage or the LLM path._

### 2. Multi-taxonomy targeting: in V1 or not?

V1 spec specifies `asset_classes` as the primary taxonomy. I ship 5 taxonomies.
**Question:** should multi-taxonomy profiles (e.g. targeting by Regions) be in V1, or restrict the UI to `asset_classes` until a dedicated "secondary guardrail" flow exists from the SOTA spec?

### 3. Execution constraints visibility

`allow_sells`, `min_trade_amount`, `whole_shares_only`, `rebalance_to` are persisted but hidden from the profile editor UI (defaults used). 
They will matter once the RebalanceService ships.

**Question:** expose these in the current editor now, or show them only in the planner context?

### 4. Scope `all` vs. `portfolio` terminology

V1 spec uses `portfolio` scope with null `scope_id` for the whole portfolio.
I implemented `all` (also null `scope_id`) to align with `AccountScope.all` from #1011. Functionally equivalent.

**Question:** rename `all` → `portfolio` in the model to match V1 spec, or keep `all` consistent with the broader scope model?

### 5. `next_review_date` management for calendar/combined triggers

`next_review_date` is persisted (nullable) but not computed yet. 
For calendar/combined triggers to work end-to-end, something must advance the date after each review.

**Question:** user-set (date picker in editor), auto-computed on activate (today + frequency), or computed by a background job? Or defer until TriggerService ships in Phase 2?

### 6. Open questions from V1 spec §15

- Sell-to-rebalance in the first planner PR, or cash-flow-only only?
- Default profile scope for new users: whole portfolio or first account?
- CSV export format for suggested trades?
- Should saved drafts surface in UI in v1, or export-only?

---

## Verification

- `cargo check -p wealthfolio-core -p wealthfolio-storage-sqlite` — clean
- `cargo test -p wealthfolio-core --lib` — 3 passed
- `cargo test -p wealthfolio-storage-sqlite --lib` — 135 passed
- `pnpm type-check` — clean on allocation-targets files
- `pnpm test` — 60 files, 657 tests passed
- Manual smoke (all 5 taxonomies, all 4 trigger types):
  create profile → apply preset → edit weights → save draft → activate →
  view drift report → archive → delete; switch taxonomy back → no dirty-flag;
  calendar trigger → frequency picker → blocks Save without frequency
  
  
  ## Test Plan

**Rust Unit Tests — 25 tests, all green** (`cargo test -p richessefolio-core --lib allocation_targets`)

`validation.rs` — 16 tests :
- [x] Empty profile name rejected
- [x] Valid profile accepted
- [x] Scope `Account` without `scope_id` rejected
- [x] Scope `Account` with `scope_id` accepted
- [x] `drift_band_bps > 10000` rejected
- [x] `drift_band_bps = 0` accepted
- [x] Trigger `calendar` without `review_Frequency` rejected
- [x] Trigger `calendar` with `review_Frequency` accepted
- [x] Trigger `combined` without `review_Frequency` rejected
- [x] Negative `min_trade_amount` rejected
- [x] `min_trade_amount` non-decimal rejected
- [x] Nodes summing at 10000 bps accepted
- [x] Nodes not summing at 10000 bps rejected
- [x] Duplicate `category_id` rejected
- [x] Negative `target_bps` rejected
- [x] Target set to zero allowed if sum = 10000

`drift_service.rs` — 9 tests:
- [x] Overweight detected (current > target + band)
- [x] Underweight detected (current < target − band)
- [x] InBand detected (|drift| ≤ band)
- [x] `is_zero_current = true` + Underweight when no position
- [x] Category not targeted → `NotTargeted` line added
- [x] `max_drift_bps` calculated on required lines
- [x] `out_of_band_count` correct
- [x] `value_delta` correct (current_value − target_value)
- [x] Empty portfolio (total_value = 0) → all `current_bps = 0`

**Frontend** : `pnpm test` → 691 tests, all green. 
  
  ## Checklist


afadil:
@Jonjon-prog , Thanks for all the work!

I think we should tighten the milestone to what actually works end to end: **Targets + Drift + manual/threshold review**.

A few things I would change before merge:

1. **Remove Calendar / Combined triggers from this milestone**

I do not think we need Calendar for V1. Manual + Threshold is enough:

- Manual: the user can review allocation whenever they want.
- Threshold: the app shows when the portfolio is outside tolerance bands.

Calendar creates a separate review-cadence workflow, but that workflow is not implemented yet. It would need next-review calculation, mark-reviewed behavior, date advancement, due/overdue UI, and possibly notifications later.

Since this feature has not shipped yet, I would avoid adding `calendar`, `combined`, `review_frequency`, and `next_review_date` to the initial schema. We can add review cadence later without changing the core target/drift model much.

2. **Merge the DB migrations into one clean migration**

The feature has not shipped yet, so we should not carry a second migration that recreates the first table just to change the initial model. This was also requested in the earlier #902 discussion.

Please merge the two migrations into one clean initial allocation-targets migration.

3. **Fix the active-profile uniqueness constraint**

The table says “one active profile per scope” is done, but the SQLite unique index does not enforce this for `scope_id IS NULL`. SQLite allows multiple NULL values in a unique index, so multiple active `all` profiles can exist.

Please use either a COALESCE expression index or separate partial indexes for null and non-null scopes.

4. **Add DB-level scope invariants**

The service validates some scope rules, but the database still allows invalid combinations.

Please add a check constraint so:

- `scope_type = 'all'` requires `scope_id IS NULL`
- `scope_type IN ('account', 'portfolio')` requires `scope_id IS NOT NULL`

5. **Fix multi-taxonomy drift or restrict V1 to asset classes**

The PR says five taxonomies are supported, but the drift service does not appear to match the taxonomy ids used by the UI.

The UI uses ids like `industries_gics` and `instrument_type`, while the drift service checks `sectors` and `security_types`. That means non-asset-class profiles can calculate drift against the wrong allocation set.

Please either fix and test every supported taxonomy, or restrict Milestone 1 to `asset_classes`.

6. **Do not trust `profile_id` from node payloads**

`save_nodes` receives the profile id as an argument, but then uses `profile_id` from each request node when constructing rows. A malformed request could delete nodes for one profile and insert nodes for another.

Please derive each node’s `profile_id` from the route/service argument, or reject mismatched payloads.

7. **Make zero-target behavior consistent**

The backend/tests allow `target_bps = 0`, but the UI filters out zero-target rows before saving. That makes it impossible to explicitly say “this category should be 0%”.

Please either preserve explicit 0% targets or remove zero-target support consistently.

8. **Validate node categories against the profile taxonomy**

`target_allocation_nodes.category_id` should belong to the selected `taxonomy_id`. Otherwise malformed data can mix categories from another taxonomy.

Please add service validation here.

9. **Avoid freezing rebalance schema before rebalance ships**

The SOTA table says `rebalance_drafts` is scaffolded but unused. I would remove it until the rebalance planner actually ships.

Same concern for execution-policy fields like `allow_sells`, `min_trade_amount`, `whole_shares_only`, and `rebalance_to`: if they are not exposed or used in this milestone, I would either defer them or make it very clear they are intentionally reserved for the next PR.

10. **Clarify `base_currency` semantics**

`base_currency` is persisted on `target_profiles`, but it does not appear to drive the current drift calculation. Drift uses the current app base currency at request time. Since Milestone 1 targets are percentage-based, I would either remove profile-level `base_currency` for now or define why it is stored and when it should be used.

If we keep `base_currency`, we should define it as the profile’s evaluation currency and use it consistently in drift/rebalance calculations. Otherwise the stored value can diverge from what the app actually displays.

10. **Docs referenced in the PR description**

The PR description references `docs/features/allocations/v1-spec.md` and `docs/features/allocations/sota-target-model-spec.md`, but I do not see those files in this PR branch. Please include them or update the description so reviewers can verify the referenced specs.

Overall, I like the feature and the direction. I would merge this as a focused **Targets + Drift** milestone after the data-model issues are fixed and the unshipped Calendar/Rebalance scaffolding is removed or deferred. Rebalance can come next as a separate milestone, starting with cash-flow-only planning.

Jonjon-prog:
Hi @afadil

Thanks for the thorough review, I happy also where its going. 
I addressed all 10 points in commit 738e0a6d.

**#1 — Calendar / Combined triggers removed**
Dropped both trigger types and their test cases. Only `manual` and `threshold` remain.

**#2 — Migrations merged**
`2026-05-28-000001_target_profile_calendar_trigger` deleted. 
Everything is now in a single migration `2026-05-25-000002_allocation_targets/up.sql`.

**#3 — NULL uniqueness bug fixed**
Active-scope unique index now uses `COALESCE(scope_id, '')` so multiple `scope_type = 'all'` rows with `scope_id IS NULL` correctly conflict.

**#4 — Scope CHECK constraint added**
Enforces `(scope_type = 'all' AND scope_id IS NULL) OR (scope_type IN ('account', 'portfolio') AND scope_id IS NOT NULL)` at the DB level.

**#5 — Taxonomy ID mismatch fixed**
`drift_service.rs` was matching on Rust field names (`"sectors"`, `"security_types"`) instead of the actual taxonomy IDs stored in `PortfolioAllocations`. 
Fixed to `"industries_gics"` and `"instrument_type"`.

**#6 — `profile_id` security fix in `save_nodes`**
Was using `n.profile_id` from user-supplied node data. 
Now always uses the validated route parameter.

**#7 — Zero-target nodes preserved**
Removed the `filter(targetBps > 0)` from the frontend save path. Zero-target nodes are stored, enabling "warn me when I hold X but target is 0" use cases.

**#8 — Category ID validation against taxonomy**
Implemented now, not deferred. `TargetProfileService` takes a `TaxonomyServiceTrait` dependency and validates each `category_id` belongs to the profile's taxonomy before persisting nodes.

**#9 — `rebalance_drafts` table removed**
No RebalanceService in M1. Removed from migration and Diesel schema.

**#10 — `base_currency` removed from `target_profiles`**
Went with Option A: drift uses the app-level `base_currency` setting at runtime. No per-profile currency field until M2 where it would actually be consumed.

**Bonus — Rebalance-only fields removed**
`rebalance_to`, `allow_sells`, `min_trade_amount`, `whole_shares_only` were in the original schema but unused in M1. Removed from migration, Rust models, Diesel schema, service, validation, and frontend types. Will reintroduce in M2 with the RebalanceService.

**Docs added**
`docs/features/allocations/` now contains `v1-spec.md`, `sota-target-model-spec.md`, and `portfolio-views-design.md`.

---

One open question: does the current UI (ProfileEditor, DriftDriversCard, TargetsTab) look right to you visually? Happy to iterate on layout/UX before merge if anything feels off.

PR #1031, opened 2026-05-29
fix(transfer-sync): cascade delete and update propagation for transfe…

Jonjon-prog:
Hey @afadil 👋

This PR addresses a set of UX pain points I ran into when working with linked internal transfers. 
The short version: when two activities share a `source_group_id` (a TRANSFER_IN + TRANSFER_OUT pair), editing or deleting one side used to leave the other completely out of sync — or worse, editing would silently create a duplicate pair. 
This fixes all of that.

---

## Summary

- **Cascade delete** — deleting either leg of a linked transfer pair now atomically deletes both legs (table view + datagrid), with a confirmation dialog warning that both sides will be permanently deleted.
- **Update propagation** — editing a linked transfer propagates `date`, `amount`, `currency`, and `notes` to the counterpart automatically. `accountId`, `activityType`, `quantity`, `unitPrice`, and `fee` are intentionally not propagated.
- **Edit form accounts** — opening the edit form on either leg now pre-populates both `From Account` and `To Account` correctly (previously only one field was filled, and in the wrong field for TRANSFER_IN).
- **Datagrid UX** — deleting a paired transfer row immediately marks both rows as pending delete before save, with a confirmation dialog matching the table-view behavior.

---

### Backend (`crates/core`, `crates/storage-sqlite`)

**`ActivityRepositoryTrait`** — new `find_transfer_counterpart(group_id, exclude_id)` method.

**`delete_activity` (repo)** — wraps the delete in a transaction that also deletes any counterpart sharing the same `source_group_id`.

**`bulk_mutate_activities` (repo)** — same cascade logic for the datagrid save path. Guard prevents double-delete when both legs are explicitly in the delete list.

**`delete_activity` (service)** — fetches the counterpart before calling the repo so both account IDs are included in the `ActivitiesChanged` event for correct cache invalidation.

**`update_activity` (service)** — after updating the primary leg, looks up the counterpart via `find_transfer_counterpart` and propagates `date`, `amount`, `currency`, `notes` via a direct repo call (bypasses asset resolution, preserves the counterpart's `activityType` and `status`).

### Frontend (`apps/frontend`)

**`use-activity-form.ts`** — internal transfers now check `isEditing` before deciding create vs. update. Previously the internal-transfer branch always created two new activities even when editing, causing duplication.

**`activity-form-config.ts`** — `getDefaults` for TRANSFER uses `counterpartAccountId` (a frontend-only field resolved at edit time) to pre-populate the missing account field.

**`activity-page.tsx`** — `handleEdit` looks up the counterpart in already-loaded activities and attaches its `accountId` to the `selectedActivity` state.

**`activity-data-grid.tsx`** — `handleDelete` for a paired activity opens `ActivityDeleteModal` (with `linkedTransfer=true`) before marking both legs for deletion.

**`use-save-activities.ts`** — after a bulk save, filters `localTransactions` using `result.deleted` (which now includes cascade-deleted rows) so counterpart rows disappear immediately without waiting for the refetch.

---

## Tests

**Rust unit tests — 2 new, all 1111 green** (`cargo test -p wealthfolio-core --lib`)

- `test_update_activity_propagates_to_transfer_counterpart` — verifies `amount`, `date`, and `notes` are propagated to the counterpart, and that `account_id` and `activity_type` are not touched.
- `test_delete_activity_cascades_transfer_pair` — verifies that deleting one leg removes both legs and that the `ActivitiesChanged` event includes both account IDs.

Side-fix: `MockActivityRepository::update_activity` now also updates `activity_date` (was missing, masking a gap in mock coverage).

**Frontend** : `pnpm test` — existing suite, all green.

**Manual smoke**:
- Table view: delete TRANSFER_OUT → "both sides" dialog → confirm → both gone ✓
- Table view: edit TRANSFER_OUT → change amount → TRANSFER_IN amount synced, account unchanged ✓
- Table view: edit TRANSFER_IN → both From/To accounts pre-populated ✓
- Datagrid: delete TRANSFER_OUT → confirmation dialog → both rows marked pending → save → both gone ✓
- Datagrid: edit TRANSFER_OUT → update comment → no duplication ✓

---

## Known limitation — investment ↔ spending transfers

Internal transfer pairs currently only work between two investment accounts. Transfers between an investment account and a spending account (e.g. moving cash from a brokerage to a chequing account tracked in the spending module) are not yet supported — the two sides live in separate activity scopes and the pairing logic doesn't cross that boundary today.

This is the next piece of work, if you want I can do It here or in a new branch. 

---

## Checklist


Jonjon-prog:
Hey @afadil 👋

Following up on the known limitation I mentioned at the bottom of this PR — transfers between investment accounts and spending accounts.

**Why this matters**

These are real, traceable money movements — moving $500 from a brokerage to a chequing account (or the reverse, funding an investment from a cash account) is exactly the kind of transaction that should be fully tracked. Right now, the user has to manually record two unlinked activities: one in the investment module, one in the spending module. They'll never sync if one side changes, and there's no way to see them as a single movement.

The sync infrastructure from this PR (`source_group_id` pairing, cascade delete, update propagation) is already there — we'd just be extending it across the module boundary.

---

**Questions before starting a new branch**

1. **Scope confirmation** — Is it intentional that the spending module supports cash ↔ investment transfers? Or is the spending tracker meant to stay purely within cash accounts?

2. **Entry point** — Should the user create the pair from the spending form (cash side first, picker for the investment account counterpart), the investment activity form (investment side first), or both?

3. **Auto-pairing existing activities** — If a user already has unlinked TRANSFER_IN/OUT on both sides (manually entered), should we offer a way to link them retroactively (like the existing "Link Transfer" modal), or is new-creation-only sufficient for v1?

Happy to start a new branch if the direction is green. Let me know!

PR #1041, opened 2026-06-02
feat(allocation): review fixes and target workspace

afadil:
## Summary

This PR reworks the allocation-targets experience end to end. The standalone **Allocation Targets** page is retired and folded into the **Portfolio Insights** workspace as a new **Overview** tab, with a dedicated current → details → targets editing flow. The legacy standalone **Holdings Insights** page is removed in favor of this Overview surface. Alongside the redesign, it hardens the Rust drift/target model, storage, and validation, wires the target APIs across both the Tauri desktop and Axum web runtimes, and stabilizes chart rendering (colors, dark mode, cash avatars, drilldowns).

### Portfolio Insights — Overview tab (replaces the old Holdings tab/page)
The first tab of Portfolio Insights is redesigned: the old "Holdings" tab is replaced by a new "Overview" surface. It composes a **value strip** (totals + currency exposure), a **portfolio explorer** (multi-lens allocation breakdown with drilldowns into the underlying holdings), and a **target rails card** (current vs. target drift at a glance). `/allocation-targets` now redirects here and the standalone "Allocation" nav entry is removed.

CleanShot 2026-06-01 at 20 34 14@2x

### Target workspace — Current / Details / Targets
A scoped, in-page editing flow replaces the old standalone profile editor. Users move between the current allocation, the full current-vs-target analysis, and the target weight editor without leaving Insights. Unsaved changes are guarded by a confirmation dialog.

CleanShot 2026-06-01 at 20 34 34@2x

### Target weight editor + model presets
New weight editor, drift-band slider, and an expanded model-preset picker (curated allocation models) replace the previous node editor.

CleanShot 2026-06-01 at 20 34 52@2x

### Frontend — Insights / Overview (new)
- Add `insights/overview/` page: `overview-page.tsx`, `portfolio-explorer.tsx`, `target-rails-card.tsx`, `value-strip.tsx`, and `allocation-derivations.ts` (+ tests).
- Switch `portfolio-insights.tsx` first tab from "Holdings" to "Overview" (`OverviewPage`); route `/allocation-targets` → `/insights?tab=overview`; remove the "Allocation" item from app navigation.
- Remove the legacy standalone holdings insights surface: delete `holdings-insights-page.tsx` and the now-orphaned `/holdings-insights` route, plus its only-used-here components (`cash-holdings-widget`, `currency-chart`, `sectors-chart`, `segmented-allocation-bar`) and the `allocation-drilldown` util/test.

### Frontend — Allocation targets workspace
[...]

PR #1036, opened 2026-05-31
feat(allocation-targets): Cash-Flow Rebalance Planner — Milestone 2 PR-A

Jonjon-prog:
Hello @afadil

PR 2/2 for the Allocation Targets feature — the Cash-Flow Rebalance Planner from Milestone 2 (PR-A: corrections before algorithm rewrite).

Rebased on top of your UI refactor (PR #1041).

---

## What's in this PR (PR-A)

- ✅ Removed `rebalance_drafts` entirely — stateless M2, export/copy only
- ✅ Cash model fixed — `total_value` stays constant (tracked cash moves within portfolio, not added)
- ✅ Cash auto-read from account holdings — shows "Cash in scope: X", pre-fills deploy input, user can override
- ✅ Backend enforces `cash_to_deploy ≤ cash_in_scope`
- ✅ `max_drift_bps_after` now filters required rows only (matches DriftService semantics)
- ✅ `available_cash < 0` rejected with validation error
- ✅ Backend warning messages — numbers only, no hard-coded currency symbol
- ✅ Debug logs — no financial amounts
- ✅ CSV export — all fields escaped via `csvCell()`
- ✅ Web mode fixed — `targetId` was sent as `profileId` in the web adapter, breaking web mode
- ✅ Empty migration `2026-05-29-000001_rebalance_planner` removed (no schema changes in this PR)
- ✅ Rebalance UI uses allocation-target-colors calm palette

**Algorithm kept as-is for PR-A.** PR-B will rewrite with the exposure-aware drift planner.

---

### V1 Spec — Milestone 2

| Requirement                                  | Status                                    |
| -------------------------------------------- | ----------------------------------------- |
| Available cash input                         | ✅ Done                                   |
| Cash-flow-only (no forced sells)             | ✅ Done                                   |
| Proportional buy within underweight sleeves  | ✅ Done                                   |
| Sleeve-level fallback when no holdings       | ✅ Done + `NoBuyCandidate` warning        |
| `whole_shares_only` rounding mode            | ✅ Done — per-profile toggle, default off |
| `min_trade_amount` filter                    | ✅ Done — sub-threshold trades dropped    |
| `rebalance_to` (nearest_band / exact_target) | ✅ Done                                   |
| Missing quote warning                        | ✅ Done (`MissingQuote`)                  |
| CSV export of suggested trades               | ✅ Done                                   |
| Save rebalance draft                         | ❌ Removed — stateless M2, export/copy only |
[...]

PR #1048, opened 2026-06-03
feat(allocation-targets): Exposure-Aware Drift Planner — Milestone 2 PR-B

Jonjon-prog:
Hello @afadil 👋

PR-B for the Allocation Targets feature : 
the exposure-aware **Drift Planner** from Milestone 2, rewriting the algorithm as we discussed in #1036.

This is the final piece of the rebalance planner. 
I documented the algorithm fully in `docs/features/allocations/rebalance-algorithm.md` and left 3 open questions at the bottom where I'd like your call before we lock it in. 🙂

---

## Summary

Replaces the proportional cash-flow planner (PR-A, #1036) with an **exposure-aware greedy optimiser**.

The previous algorithm treated each taxonomy category as an isolated sleeve and split cash proportionally to current holdings. That breaks for assets with split classification (e.g. a global ETF that spans US / developed / emerging markets) — buying one such ETF affects several categories at once, but the proportional model couldn't see that coupling.

This PR introduces `DriftPriorityOptimizer`: each iteration buys 1 share of the asset that maximises **total drift reduction per dollar** across *all* taxonomy categories. A single buy updates every affected category's exposure simultaneously.

## Design

```
Build exposure vectors (per-share value per category, from AllocationService contributions)
Greedy loop:
  score(asset) = (Σ|drift_bps| before − Σ|drift_bps| after) / price
  pick highest score > 0   (tie-break: price ASC)
  apply 1-share buy, repeat until no asset improves drift or cash exhausted
Post-process:
  aggregate shares → 1 trade per asset
  drop trades below min_trade_amount
  recompute after-drift from kept trades only
```

- **Greedy, not LP/MILP** — deterministic, explainable, fast. Solver-compatible via the new `RebalanceOptimizer` trait (a future `MilpOptimizer` / `TaxAwareOptimizer` can slot in without touching `RebalanceService`).
- **Phase 1 + Phase 2 removed** — the greedy 1-share loop subsumes both proportional allocation and residue absorption.
- **`cash_used`** = sum of recommended trade amounts after the min-trade filter, so it always matches what the user will actually execute. After-drift is recomputed from the kept trades.
- **`after_bps_by_category`** added to `RebalancePlan` so the UI Before/Target/After bars are accurate for multi-category ETFs (previously re-derived from the primary trade category, which was wrong for split assets).

## Classification rules (from your direction in #1036)

| Case | Behaviour |
| ---- | --------- |
| No assignments on active taxonomy | Skip candidate, warn `UnclassifiedAsset` |
| Partial weights (<100%) | Use known exposure as-is, warn `PartialClassification`, no normalisation |
| Weights only on another taxonomy | Treated as unclassified → skip + warn |
| Empty category | `NoBuyCandidate` + sleeve-level dollar suggestion (no fake ticker) |
| Holdings outside target sleeves | Informational only (sell = M3) |
| Cash holdings | Excluded from candidates |

5 of these 6 rules fall out naturally from the `AllocationService` contributions layer (which already emits a `__UNKNOWN__` row for partial/unassigned classification).

## Changes

- **`optimizer.rs`** (new) — `RebalanceOptimizer` trait, `DriftPriorityOptimizer`, input types.
- **`rebalance_service.rs`** — rewritten orchestration: fetch holdings + contributions, build candidates, delegate to optimiser. Authoritative tracked-cash enforcement from PR-A preserved.
- **`model.rs`** — `after_bps_by_category` on `RebalancePlan`; new warning kinds `UnclassifiedAsset` / `PartialClassification`; dropped now-unused `WholeShareResidue`.
- **Frontend** — `rebalance-tab.tsx`: "Drift planner" badge, new warning labels, Before/After bars use backend after-drift; CSV export now carries plan metadata (profile, date, cash, currency) with currency-aware formatting.
- **Docs** — `rebalance-algorithm.md` rewritten for the exposure-aware model.

## Tests

- `cargo test -p wealthfolio-core` — 1214 pass (50 in `allocation_targets`, incl. multi-category ETF coupling, classification edge cases, whole-share, nearest-band, cash enforcement).
- `cargo clippy -p wealthfolio-core --lib -- -D warnings` — clean.
- Frontend type-check clean on touched files.

## Questions for you @afadil

1. **Weights >100%** — you wrote "block the plan" for over-allocated assets, but your allocation hardening (#1041, `contribution_shares_for_holding`) already normalises >100% to 100% silently, and the drift report / Overview tab consume that normalised data everywhere. Blocking only in the rebalance planner would be inconsistent with the rest of the app. I aligned the planner with the existing normalisation rather than blocking. If you want a hard block, I think the right place is **write-time validation on `AssetTaxonomyAssignment`** (a separate data-quality PR), not the planner. Agree? (Noted in §9 of the algorithm doc.)

2. **Top-up / starvation warning** — in the proportional model we emitted a `WholeShareResidue` warning when an expensive holding's per-holding budget was below its share price. The exposure-aware model has no per-holding proportion concept (it minimises category drift), so structural starvation isn't really a thing — undeployed cash just surfaces as `cash_remaining`. I dropped the warning. Do you want a per-asset "add $X to fund 1 more share" hint back, or is `cash_remaining` enough?

3. **Fractional last slice** — the loop stops when remaining cash < cheapest candidate's full share price; the gap (< 1 share) stays as `cash_remaining`. Fractional mode could deploy that final slice as a sub-share buy. Left as a nice-to-have — worth doing now or defer?

---

## Checklist

### DIFF
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -2408,7 +2408,12 @@ export interface DriftHoldingsReport {
   rows: DriftHoldingRow[];
 }
 
-export type RebalanceWarningKind = "missing_quote" | "no_buy_candidate" | "whole_share_residue";
+export type RebalanceWarningKind =
+  | "missing_quote"
+  | "no_buy_candidate"
+  | "whole_share_residue"
+  | "unclassified_asset"
+  | "partial_classification";
 
 export interface RebalanceWarning {
   kind: RebalanceWarningKind;
@@ -2438,4 +2443,5 @@ export interface RebalancePlan {
   maxDriftBpsAfter: number;
   trades: SuggestedManualTrade[];
   warnings: RebalanceWarning[];
+  afterBpsByCategory: Record<string, number>;
 }
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -2411,7 +2411,6 @@ export interface DriftHoldingsReport {
 export type RebalanceWarningKind =
   | "missing_quote"
   | "no_buy_candidate"
-  | "whole_share_residue"
   | "unclassified_asset"
   | "partial_classification";
 
--- a/apps/frontend/src/pages/allocation-targets/components/rebalance-tab.tsx
+++ b/apps/frontend/src/pages/allocation-targets/components/rebalance-tab.tsx
@@ -271,7 +271,6 @@ function KpiStrip({ plan, currency }: { plan: RebalancePlan; currency: string })
 const WARN_LABEL: Record<string, string> = {
   missing_quote: "Missing quote",
   no_buy_candidate: "No buy candidate",
-  whole_share_residue: "Residual cash",
   unclassified_asset: "Unclassified",
   partial_classification: "Partial classification",
 };
--- a/apps/frontend/src/pages/allocation-targets/components/rebalance-tab.tsx
+++ b/apps/frontend/src/pages/allocation-targets/components/rebalance-tab.tsx
@@ -70,36 +70,18 @@ function parseCashValue(value: string): number {
 }
 
 function computeSleeveSummary(driftReport: DriftReport, plan: RebalancePlan) {
-  // Total value is constant — cash moves within the portfolio, not added from outside.
-  const total = driftReport.totalValue;
   const colorMap = buildAllocationTargetColorMap(driftReport.rows);
-  const deployedByCategory = new Map<string, number>();
-  for (const trade of plan.trades) {
-    deployedByCategory.set(
-      trade.categoryId,
-      (deployedByCategory.get(trade.categoryId) ?? 0) + trade.estimatedAmount,
-    );
-  }
-
   return driftReport.rows
     .filter((r) => r.status !== "not_targeted")
-    .map((row, i) => {
-      // Deploying cash moves value out of the cash sleeve into buy sleeves.
-      // Total value is unchanged, so the cash row sheds cash_used.
-      const deployed = deployedByCategory.get(row.categoryId) ?? 0;
-      const afterValue = row.isCash
-        ? Math.max(row.currentValue - plan.cashUsed, 0)
-        : row.currentValue + deployed;
-      const afterBps = total > 0 ? Math.round((afterValue / total) * 10000) : 0;
-      return {
-        categoryId: row.categoryId,
-        categoryName: row.categoryName,
-        color: allocationTargetColorForRow(row, colorMap, i),
-        currentBps: row.currentBps,
-        targetBps: row.targetBps,
-        afterBps,
-      };
-    });
+    .map((row, i) => ({
+      categoryId: row.categoryId,
+      categoryName: row.categoryName,
+      color: allocationTargetColorForRow(row, colorMap, i),
+      currentBps: row.currentBps,
+      targetBps: row.targetBps,
+      // Use backend-computed after-bps (accounts for multi-category ETF exposure).
+      afterBps: plan.afterBpsByCategory[row.categoryId] ?? row.currentBps,
+    }));
 }
 
 function csvCell(value: string): string {
@@ -273,6 +255,8 @@ const WARN_LABEL: Record<string, string> = {
   missing_quote: "Missing quote",
   no_buy_candidate: "No buy candidate",
   whole_share_residue: "Residual cash",
+  unclassified_asset: "Unclassified",
+  partial_classification: "Partial classification",
 };
 
 function Warnings({ items }: { items: RebalanceWarning[] }) {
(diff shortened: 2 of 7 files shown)


afadil:
Thanks @Jonjon-prog. quick review findings:

1. **High: `NoBuyCandidate` trades can overspend cash**

   `crates/core/src/portfolio/allocation_targets/optimizer.rs` sizes each uncovered-category manual trade with `shortfall.min(available_cash)`, not remaining cash. If multiple uncovered categories exist, each can consume the full cash amount.

   I reproduced a case where `available_cash = 1000` and the plan returns `cash_used = 2000`, `cash_remaining = -1000`.

2. **High: manual category trades are excluded from after-bps**

   The after-drift recompute only applies trades with an `asset_id`, but still subtracts full `cash_used` from cash. For no-ticker/manual category trades, cash leaves the cash sleeve, but the target category never receives the value.

   I reproduced a `$1000` bonds manual trade where `after_bps_by_category["bonds"]` stays `0` instead of `1000`.

3. **High: fractional mode does not actually deploy fractional shares**

   `whole_shares_only` is passed into the optimizer, but the greedy loop always buys one full share and skips candidates when remaining cash is below the full share price.

   I reproduced fractional mode with `$50` cash and a `$100` ETF: result was no trade, not `0.5` shares. If fractional support is intentionally deferred, the flag/docs should make that explicit.

4. **Medium: `NearestBand` still optimizes to exact target**

   `total_drift` compares against `target_bps` directly. `NearestBand` is used for no-candidate shortfall detection, but not for greedy scoring.

   I reproduced a 60% equity / 70% target / 5% band case: it spent `$1000` to hit exact target instead of `$500` to stop at the band edge.

5. **Medium: one-share loop has a scalability tail**

   The greedy loop buys exactly one share per iteration, so runtime scales with `available_cash / min_price`. Typical cases are probably fine, but cheap assets plus large cash balances can make the command noticeably slow.

6. **Low/Medium: equal-price ties are not deterministic**

   Candidate construction iterates a `HashMap`, then sorting by price preserves that input order for equal-price candidates. Warning order and tie outcomes can vary run-to-run. Sorting by a stable key like `(price, symbol, asset_id)` would make plans reproducible.

Jonjon-prog:
Thanks for the thorough review @afadil — great catches, all six were real. 🙏

Pushed one commit per finding (each with a regression test reproducing your exact repro case). `cargo test -p wealthfolio-core` → 1220 pass, clippy clean.

| # | Finding | Fix |
| - | ------- | --- |
| 1 | `NoBuyCandidate` overspend | Manual sleeve trades now draw from the cash left after the greedy buys, decremented per trade, so multiple uncovered categories can never collectively overspend. The advisory now reports the true shortfall instead of the misleading per-category cash cap. (`a0a02a0`) |
| 2 | Manual trades missing from after-bps | Manual (no-ticker) trades now credit their target category by `estimated_amount` in the after-drift recompute, so the filled category is reflected. (`7fe23dc`) |
| 3 | Fractional mode not deploying fractions | `whole_shares_only` is now honoured: when fractional is enabled, a final fractional slice deploys the leftover into the candidate that reduces drift most, sized `cash / price` and capped so no category overshoots its desired value. Whole-share mode is unchanged. (`f4c93c3`) |
| 4 | `NearestBand` optimizing to exact target | New band-aware `category_drift` helper: `NearestBand` only counts the distance outside `[target ± band]`, so a category inside its band contributes zero drift and the greedy stops at the band edge. Your 60% / 70% / 5% case now spends $500. (`c7e380269`) |
| 5 | One-share loop scalability | The greedy now batches buys of the chosen candidate within each linear drift region (up to the next category bend). Per-share improvement is constant across that region, so the batch equals that many single-share iterations — but a $1 asset with $8000 cash now resolves in one step instead of 8000 iterations. (`e36d005`) |
| 6 | Non-deterministic ties | Candidates are sorted by `(price, symbol, asset_id)` and holdings are iterated in sorted order when building candidates, so tie outcomes and warning order are reproducible. (`d26e0e7`) |

One note on #5: the batching is exactly equivalent to the strict one-share greedy for the dominant slow case (a single driving category). In a coupled multi-category case where another candidate would overtake the current best partway through a linear region, the plan could diverge marginally from strict one-share — none of the existing tests are affected. If you'd rather have zero divergence I can swap in a slope-based crossover bound that's provably faithful (~40 extra lines) — happy to do that if you prefer.

Still on your call from the original PR description: **Q1 (>100% weights)** — block in the planner vs. align with the existing `AllocationService` normalisation (I lean toward a separate write-time validation PR), and **Q2 (top-up / starvation hint)** — keep `cash_remaining` as-is or add a per-asset "add $X for one more share" hint. Let me know and I'll fold those in.

afadil:
Thanks @Jonjon-prog. I pushed a follow-up.

Decisions:
- Q1: no planner-only block for >100% weights; this stays aligned with `AllocationService`.
- Q2: no `WholeShareResidue`; `cash_remaining` is enough for this PR.
- Q3: fractional mode is now fractional throughout.
- Batching: adjusted to preserve strict greedy equivalence.

Also fixed the currency-price sizing edge case and the min-trade/manual-sleeve cash interaction.

I'll merge this PR, if you see any other adjustement or fixes please let me know

PR #1060, opened 2026-06-05
feat(holdings): add Avg Cost and Allocation optional columns

ki-adi-mundi:
Description:
  ## Problem
  The holdings table doesn't show average cost per share or the holding's
  weight within the currently selected portfolio/account scope.

  ## Solution
  Two new optional columns, hidden by default and toggleable via the
  Columns dropdown (same as Book Cost, Currency, etc.):

  - **Avg Cost** — average cost per share (`costBasis.local / quantity`),
    displayed in the asset's local currency, supports balance privacy,
    sortable.
  - **Allocation** — portfolio weight (`weight` field from the backend),
    already scoped to the active AccountScope filter, so it reflects the
    correct percentage when a Portfolio or multi-account filter is selected.

  No backend changes needed — both values are already available on the
  `Holding` object.
  
### DIFF
--- a/apps/frontend/src/pages/holdings/components/holdings-table.tsx
+++ b/apps/frontend/src/pages/holdings/components/holdings-table.tsx
@@ -123,6 +123,8 @@ export const HoldingsTable = ({
           symbolName: false,
           holdingType: false,
           bookValue: false,
+          avgCost: false,
+          portfolioWeight: false,
         }}
         defaultSorting={[{ id: "symbol", desc: false }]}
         scrollable={true}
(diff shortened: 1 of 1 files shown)


afadil:
two issues to fix before merge:

1. The new columns are only hidden by default for users without existing persisted table visibility. `DataTable` reads `holdings-table:column-visibility` from localStorage verbatim, so existing users whose saved visibility object lacks `avgCost` / `portfolioWeight` will see both new columns immediately. Please merge persisted visibility with the new defaults or bump/migrate the table storage key.

2. `Avg Cost` is incorrect for option holdings. `costBasis.local / quantity` produces a per-contract value because option cost basis includes `contractMultiplier`, while the existing “Today’s Price” column shows the per-share premium. For example, 5 contracts at $3 with multiplier 100 displays `$302` instead of roughly `$3.02`. Please divide by `quantity * (contractMultiplier ?? 1)` or make the labeling/display explicitly per-contract.

PR #1061, opened 2026-06-05
feat(asset): show avg cost reference line on price chart

ki-adi-mundi:
## Problem
The asset price chart has no visual reference for the user's average cost per share,
making it hard to gauge at a glance whether the current price is above or below
the break-even level.

## Solution
Add a dashed horizontal reference line at the average cost price (`costBasis.local / quantity`)
on the holding detail price chart.

- Visible whenever the user holds a position in the asset (`avgCost > 0`)
- Hidden for assets with no holding or zero cost basis (e.g. FX rates, unwatched assets)
- Labelled `Avg $xxx.xx` at the right edge of the chart
- Always on — no extra toggle needed, unlike activity markers which are more situational

## Changes
- `history-chart-symbol.tsx` — add `avgCost?: number` prop + `ReferenceLine` from recharts
- `asset-history-card.tsx` — thread `avgCost` prop through to `HistoryChart`
- `asset-profile-page.tsx` — pass `symbolHolding.averagePrice` (already computed) to `AssetHistoryCard`

No backend changes. No new dependencies (recharts already used in the chart).

### DIFF
--- a/apps/frontend/src/components/history-chart-symbol.tsx
+++ b/apps/frontend/src/components/history-chart-symbol.tsx
@@ -6,6 +6,7 @@ import {
   Area,
   AreaChart,
   ReferenceDot,
+  ReferenceLine,
   ResponsiveContainer,
   Tooltip,
   XAxis,
@@ -55,6 +56,7 @@ export default function HistoryChart({
   interval,
   activityMarkers = [],
   onActivityMarkerClick,
+  avgCost,
   height = 350,
 }: {
   data: HistoryChartData[];
@@ -62,6 +64,7 @@ export default function HistoryChart({
   height?: number;
   activityMarkers?: HistoryChartActivityMarker[];
   onActivityMarkerClick?: (marker: HistoryChartActivityMarker) => void;
+  avgCost?: number;
 }) {
   const [hoveredMarker, setHoveredMarker] = useState(false);
   const markerByTimestamp = useMemo(() => {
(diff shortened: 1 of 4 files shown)


afadil:
Hi @ki-adi-mundi , thanks for this one, I found a few issues that should be addressed before merging:

1. The avg-cost line bypasses balance privacy. `avgCost` is user-specific cost-basis data, but `HistoryChart` renders the line and exact `Avg $...` label even when balances are hidden. This should either hide the line entirely or hide the label/value when privacy mode is enabled.

2. The chart value is wrong for options / multiplied instruments. `symbolHolding.averagePrice` is computed as `costBasis / quantity`, but quote history is unit price and backend valuation applies `contractMultiplier` separately. For standard options this can put the line around 100x too high. The chart reference should use `costBasis / (quantity * contractMultiplier)` or skip multiplied instruments.

3. `ReferenceLine` defaults to `ifOverflow="discard"`, so if avg cost is outside the current price-domain, the line disappears. That breaks the main goal of seeing whether price is above/below break-even. Consider `ifOverflow="extendDomain"` or explicitly include avg cost in the Y-domain.

4. The new hidden holdings columns may show for existing users. `holdings-table:column-visibility` is persisted, and older stored visibility objects won’t include `avgCost` / `portfolioWeight`, so they’ll default visible. This needs a visibility migration, default merge, or storage key bump.

PR #1062, opened 2026-06-05
fix(holdings): persist account scope filter across navigation

ki-adi-mundi:
## Problem
Selecting a portfolio or account filter on the Holdings page (e.g. "USA Stocks")
and then navigating into a holding detail resets the filter to "All Accounts" when
the user presses back. This forces users to re-select their filter on every visit.

## Solution
Replace `useState` with the existing `usePersistentState` hook (already used for
column visibility, sorting, etc. in `DataTable`) to persist the selected
`AccountScope` in `localStorage` under the key `holdings:account-scope`.

The selected scope now survives:
- Back-button navigation from the holding detail page
- Page refreshes
- Tab switches (Assets ↔ Investments ↔ Liabilities)

## Changes
- `holdings-page.tsx` — one-line change: `useState` → `usePersistentState`

No new dependencies. `usePersistentState` is already imported in this file.

### DIFF
--- a/apps/frontend/src/components/history-chart-symbol.tsx
+++ b/apps/frontend/src/components/history-chart-symbol.tsx
@@ -6,6 +6,7 @@ import {
   Area,
   AreaChart,
   ReferenceDot,
+  ReferenceLine,
   ResponsiveContainer,
   Tooltip,
   XAxis,
@@ -55,6 +56,7 @@ export default function HistoryChart({
   interval,
   activityMarkers = [],
   onActivityMarkerClick,
+  avgCost,
   height = 350,
 }: {
   data: HistoryChartData[];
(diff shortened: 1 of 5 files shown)


afadil:
Thanks for the PR. I think this needs a couple of fixes before merge:

1. Persisting `accountFilter` restores the visible account scope, but `selectedAccount` still starts as `null` and is only set when the selector changes. After reload/back navigation with a saved account scope, holdings are filtered correctly, but HOLDINGS-mode account actions remain disabled because `canEditHoldings` has no selected account. Please derive `selectedAccount` from `accountFilter` + `accounts`, or sync it when accounts load.

2. The avg-cost reference line uses `symbolHolding.averagePrice` directly against quote-history `close` values. For options, average cost includes the contract multiplier while quote close is the premium, so the line is off by 100x for standard contracts. Please divide by `contractMultiplier` for charting or suppress the line for options.

3. `apps/frontend/src/pages/holdings/holdings-page.tsx` needs Prettier formatting; `prettier --check` flags the new `usePersistentState` line.

afadil:
implemented in v3.8

PR #1067, opened 2026-06-05
feat(spending): add transfer support from spending tab

Jonjon-prog:
Hey @afadil 👋

This is a follow-up to #1031 
The paired transfer infrastructure was already in place (cascade delete, update propagation, `source_group_id` sync), but there was no way to *initiate* a linked transfer from the Spending tab. Users had to go to the Investments tab to do it, which wasn't obvious at all.

---

## What this does

When a user is in **Spending → Add Transaction**, instead of having TRANSFER_IN / TRANSFER_OUT in the type dropdown (which created standalone, unlinked legs), there's now a **"Transfer between accounts"** button. Clicking it opens the standard transfer form — the same one used on the Investments tab.
With the spending account pre-filled as "From Account". The user picks the destination account and saves. Both legs are created atomically with a shared `source_group_id`, so the existing sync from #1031 kicks in automatically (update propagation, cascade delete, everything).

Changes are minimal:

- **`cash-activity-form.tsx`** — hides TRANSFER_IN/OUT from the type dropdown in creation mode; adds a "Transfer between accounts" button with an `onTransferClick` callback
- **`activity-form.tsx`** — adds a `hidePicker` prop to skip the type carousel when the caller has already determined the type
- **`spending-transactions-tab.tsx`** — wires up the full transfer flow (account list, open/close state, `ActivityForm` instance that remounts fresh on each open)

Editing existing linked or standalone transfer activities from the Spending tab still works as before.

---

## Tests

No unit tests added — the changes are purely UI/wiring with no new business logic. The existing `use-activity-form.test.ts` suite (which covers the transfer pair submission path we rely on) passes unchanged.

**Manual smoke:**
- Spending → Add Transaction → Transfer button → Transfer form opens with spending account pre-set as From Account ✓
- Pick a destination invest account + amount → both legs created, pair linked via `source_group_id` ✓
- Edit the transfer from Spending tab → update propagates to counterpart (PR #1031 sync) ✓
- Delete from Spending tab → both legs removed (PR #1031 cascade) ✓
- Existing non-transfer cash activities (Withdrawal, Deposit, etc.) unaffected ✓

---

## Small question on the saving color 🎨

One thing I noticed while working on this: a cross-scope TRANSFER_OUT (spending → invest) gets `cash_flow_bucket = "saving"` and renders in olive (`#6B8E54`) in the amount column. I understand the intent — it's savings, not a loss — and I've left it exactly as-is.

That said, since olive is close to green visually, some users might find it a bit ambiguous next to income (which is also greenish). Would you consider a neutral color.
Something like amber/gold or a muted blue — to say "this money moved somewhere positive" without reading as income? 
Totally your call, just flagging it in case you had a specific palette in mind for the saving bucket.

---

## Checklist

### DIFF
--- a/apps/frontend/src/features/spending/components/cash-activity-form.tsx
+++ b/apps/frontend/src/features/spending/components/cash-activity-form.tsx
@@ -89,9 +89,16 @@ interface CashActivityFormProps {
     categoryTaxonomyId?: string;
     categoryId?: string;
   };
+  /** Called when user selects Transfer in creation mode; receives the selected spending account id */
+  onTransferClick?: (accountId: string) => void;
 }
 
-export function CashActivityForm({ open, onOpenChange, activity }: CashActivityFormProps) {
+export function CashActivityForm({
+  open,
+  onOpenChange,
+  activity,
+  onTransferClick,
+}: CashActivityFormProps) {
   const isEditing = !!activity?.id;
   const qc = useQueryClient();
   const { accounts } = useAccounts({ filterActive: false });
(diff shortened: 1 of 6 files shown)


afadil:
@Jonjon-prog Thanks!

PR #1075, opened 2026-06-07
Improve spending month selection

afadil:
## Summary
- Replace the rolling 30-day dashboard shortcut with last-month and custom month selection semantics.
- Add a mobile bottom-sheet month selector with larger touch targets while keeping the desktop popover compact.
- Add budget widget month navigation and keep budget-card links/ranges tied to the selected budget month.

## Validation
- pnpm format:check
- pnpm lint (passes with existing repo warnings)
- pnpm type-check
- pnpm test
- Browser smoke test at mobile width for the spending month selector

### DIFF
--- a/apps/frontend/src/features/spending/components/budget-line-chart-card.tsx
+++ b/apps/frontend/src/features/spending/components/budget-line-chart-card.tsx
@@ -12,8 +12,15 @@ import type { DayBucket } from "../types/report";
 
 type Status = "ok" | "warn" | "over";
 type PacePoint = { day: number; value: number };
+type BudgetToday = { year: number; month: number; day: number };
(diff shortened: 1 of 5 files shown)


Jonjon-prog:
Hey @afadil  :) 

So I tested this,  and I got a little feedback in term of user experience. 

I think we should linked the "this month / last month" selection to the monthly budget. Like this if the user switch from one to an other he can easily see the widget also changing. 
This avoid a double manipulation. 
We can linked also this with the month selection. 
We should just avoid that linked when user is selecting 3M, 6M, YTD and 1Y as it would need more calculation to average those budget on track or not. 

Is this making sense to you ?

PR #1078, opened 2026-06-08
feat(import): split CSV vs existing duplicate types in review UI

Jonjon-prog:
Hey @afadil. 

I hope you are doing great. 

Small UX improvement to the import review step .

When importing activities, duplicates can come from two very different sources — a row appears twice in the same CSV file, or a row already exists in the portfolio. Until now both were lumped into a single "duplicates" count, leaving the user to hover on each dot tooltip to understand which is which.

## Changes

**Banner**
- Two separate clickable badges instead of one: `42 CSV duplicates` and `36 existing duplicates`
- Title and description are specific when only one type is present:
  - DB-only: *"3577 activities already in your portfolio"* + explanation that they will be skipped with an "Import anyway" override path
  - CSV-only: *"42 duplicates within this file"* + explanation
- Mixed case shows a breakdown in the description: *"All 3629 activities are importable — 5 warnings, 42 CSV duplicates, 36 existing duplicates to review."*

**Status filter (toolbar)**
- `"Duplicates"` entry replaced by `"CSV duplicates"` and `"Existing duplicates"` as separate filterable entries
- Each filters and groups correctly (CSV filter still shows the original row as non-selectable context)
- "Import anyway" bulk action available on both sub-type filters

No changes to the tooltip (already correct), no changes to `DraftActivityStatus` values, no backend changes.

## Test plan

- [ ] Import a CSV with intra-file duplicates only → banner shows "X duplicates within this file", single CSV badge, no existing-duplicate badge
- [ ] Import a CSV where rows already exist in DB → banner shows "X activities already in your portfolio", single existing-duplicate badge
- [ ] Import a CSV with both types → banner description lists both counts
- [ ] Click each badge → filters to correct sub-set of rows
- [ ] Status filter dropdown shows correct entries depending on what types are present
- [ ] "Import anyway" available when filtering by either duplicate sub-type
- [ ] Clear filters resets correctly

---

## Checklist

### DIFF
--- a/apps/frontend/src/pages/activity/import/steps/review-step.tsx
+++ b/apps/frontend/src/pages/activity/import/steps/review-step.tsx
@@ -217,8 +217,27 @@ export function ReviewStep() {
       nonDuplicateFilters.size === 0 && (hasDuplicateCsvFilter || hasDuplicateDbFilter);
 
     if (isDuplicateOnlyFilter) {
-      const filtered = draftsMatchingFacetFilters.filter(matchesStatusFilter);
-      const groupedDrafts = buildDuplicateReviewRows(filtered);
+      const duplicateRows = draftsMatchingFacetFilters.filter(matchesStatusFilter);
+
+      // buildDuplicateReviewRows looks up source rows (the original of a CSV duplicate)
+      // from the dataset it receives. Re-attach any source rows from the full facet-filtered
+      // set so they appear as non-selectable context when filtering by duplicate_csv.
+      let rowsForGrouping = duplicateRows;
+      if (hasDuplicateCsvFilter) {
+        const byLineNumber = new Map(draftsMatchingFacetFilters.map((d) => [d.rowIndex + 1, d]));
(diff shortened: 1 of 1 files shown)


afadil:
Thanks @Jonjon-prog. Codex reported this finding:
 CSV duplicate filtering drops the original context row. In review-step.tsx (line 219), the PR filters to duplicate rows before calling buildDuplicateReviewRows. That helper builds its source-line lookup from the rows it receives, so the original row for a CSV duplicate is usually absent and cannot be inserted as non-selectable context. Repro: row 1 valid, row 2 status: "duplicate", duplicateOfLineNumber: 1; clicking CSV duplicates shows only row 2. This also affects selecting both duplicate subtype filters. Fix by adding the selected CSV duplicates’ source rows back before grouping, or by changing the helper to receive both duplicateRows and the full facet-filtered row universe.

Jonjon-prog:
Good catch @afadil , thanks! 
The fix is in the latest commit . 
When the CSV duplicates filter is active, source rows are now fetched back from the full facet-filtered set before passing to `buildDuplicateReviewRows`, so the non-selectable context row reappears correctly.

PR #1080, opened 2026-06-08
feat(spending): sync budget card month with spending period selector

Jonjon-prog:
Hey @afadil  😄 

A little add up on the dashboard spending. 

When browsing the spending dashboard, it was a bit frustrating to select "Last Month" (or a specific month via the calendar picker) in the spending chart, only to find the Monthly Budget card still showing the current month — requiring a second click to align them. 
This PR wires the two together so switching the spending period automatically updates the budget card.

## What changed

- **Spending → Budget sync**: selecting `Last Month` or a custom month in the spending period selector now automatically moves the budget card to the same month. Multi-month periods (`MTD`, `3M`, `6M`, etc.) reset the budget card to the current month.
- **Persist across navigation**: previously, custom month selections were stored only in the URL and lost when navigating to another page. Both the interval and the custom month are now persisted to `localStorage`, so the full selection is restored when returning to the spending tab.
- **Budget card initializes from selection**: on mount, the budget card now derives its initial month from the persisted spending selection instead of always defaulting to the current month. This fixes a mismatch where the spending chart would show "Last Month" but the budget card would show the current month after a page navigation.

## Behavior summary

| Spending selection | Budget card on select | Budget card on return |
|---|---|---|
| `Last Month` | Jumps to last month | Restored to last month |
| Custom month (picker) | Jumps to selected month | Restored to selected month |
| `MTD`, `3M`, `6M`, `YTD`, `1Y` | Resets to current month | Resets to current month |

The sync is intentionally **unidirectional** (spending → budget). The budget card's own prev/next navigation remains independent — it wouldn't make sense to change the spending chart range when browsing month-by-month in the budget card.

---

## Checklist

EOF

### DIFF
--- a/apps/frontend/src/features/spending/components/spending-tab-content.tsx
+++ b/apps/frontend/src/features/spending/components/spending-tab-content.tsx
@@ -93,6 +93,7 @@ const VISIBLE_SPENDING_DASHBOARD_PERIODS: SpendingDashboardPeriod[] = [
 
 const DEFAULT_INTERVAL: SpendingDashboardPeriod = "MTD";
 const INTERVAL_STORAGE_KEY = "spending-interval";
+const MONTH_STORAGE_KEY = "spending-month";
 const MONTH_PARAM = "spendingMonth";
 const INTERVAL_DESCRIPTIONS: Record<SpendingDashboardPeriod, string> = {
   MTD: "this month",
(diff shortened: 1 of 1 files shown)


afadil:
Thanks @Jonjon-prog

Issue #1077, opened 2026-06-08
Local inference: support for llama.cpp


erikvanzijst:
Is it possible to connect Wealthfolio to a self-hosted llama.cpp instance, or is only Ollama supported? I do see that the OpenAI provider has the ability to override the endpoint URL, but when I put it my llama.cpp /v1 endpoint along with my local api key, the model refresh button fails with `Provider error: Provider returned error 404 Not Found: `

Manually putting in the model name doesn't work either and it only lists OpenAI's own models.

When I put in the base-base-url (without the `/v1` path segment), Wealthfolio is able to properly detect the models available in my llama.cpp backend, but when I then start a chat with the assistant, it fails with:

```
Provider error

CompletionError: ProviderError: Invalid status code 404 Not Found with message:

Check your [AI Providers settings](https://wealthfolio.xxxx.xx/settings/ai-providers) if this persists.
```

afadil:
Is your server OpenAI API compatible? Are you using llama-server or something else around llama.cpp?

erikvanzijst:
Yes, I'm running [llama-swap](https://github.com/mostlygeek/llama-swap) as the entrypoint.

It serves a good number of diverse applications and workloads and have so far not seen any incompatibilities.

erikvanzijst:
Looking the access log I can see that when I put in just my base URL (without the `/v1` path segment), Wealthfolio's admin section accepts it and can retrieve and list models, but then the assistant appears to forget to append `/v1` at runtime and makes requests to `/chat/completions` instead of `/v1/chat/completions`:

```text
[INFO] Request 45.86.93.16 "POST /chat/completions HTTP/1.1" 404 0 "" 7.04µs
```

erikvanzijst:
I've created a quick PR for it (https://github.com/wealthfolio/wealthfolio/pull/1085) and that fixes it for me.

Note: I have not contributed to Wealthfolio before, nor do I have much experience with Rust and had AI help out.

afadil:
fixed in 3.5.2

erikvanzijst:
That was fast!

PR #1085, opened 2026-06-09
fix(ai): normalize OpenAI-compatible base URL for both model listing …

erikvanzijst:
## Description

Two code paths handled `custom_url` differently, making any configured URL work for one operation but not the other:

- `list_models` treated the URL as a bare host and appended `/v1/models`
- `create_openai_client` passed the URL directly to rig, which appends `/chat/completions` — expecting the URL to already include `/v1`

Result: `http://host` broke chat (missing /v1); `http://host/v1` broke model listing (double /v1/v1/models). Fixes issue #1077.

Fix: strip any trailing `/v1` before appending `/v1/models` in list_models, and ensure the URL ends with `/v1` before handing it to rig's OpenAI client. Both `http://host` and `http://host/v1` now work for all operations.

This addresses https://github.com/wealthfolio/wealthfolio/issues/1077

## Checklist

- [X] I have read and agree to the


afadil:
@erikvanzijst Thanks for this fix 🙏

PR #1102, opened 2026-06-10
feat(spending): show savings categories on expense transactions

Jonjon-prog:
Hello @afadil, 

I was digging in the spending categories and I remarked something weird to me. 
Currently, savings categories only appear in the category picker when a transaction's `cashFlowBucket` is `'saving',which only happens for `TRANSFER_OUT` from a cash account to an investment account. 
Regular withdrawals (e.g. a life insurance payment, an RRSP contribution by cheque, a savings plan debit) are classified as `'spending'` and therefore can never be tagged with a savings category.

## What changed

One condition in `QuickCategorizePopover`:

```diff
- if (scope === "saving") {
+ if (scope === "saving" || scope === "expense" || scope === "both") {
```

Savings categories now appear alongside expense categories for any outgoing cash transaction. The `cashFlowBucket` is unchanged — categories label the bucket, they do not change it.

## Why this matters

Many real-world cash outflows are savings in nature even though they leave the account as a withdrawal: life insurance premiums, contributions to a savings plan, transfers to a joint savings account without a linked investment account. Without this change users have no way to categorise these correctly.

## Open question for review

The original design restricted savings categories to transfer-type transactions only. This PR relaxes that. Is there a reporting or analytics reason to keep savings categories exclusive to `cashFlowBucket = 'saving'`? If so, a middle ground could be a dedicated `'expense_or_saving'` scope rather than broadening `'expense'`.

## Test plan

- [ ] Open spending tracker, find a regular withdrawal transaction
- [ ] Click + Categorize — confirm Savings group is now visible alongside Expense
- [ ] Assign a savings category (e.g. Life Insurance) — confirm it saves correctly
- [ ] Confirm saving-type transactions (TRANSFER_OUT to investment) still show savings categories
- [ ] Confirm income transactions still show only income categories

### DIFF
--- a/apps/frontend/src/features/spending/components/quick-categorize-popover.tsx
+++ b/apps/frontend/src/features/spending/components/quick-categorize-popover.tsx
@@ -75,7 +75,7 @@ export function QuickCategorizePopover({
     if (scope === "expense" || scope === "both") {
       out.push(...flattenTaxonomy(SPENDING_TAXONOMY, spending.data?.categories ?? [], "Expense"));
     }
-    if (scope === "saving") {
+    if (scope === "saving" || scope === "expense" || scope === "both") {
       out.push(...flattenTaxonomy(SAVINGS_TAXONOMY, savings.data?.categories ?? [], "Savings"));
     }
     if (scope === "income" || scope === "both") {


Jonjon-prog:
Closing this PR — after further analysis, the fix is too broad and not aligned with the intended design.

## Why this approach is wrong

The original design is semantically correct: savings categories should only appear for transactions whose `cashFlowBucket` is `'saving'`. Showing savings categories on all expense transactions conflates two distinct concepts:

- **cashFlowBucket** = what actually happened to the money (did it leave your net worth, or move between accounts?)
- **category** = how you label/classify that movement

Mixing them breaks analytics: if a withdrawal tagged with a savings category still counts as `spending` in the bucket, totals become ambiguous.

## The real problem

The system currently classifies `cashFlowBucket = 'saving'` **only** for `TRANSFER_OUT` from a cash account to an investment account. This works perfectly when both sides of the transaction are tracked in Wealthfolio.

But it breaks for a common real-world scenario:

> **Account A** — cash/chequing account (full transaction mode)
> **Account B** — life insurance policy, pension plan, employer savings plan (holdings-only mode — user tracks PRU/market value, not individual transactions)

Since Account B doesn't have transactions, the user can't create a proper `TRANSFER_OUT → TRANSFER_IN` pair. The withdrawal from A shows up as a plain spending transaction with no way to mark it as savings.

Other examples of the same pattern:
- Monthly debit for a group RRSP managed by an employer
- Automatic contribution to a savings plan at a third-party institution not connected to Wealthfolio
- Life insurance premium where the cash value accrues in a policy the user tracks as a single holding

## What should be built instead

A way for the user to **manually override `cashFlowBucket`** on a transaction — changing a withdrawal from `spending` → `saving`. This would:

1. Keep the scope logic intact (savings categories naturally appear once bucket = saving)
2. Keep analytics correct (the transaction counts as savings outflow, not spending)
3. Surface correctly in the savings section of the spending tracker

This could be as simple as a "Reclassify as savings" option in the transaction row menu, or a bucket selector in the cash activity form when the transaction type is a withdrawal with no linked transfer.

Happy to open a proper issue if that would be useful.

PR #1126, opened 2026-06-17
fix: activity filter reset button not clearing

eldossjogy:
### Issue (desktop: activity filter reset button)
On desktop, the reset button on the activity filter tab is not clearing all filters; instead, it only clears the date filter.

Before:

https://github.com/user-attachments/assets/adfb38a7-9783-41c6-8502-818882579335

## Fix
- Switched to use ```clearInvestmentsFilters```  useCallback function used by all other reset buttons

After:

https://github.com/user-attachments/assets/9315138c-2f4f-47a7-b98f-69bc2c564620

## Checklist


afadil:
fixed in #1131

PR #1070, opened 2026-06-07
docs(allocations): refresh algorithm notes + add roadmap, research & open questions

Jonjon-prog:
Hey @afadil 👋

Now that the cash↔investment transfer work is merged, I went back over the whole allocation/rebalance feature to take stock and plan what's next. **This PR is docs-only — no code changes.** I'll sign the CLA as usual.

The goal is a shared map so you, me, and any future contributor agree on *where we are* and *where we're going* before the next code PR. I'd love your read on the open questions near the end.

---

## Why this PR

Two docs had drifted from reality:
- `rebalance-algorithm.md` predated the M3 sell/hybrid work and the proportional top-up, so it described an older planner.
- `v1-spec.md` is **your** original scope doc — I've **kept it frozen** as the historical reference. Instead of editing it, I added a *living companion* (`current-state-and-roadmap.md`) for everything that happened since.

## 1. `rebalance-algorithm.md` — brought up to date
- Status now reflects **M3 + proportional top-up**.
- New **§6.5** documents the proportional top-up (deploy residual cash ∝ target weight once the greedy can't improve drift further).
- Fixed the stale §9/§10 notes and corrected the test count.

## 2. New doc: `current-state-and-roadmap.md`

**Where we are.** M1–M3 are delivered (targets/drift, cash-flow planner, exposure-aware drift planner, sell + hybrid, top-up). **M4** (tax-lot trait boundary) is the only unshipped v1 milestone.

**What we deliberately changed vs. the spec** (so the two don't silently disagree):
- **Drafts removed** — your call in #1036; the planner is stateless. If we ever want history, the right home is `RebalanceRun` (SOTA §5.8), not `rebalance_drafts`.
- `base_currency` per target deferred (app-level only for now).
- Drift band is currently absolute — see the proposal in §4 below.

## 3. Research grounding (and a cross-check against official sources)

Every recommendation is backed by research, and I deliberately cross-checked the commercial sources against **primary institutional/government** ones:
- **Our 3 scenario modes** (cash-flow / sell / hybrid) are exactly the three rebalancing methods **the SEC / Investor.gov** teaches investors — not just robo marketing.
- **Threshold > calendar**: backed by **Vanguard (Dec 2024)** and **J.P. Morgan AM (Jan 2025)**.
- **Tax wrapper archetypes** validated against **IRS, HMRC (gov.uk), CRA (canada.ca), ATO, impots.gouv.fr** — including exact holding-period bounds (FR assurance-vie 8y, PEA 5y).

[...]

PR #1153, opened 2026-06-23
feat(allocations): hybrid drift bands with per-sleeve tolerance

Jonjon-prog:
Hey @afadil 👋

Following up on the agreed direction from PR #1070 — this implements hybrid drift bands as discussed.

## What this does

Absolute bands apply the same fixed tolerance to every sleeve, which under-corrects small sleeves and over-corrects large ones (a 5% target sleeve under a ±5% band can double to 10% without triggering). Hybrid bands scale tolerance to each sleeve's target weight with an absolute floor.

**Per-sleeve effective band:**
```
band = max(target_bps × relative_factor, absolute_floor)
```

**Defaults (from our discussion):** 20% relative factor + 100 bps (1%) absolute floor. Existing targets keep absolute behavior unchanged — no silent migration.

### Example with hybrid 20% / 1% floor

| Sleeve target | Relative term | Floor | Effective band | In-band range |
|---|---|---|---|---|
| 50% | ±10% | 1% | ±10% | 40–60% |
| 10% | ±2% | 1% | ±2% | 8–12% |
| 1% | ±0.2% | 1% | ±1% | 0–2% |

## Changes

**Backend (Rust):**
- Migration: `band_type` (absolute/hybrid) + `relative_factor_bps` on `allocation_targets`
- `BandType` enum with `effective_band_bps()` — central computation, used everywhere
- DriftService: per-sleeve band for status (InBand/Overweight/Underweight) + `effective_band_bps` on DriftRow
- Optimizer: refactored from a single global band to per-category band through `RebalanceProfile`
- Validation: `relative_factor_bps` bounded 0–10000
- 13 new tests covering BandType unit, drift service hybrid scenarios, validation

**Frontend (React):**
- `BandType` + `relativeFactorBps` + `effectiveBandBps` in TS types
- New "Drift tolerance" section (step 3) with hybrid/absolute toggle using `AnimatedToggleGroup`
- Smart defaults on toggle: absolute→hybrid resets to 20%+1%, hybrid→absolute resets to 5%
- Separate from rebalance settings (drift tolerance is about monitoring, not rebalancing)

## Test plan

- [ ] Existing targets load as `absolute` with unchanged band values
- [ ] New targets default to `hybrid` (20% relative, 1% floor)
- [ ] Toggle between absolute↔hybrid applies smart defaults
- [ ] Save hybrid target, reload → values persist
- [ ] Drift status changes correctly between modes (small sleeve more sensitive in hybrid)
- [ ] Rebalance plan output differs between absolute and hybrid for same portfolio
- [ ] 84 Rust tests pass (`cargo test -p wealthfolio-core -- allocation_targets`)

## Checklist

### DIFF
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -2400,6 +2400,8 @@ export type ScenarioMode = "cash_flow_only" | "sell_to_rebalance" | "hybrid";
 export type DriftStatus = "in_band" | "underweight" | "overweight" | "not_targeted";
 export type RebalanceTo = "nearest_band" | "exact_target";
 
+export type BandType = "absolute" | "hybrid";
+
 export interface AllocationTarget {
   id: string;
   name: string;
@@ -2408,6 +2410,8 @@ export interface AllocationTarget {
   taxonomyId: string;
   triggerType: TriggerType;
   driftBandBps: number;
+  bandType: BandType;
+  relativeFactorBps: number;
   rebalanceGoal: RebalanceGoal;
   minTradeAmount: string;
   wholeSharesOnly: boolean;
@@ -2424,6 +2428,8 @@ export interface NewAllocationTarget {
   taxonomyId: string;
   triggerType: TriggerType;
   driftBandBps: number;
+  bandType?: BandType;
+  relativeFactorBps?: number;
   rebalanceGoal?: RebalanceGoal;
   minTradeAmount?: string;
   wholeSharesOnly?: boolean;
@@ -2464,6 +2470,7 @@ export interface DriftRow {
   currentValue: number;
   targetValue: number;
   valueDelta: number;
+  effectiveBandBps: number;
   status: DriftStatus;
   isRequired: boolean;
   isZeroCurrent: boolean;
--- a/apps/frontend/src/pages/allocation-targets/components/allocation-donut.test.tsx
+++ b/apps/frontend/src/pages/allocation-targets/components/allocation-donut.test.tsx
@@ -19,6 +19,7 @@ function row(
     currentValue: 0,
     targetValue: 0,
     valueDelta: 0,
+    effectiveBandBps: 0,
     status: "in_band",
     isRequired: true,
     isZeroCurrent: false,
--- a/apps/frontend/src/pages/allocation-targets/components/drift-band-slider.tsx
+++ b/apps/frontend/src/pages/allocation-targets/components/drift-band-slider.tsx
@@ -1,59 +1,125 @@
+import { AnimatedToggleGroup } from "@wealthfolio/ui";
+
+import type { BandType } from "@/lib/types";
+
 interface DriftBandSliderProps {
-  value: number;
-  onChange: (value: number) => void;
+  driftBandPct: number;
+  onDriftBandChange: (value: number) => void;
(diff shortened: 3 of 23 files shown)


afadil:
@Jonjon-prog Thank you for this PR.
few changes to do before merge:

1. The UI still treats `driftBandBps` as the full tolerance in some places, but under hybrid it is only the absolute floor. The #1070 roadmap explicitly says drift rows should show each sleeve’s own band. `TargetRailsCard` and `RebalanceTab` should use `effectiveBandBps` where they display/scale tolerance, or avoid showing a single “tolerance ±1pp” label for hybrid targets.

2. `maxDriftBps` in the rebalance plan remains raw absolute drift. That is fine if it is explicitly “raw max drift,” but it should not be used as if it means “outside tolerance” for hybrid bands. Consider adding a separate band-aware metric like `maxBandExcessBps` / `maxOutOfBandBps`, or keep raw max drift but update labels/visuals so users are not told they are outside range when the effective hybrid band says they are in range.

3. Please make the new-target default policy explicit. The editor creates hybrid targets by sending `bandType: "hybrid"`, but the service defaults omitted `band_type` to `absolute`. That may be intentional for API compatibility, but #1070 says new targets default to hybrid while existing targets preserve absolute behavior. If the backend default stays absolute, add a comment/test explaining that API clients must opt in and only the UI defaults to hybrid. Otherwise centralize the new-target default in core while keeping the DB migration default absolute for existing rows.

4. There is also an inconsistent defensive fallback in `TargetEditor`: one reset path treats a missing `target.bandType` as `"hybrid"`, while `handleCancel()` treats it as `"absolute"`. Real migrated DB rows should always include `bandType`, so this is latent, but for existing targets the safe fallback should be `"absolute"` everywhere to preserve the “no silent migration” guarantee.

5. Please add at least one planner-level regression test for hybrid bands. The current tests cover the band math and drift status well, but the highest-risk behavior is the rebalance planner using different effective bands per sleeve. A good test would show the same portfolio/target producing different sell/buy behavior under `absolute` vs `hybrid`, especially for a large sleeve where hybrid should allow a wider nearest-band range.

Jonjon-prog:
@afadil Separate UX question — I'd love your input on this.

**Drift notation: "pp" vs "%"**

I currently show drift as "+5.2 pp" (percentage points). A few friends tested the app and didn't immediately understand what "pp" meant — they expected to see "%".

Looking at what other platforms do:
- **Retail/prosumer** (Wealthfront, Betterment, M1 Finance, Vanguard, IBKR): all use "%"
- **Institutional** (Bloomberg, Aladdin): use "bps" or "pp"

Since Wealthfolio targets retail/prosumer users, "%" might be more intuitive. The technical ambiguity (5 percentage points vs 5% relative) doesn't really apply in allocation drift context — users understand "+5%" means 5 points of gap.

Would you be open to switching from "pp" to "%" for drift values across the UI? Happy to make the change in this PR or a follow-up — just want to align first.


___

I made all the fixes we talked before :)

afadil:
@Jonjon-prog thanks for the fixes. Yeah, I think it's better to switch from pp to %, as it's more aligned with the retail target.

Jonjon-prog:
@afadil 
Normally it's done for the switch from "pp" to "%" :)

PR #1160, opened 2026-06-23
feat(accounts): asset class override for cash account allocation

Jonjon-prog:
Hey @afadil 👋

Following up on the discussion in PR #1070 (§8) and [discussion #1043](https://github.com/wealthfolio/wealthfolio/discussions/1043) — this implements the minimal `asset_class_override` for CASH accounts.

## The problem

A Livret A (or HYSA, Cash ISA, Tagesgeld, CD…) is tracked as a CASH account and shows up as "Cash" in allocation charts, even though it's economically fixed income. The root cause is `cash_category_id()` in `allocation_service.rs` hardcoding `CASH_BANK_DEPOSITS` for all CASH accounts with no override mechanism.

## What this does

One optional `asset_class_override` field on accounts. When set, the allocation service uses it instead of the default `CASH_BANK_DEPOSITS`. The user picks "this CASH account counts as Fixed Income in my allocation."

- Zero tax logic, zero auto-accrual
- Works for every country (Livret A, ISA, HYSA, CD, Tagesgeld, …)
- Only affects allocation categorization — performance metrics, holdings, valuations are untouched

## Changes

**Backend:**
- Migration: nullable `asset_class_override` column on `accounts`
- `AllocationService`: loads overrides via `AccountServiceTrait` (builder pattern, optional dependency), applies per-account category resolution in `contribution_shares_for_holding`
- Handles merged multi-account holdings via `source_account_ids`
- Wired into both Tauri and server entry points

**Frontend:**
- "Allocation category" selector in the account edit form, shown only for CASH accounts
- Dynamic options loaded from `asset_classes` taxonomy (all top-level categories except Cash)
- Persists correctly on edit (shows current value when reopening)

## Known limitation — breakdown sub-lines

When a CASH account is overridden to e.g. Fixed Income, the amount contributes correctly to the Fixed Income total in the allocation breakdown. However, there's no per-account sub-line under Fixed Income — the user sees the total but can't trace which account contributes what without clicking into the drill-down.

The root cause is that the allocation breakdown tree builds children from taxonomy sub-categories (e.g., Sovereign Bonds under Fixed Income), but overridden cash doesn't belong to any sub-category — it maps directly to the top-level.

A few options for the future:
1. **Per-account sub-lines** — extend the allocation model to create synthetic child entries per account for overridden cash (more complex, touches the `CategoryAllocation` model)
2. **Keep as-is** — the drill-down (click on Fixed Income) already shows individual holdings including overridden cash accounts
3. **Tooltip or annotation** — show a note on the category row when it includes overridden cash

For now I went with option 2 (simplest). Curious if you have a preference for the future.

## Test plan

- [ ] Existing accounts unaffected (no override set)
- [ ] CASH account form shows "Allocation category" selector
- [ ] Setting override to Fixed Income → allocation charts show cash under Fixed Income
- [ ] Override persists on edit (reopening form shows current value)
- [ ] Non-CASH accounts don't show the selector
- [ ] Allocation drill-down shows overridden cash holdings
- [ ] Drift and rebalance correctly treat overridden cash as the override category
- [ ] Other taxonomy views (Sectors, Regions) unaffected
- [ ] 86 allocation tests pass, full core test suite passes

---

## Checklist

### DIFF
--- a/apps/frontend/src/lib/schemas.ts
+++ b/apps/frontend/src/lib/schemas.ts
@@ -102,7 +102,6 @@ export const newAccountSchema = z
     currency: z.string({ required_error: "Please select a currency." }),
     trackingMode: trackingModeSchema.optional().default("NOT_SET"),
     meta: z.string().nullable().optional(),
-    assetClassOverride: z.string().nullable().optional(),
   })
   .refine(
     (data) => data.accountType !== AccountType.CREDIT_CARD || data.trackingMode !== "HOLDINGS",
--- a/apps/frontend/src/lib/schemas.ts
+++ b/apps/frontend/src/lib/schemas.ts
@@ -102,6 +102,7 @@ export const newAccountSchema = z
     currency: z.string({ required_error: "Please select a currency." }),
     trackingMode: trackingModeSchema.optional().default("NOT_SET"),
     meta: z.string().nullable().optional(),
+    assetClassOverride: z.string().nullable().optional(),
   })
   .refine(
     (data) => data.accountType !== AccountType.CREDIT_CARD || data.trackingMode !== "HOLDINGS",
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -91,6 +91,7 @@ export interface Account {
   meta?: string; // Optional - additional metadata as JSON string
   provider?: string; // Optional - sync provider (e.g., 'SNAPTRADE', 'PLAID', 'MANUAL')
   providerAccountId?: string; // Optional - account ID in the provider's system
+  assetClassOverride?: string; // Optional - override allocation category for cash
 }
 
 /**
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -91,7 +91,6 @@ export interface Account {
   meta?: string; // Optional - additional metadata as JSON string
   provider?: string; // Optional - sync provider (e.g., 'SNAPTRADE', 'PLAID', 'MANUAL')
   providerAccountId?: string; // Optional - account ID in the provider's system
-  assetClassOverride?: string; // Optional - override allocation category for cash
 }
 
 /**
--- a/apps/frontend/src/pages/settings/accounts/components/account-edit-modal.tsx
+++ b/apps/frontend/src/pages/settings/accounts/components/account-edit-modal.tsx
@@ -29,6 +29,7 @@ export function AccountEditModal({ account, open, onClose }: AccountEditModalPro
     isArchived: account?.isArchived ?? false,
     trackingMode: account?.trackingMode,
     meta: account?.meta,
+    assetClassOverride: account?.assetClassOverride ?? null,
   };
 
   return (
--- a/apps/frontend/src/pages/settings/accounts/components/account-edit-modal.tsx
+++ b/apps/frontend/src/pages/settings/accounts/components/account-edit-modal.tsx
@@ -29,7 +29,6 @@ export function AccountEditModal({ account, open, onClose }: AccountEditModalPro
     isArchived: account?.isArchived ?? false,
     trackingMode: account?.trackingMode,
     meta: account?.meta,
-    assetClassOverride: account?.assetClassOverride ?? null,
   };
 
   return (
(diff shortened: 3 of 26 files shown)


afadil:
@Jonjon-prog , 
Thanks for putting this together. I agree with the product problem and the narrow direction: this should stay an allocation-categorization feature, not savings-account product support, tax logic, rate history, or auto-accrual.

For core, I’d keep the model simple:

- `account_type = CASH` stays the operational type.
- Optional allocation treatment says how cash in this account appears in asset allocation.
- Default behavior remains `CASH_BANK_DEPOSITS`.
- Only the `asset_classes` taxonomy should be affected.

I don’t think this setting is important enough yet to add a dedicated `accounts` column. It’s more of an allocation hint/user preference than account identity. I’d prefer storing it in account metadata for now, behind a typed helper so JSON details don’t leak everywhere, e.g.

```json
{
  "allocation": {
    "cashCategoryId": "FIXED_INCOME"
  }
}
```

Then core can expose something like `account.cash_allocation_category_id()` and the allocation service can consume that.

Before merge, I’d like a few implementation fixes:

1. **Apply the override only to `asset_classes`.**
   The PR does not update `instrument_type` data, but the runtime allocation calculation passes the same cash override map into `instrument_type`. That can make a cash holding calculate as `FIXED_INCOME` inside the Instrument Type allocation view. The fix is to gate override usage with `taxonomy_id == "asset_classes"` and leave all other taxonomies on their existing cash defaults.

2. **Web mode needs to round-trip the value.**
   The server create/update models accept the setting, but the server `Account` response does not return it. In web mode, the setting can disappear after refetch/edit and may be cleared on a later save. Add it to the server response model if this remains a first-class field, or make sure metadata is returned/preserved if moved into `meta`.

3. **Aggregated cash needs a defined behavior.**
   Multi-account cash is currently merged by currency. If two USD cash accounts have different allocation treatments, the merged holding only has one total amount plus `source_account_ids`, so applying the first matching override can classify the whole merged balance incorrectly. Fix by classifying cash before merging, merging cash by `(currency, resolved asset-class category)`, or carrying per-source cash amounts into allocation.

4. **Add targeted tests.**
   Please cover: default cash behavior unchanged, cash override maps to Fixed Income in asset classes, instrument type remains Cash/FX, web account response includes/preserves the setting, and mixed-source cash behavior.

5. **Keep naming/scope cash-specific.**
   If we keep a typed field instead of metadata, I’d prefer `"allocation.cashCategoryId` over `asset_class_override`,

Jonjon-prog:
Thanks for the detailed review @afadil! All 5 points are addressed:

**1. Override gated to `asset_classes` only**
Added a `taxonomy_id == "asset_classes"` check in `contribution_shares_for_holding`. Instrument type, regions, sectors etc. are unaffected — confirmed with a dedicated test.

**2. Web round-trip via meta JSON**
Moved from a dedicated column to account metadata, as you suggested:
```json
{"allocation": {"cashCategoryId": "FIXED_INCOME"}}
```
The migration is removed. Core exposes `Account::cash_allocation_category_id()` as a typed helper. Since `meta` is already returned in the server Account response, the value round-trips correctly in web mode.

**3. Aggregated cash**
When overrides exist, the allocation service now loads holdings per-account (bypassing the currency-based cash merge). Each cash holding retains its original `account_id`, so overrides resolve correctly per-account. When no overrides are set, the existing merged path is used — zero behavior change for users who don't use this feature.

**4. Targeted tests**
Added 9 tests:
- `Account::cash_allocation_category_id()` parsing (4 tests: valid, missing, no meta, mixed with accounting settings)
- Override maps to FIXED_INCOME in asset classes
- Instrument type remains CASH_FX even with override
- Default behavior unchanged without override
- Mixed-source accounts fall back to default
- All sources with same override → override applies

**5. Naming**
Field is now `meta.allocation.cashCategoryId`, no dedicated DB column.

---

**One UX question** — in the allocation breakdown view, overridden cash contributes to the top-level total (e.g., Fixed Income shows €9,685) but doesn't appear as a visible sub-line. Only actual taxonomy children like "Sovereign Bonds" show. The drill-down (click on Fixed Income) does show the cash holding correctly.

Would you prefer:
1. Keep as-is — drill-down shows the detail
2. Show a sub-line per overridden account using the account name (e.g., "Livret A Boursorama") so multiple overridden accounts each get their own visible line
3. Something else?

afadil:
@Jonjon-prog , I think it we can keep as is with click to show detail

PR #1168, opened 2026-06-25
fix(valuation): forward-fill per-position quotes to prevent zero-value spikes

Jonjon-prog:
Hey @afadil ! 

This PR fixes a bug where a missing quote for a single asset on a given day would cause that position's market value to be treated as **zero**.
Even when other assets in the same account had valid quotes. 
This created dramatic vertical spikes (sudden drops) in portfolio charts that recovered after 1-2 days when the quote became available again.

This is different from [#1022](https://github.com/afadil/wealthfolio/pull/1022) which handles the case where *all* quotes for an account are missing on a day (full-gap skip at the aggregation level). 
This fix targets the **per-position** level within a single account.

## What changed

| File | Change |
|------|--------|
| `valuation_service.rs` | Maintain a `last_known_quotes` map across the daily valuation loop. When a position has a partial quote gap, forward-fill from the most recent known quote for that asset before passing quotes to the calculator. |
| `valuation_calculator.rs` | Updated the warn message to reflect that forward-fill was attempted (no longer "treated as ZERO" when the service layer already tried to fill). |

## How it works

The quotes service (`fill_missing_quotes`) already does carry-forward at the quote series level, which handles most gaps (weekends, holidays). This fix adds a **second layer of defense** in the valuation loop itself:

1. Each day, store all available quotes in `last_known_quotes`
2. For positions with a partial quote gap, fill from `last_known_quotes`
3. Log forward-filled assets at `debug` level, truly missing assets (no prior quote at all) at `warn` level
4. Pass the filled quote map to the calculator

The full-gap skip logic (all quotable positions missing → skip day) is unchanged and still operates on raw quotes.

## Test plan

- [x] All 1454 existing tests pass (`cargo test -p wealthfolio-core`)
- [ ] Manual: verify portfolio chart with an asset that has sporadic quote gaps no longer shows vertical spikes

---

## Checklist

### DIFF
--- a/crates/core/src/portfolio/valuation/valuation_calculator.rs
+++ b/crates/core/src/portfolio/valuation/valuation_calculator.rs
@@ -328,7 +328,7 @@ fn calculate_investment_market_value_acct(
         } else {
             unpriced_positions += 1;
             warn!(
-                "Missing quote for asset {} on date {}. Position market value treated as ZERO.",
+                "No quote (including forward-fill) for asset {} on date {}. Position market value treated as zero.",
                 asset_id, target_date
             );
         }
--- a/crates/core/src/portfolio/valuation/valuation_service.rs
+++ b/crates/core/src/portfolio/valuation/valuation_service.rs
@@ -1638,6 +1638,7 @@ impl ValuationServiceTrait for ValuationService {
         };
 
         let mut skipped_incomplete_dates: Vec<(NaiveDate, String)> = Vec::new();
+        let mut last_known_quotes: HashMap<String, Quote> = HashMap::new();
         let mut newly_calculated_valuations: Vec<DailyAccountValuation> = snapshots_to_process
             .into_iter()
             .filter_map(|holdings_snapshot| {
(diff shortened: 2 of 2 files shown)


afadil:
@Jonjon-prog , Thanks for the PR. I think this duplicates existing behavior: `calculate_valuation_history` already calls `get_quotes_in_range_filled`, and that path already carries forward the last known quote per asset before valuations are calculated.

Could you add a minimal regression test that fails on `main` and passes with this change? If the issue is still reproducible, the root cause is likely in the existing quote-fill path or in missing prior quote data, not in the valuation loop itself.

Jonjon-prog:
You're right — after digging deeper into the flow, I can confirm this is indeed redundant with the existing quote-fill path.

Here's what I traced:

1. `calculate_valuation_history` collects all non-alternative asset IDs across all snapshots into `required_asset_ids`
2. `get_quotes_in_range_filled` calls `fill_missing_quotes` which carries forward the last known quote per symbol for **every calendar day** in the range
3. `append_historical_seed_quotes` goes further — for any symbol without a quote in the 30-day lookback window, it calls `get_latest_quote_before()` with no time limit, ensuring even stale assets get a seed quote
4. The resulting `quotes_by_date` map is complete for every asset that has at least one quote in the DB

The only remaining gap is assets with **zero quotes in the DB** (e.g. newly imported position, provider doesn't cover it) — but the forward-fill in this PR can't help there either since there's no prior price to carry forward from.

The spikes I originally reported were likely from before the `fill_missing_quotes` improvements (Jan 2026) or `append_historical_seed_quotes` (Mar 2026). Closing this PR — thanks for catching the duplication!

PR #1177, opened 2026-06-26
feat(rebalance): generic sell constraints, turnover cap & plan explanations (Phase 2)

Jonjon-prog:
Hey @afadil 👋

This PR starts Phase 2 work: generic trade constraints + plan explanations, as agreed in #1070. **Work in progress** — the backend is complete and tested, but the frontend constraint UI needs your input on placement before I finalize it.

### 1. Generic sell constraints (Rust + DB + API)

Three constraint types, all working end-to-end:

- **`do_not_sell` per asset** — excludes specific assets from sell candidates. Persisted in new `rebalance_sell_constraints` table (synced via device-sync).
- **`avoid_selling` per account** — excludes all holdings from a specific account. Same table.
- **`max_turnover_pct`** on `AllocationTarget` — caps the % of portfolio value that can be sold in one plan. New column on `allocation_targets`.

The optimizer filters sell candidates before the sell phase and emits `ConstraintSkippedSell` / `TurnoverCapReached` warnings explaining why assets were skipped.

### 2. Enriched plan explanations (Rust)

The `reason` field on each trade now explains the *why*:
- Buy: `"Buy VTI: Equity underweight by 2000 bps."`
- Sell: `"Sell BND: Bond overweight by 4000 bps."`
- Manual: `"Equity is underweight — no classified holding available. Allocate manually."`
- Constraint skip: `"Skipped selling AAPL: do-not-sell constraint."`
- Turnover cap: `"Turnover cap (5.0%) reached — 500.00 of 10000.00 portfolio already sold."`

### 3. Frontend (partial)

- `max_turnover_pct` field in target editor (settings section)
- New warning labels for constraint/turnover warnings
- Constraint summary badges bar
- Full adapter wiring (Tauri + web) for constraint CRUD

### 4. Tests

5 new optimizer tests covering all constraint types + reason strings. 103 total allocation tests pass.

### Open question 1: Where should do-not-sell constraints be configured?

I initially put a lock toggle directly on sell trades in the results table (toggle after calculating). But stepping back, it felt counter-intuitive — the user should decide which assets to protect *before* running the plan, not discover them in the output and retroactively lock them.

That raises the question of where these controls should actually live. 

Looking at the SOTA spec, there are two natural options:

**Option A — Target profile settings (SOTA §5.7 ExecutionPolicy)**
The `blocked_asset_ids` in ExecutionPolicy is profile-level. This would go near `allow_sells` / `max_turnover_pct` in the settings section. Requires an asset picker since the target editor doesn't currently show holdings.

**Option B — Drill-down inside sleeve (SOTA §5.4 HoldingTarget)**
`is_sellable` lives on HoldingTarget, which is per-asset inside a sleeve. This would mean clicking on a sleeve in the target allocation table → seeing the holdings → toggling sell/buy per holding.

Option B also sets the pattern for where future `HoldingTarget` (per-ticker targets, Phase 4-5) would live.

**My lean:** Option B feels more intuitive for users but is more work. Option A is simpler and ships faster. Both are valid per SOTA.

### Open question 2: Do we need `do_not_buy` too?

Currently we only have sell constraints. The SOTA §5.4 has both `is_buyable` and `is_sellable` on HoldingTarget. Should we add `do_not_buy` now while we're building this, or defer?

### Open question 3: How does this relate to HoldingTarget UI?

The do-not-sell UI will set the pattern for where HoldingTargets eventually live. If you have a preference for how per-holding controls should look in the app, that would help us design the constraint UI to be forward-compatible.

## Phase 2 progress

| Item | Status |
|---|---|
| Hybrid drift bands | ✅ Merged |
| Generic sell constraints (backend) | ✅ This PR |
| Turnover cap (backend + UI) | ✅ This PR |
| Plan explanations (backend) | ✅ This PR |
| Constraint UI placement | ⏳ Waiting for your input |
| Addon extension points | 🔜 Next |

## Migration

Single migration adding:
- `max_turnover_pct TEXT DEFAULT NULL` on `allocation_targets`
- `rebalance_sell_constraints` table with device-sync support

## Checklist


afadil:
@Jonjon-prog , I reviewed this against the current allocation/rebalance architecture and I think the feature direction is relevant, but I’d avoid making `rebalance_sell_constraints` the long-term model.

Here is a proposed design document (by Codex :) [rebalance-execution-policy-design.md](https://github.com/user-attachments/files/29388564/rebalance-execution-policy-design.md). Can you have a look and let me know if you see a different, better approach?

Summary:
- keep scalar execution settings on `allocation_targets` (`max_turnover_bps`, min trade, whole shares, allow sells);
- add a generic `allocation_target_constraints` child table instead of a sell-only constraints table;
- store constraints as rows, not JSON, for sync/uniqueness/conflict handling;
- make sell candidates account/holding-aware before applying account-level rules;
- keep v1 constraint actions explicit (`buy` / `sell`), and defer tax-specific behavior.

This keeps Phase 2 small while leaving a clean path for do-not-buy, addon policies, and future tax-aware integration.

Jonjon-prog:
Hello @afadil

Thanks for the design doc. 
The direction is solid and aligns with what I had in mind. Here's my take on it plus a couple of questions.

## On the design doc

I agree with the core decisions:

- **`allocation_target_constraints` as a generic child table** instead of `rebalance_sell_constraints` — the generic shape (subject_type + action + effect) is cleaner and leaves room for do-not-buy, category-level blocks, and addon policies without renaming anything later.
- **`max_turnover_bps` as an integer** on `allocation_targets` — basis points is consistent with how everything else is stored. I had it as `max_turnover_pct TEXT`, will convert.
- **Constraints loaded by the service**, not passed from the frontend — simpler contract, the planner owns its data.
- **No FK to assets/accounts** in v1 — pragmatic for sync and archived records.
- **Row-based sync** over JSON — better conflict isolation.

One question on `effect: block | avoid`: for v1, `avoid` doesn't have a concrete optimizer behavior (it would need a scoring penalty, and the main use case — "prefer selling from taxable" — requires the tax wrapper framework). My plan is to implement `block` only in the optimizer and store `avoid` as a valid DB value for forward-compat. Does that match your intent?

## UI placement — two layout options

I initially had the constraint toggles on sell trades after calculating, but realized it was counter-intuitive, the user should decide what to protect *before* running the plan. 
Your doc confirms this: constraints are policy, not output annotations.

I mocked up two layout options for the target editor. Both use the same component designs (pills with X, searchable popover for holdings/accounts, badge counts).

### Option A — Constraints in the right column

Left column: Name & scope, Tolerance bands, Trade settings (allow sells, share sizing, min trade, max turnover).
Right column: Starting point, Weights, **+ Sell constraints section** (protected assets + protected accounts).

screenshot 2026-06-29 à 10 33 23
</br>

### Option B — All execution settings below weights

Left column: Name & scope, Tolerance bands only.
Right column: Starting point, Weights, **Execution card** (trade settings + protected assets + protected accounts in a 2-column grid).

screenshot 2026-06-29 à 10 33 35
</br>

**My lean: Option A.** It keeps better visual balance between columns, requires fewer layout changes to the existing code, and leaves room for future HoldingTargets in the same right-column area (drill-down from weights). Option B makes the left column too empty.

Also open on the section naming — "Protected" might not be the clearest label. "Sell constraints", "Execution constraints", "Do-not-sell".
Happy to align on whatever wording fits the app's voice.

## On the open questions

**Do-not-buy:** the generic table supports it already (`action: 'buy'`). Should I wire it in the optimizer now (filter buy candidates the same way), or ship sell constraints first and add buy in a follow-up?

**HoldingTarget UI:** wherever I put the constraint lists will likely set the pattern for where per-holding targets live later. If you have a preference on that, it would help me be forward-compatible.

## What's already done (backend)

The optimizer logic, enriched reasons, turnover cap, warnings, CRUD, sync wiring, and tests are all in place. 
The refactor to match your design doc (rename table, switch to bps, add action/effect enums, load constraints server-side) is mostly renaming.
Tthe planner architecture already follows the pipeline you described.

Ready to start the refactor + UI implementation once you confirm the layout direction.

afadil:
@Jonjon-prog thanks, this lines up with the design direction.

Let's go with Option A for the UI: keep the scalar rebalance settings in the left column, and add a right-column section under weights for the sell policy. I’d call the section **Sell protection** rather than “constraints” in the UI, with **Protected assets** and **Protected accounts** inside it.

For v1, please implement/expose only `action = sell` + `effect = block`. I’m fine with keeping `avoid` as a reserved enum/table value for forward compatibility, but it should not be selectable or created by the current UI/API until we have real optimizer semantics for it. Same for `do_not_buy`: keep the generic model ready for it, but defer buy filtering/UI to a follow-up.

A couple of implementation details I’d like to keep crisp:

- Use `allocation_target_constraints` and `max_turnover_bps` as in the design doc.
- Load persisted constraints in the service/planner path; don’t pass constraint arrays from the frontend into plan calculation.
- For account-level sell protection, sell candidates need to be account/holding-aware before filtering. If the same asset exists in multiple accounts, blocking one account should only remove that account’s sell candidate.
- For sync, prefer row-level add/remove or stable constraint identity/natural uniqueness, so duplicate toggles across devices converge cleanly.

For HoldingTarget UI, let’s not solve that in this PR. Asset classification already covers which sleeve/category an asset belongs to today. The v1 **Sell protection** section should stay focused on one job: assets/accounts that should not be sold.

If we later add per-asset target percentages, we can design that as a separate flow from the weights UI.

Jonjon-prog:
@afadil Implemented per the design doc — here's what's in place now:

**Backend (renamed/refactored per your feedback):**
- `allocation_target_constraints` table with `subject_type` (asset/account/category), `subject_id`, `action` (buy/sell/trade), `effect` (block/avoid), `reason`, `metadata_json`
- `max_turnover_bps` INTEGER on `allocation_targets`
- Constraints loaded in `RebalanceService` from DB, no longer passed from the frontend
- `SellCandidate` is account/holding-aware — account-level blocks only remove that account's candidate, not the asset everywhere
- Row-level sync with `UNIQUE(target_id, subject_type, subject_id, action, effect)` for natural convergence

**V1 scope, as agreed:**
- Only `action: sell` + `effect: block` are creatable from the UI/API surface I built. `avoid` and `buy` are valid enum values in the schema but nothing in the app produces them yet.

**UI — Option A, built and tested locally:**
- "Sell protection" section in the right column under weights, only visible when sell mode is enabled
- "Protected assets" / "Protected accounts" with pill badges + searchable popover pickers
- Constraints are buffered in local state and only persisted when the user clicks "Save target" — matches how the rest of the target form behaves (weights, bands, trade settings), so Cancel reverts everything including protection changes

103 tests passing, manually verified the full flow: add protection → calculate Sell/Hybrid plan → protected holdings/accounts excluded with a `ConstraintSkippedSell` warning → remove protection → re-calculate → included again.

Ready for review whenever you have time.

afadil:
@Jonjon-prog  Thanks for the refactor.  found a few issues that should be fixed before merge.

## Findings

1. **Web mode cannot load/save sell protection constraints**

The web adapter calls:

- `GET /allocation-targets/{id}/constraints`
- `POST /allocation-targets/{id}/constraints`

from `apps/frontend/src/adapters/web/core.ts`.

But the Axum server registers:

- `/allocation-targets/{id}/sell-constraints`

in `apps/server/src/api/allocation_targets.rs`.

So the new constraint UI should work in Tauri, but web mode will 404 for constraint load/save. Please align the adapter and server route. Given the generic table/model, I’d lean toward renaming the server route to `/constraints`.

2. **Account-level sell protection is not actually account-aware**

The design doc called out that sell candidates need account/holding granularity before filtering. The current implementation still builds sell candidates by `asset_id` in `RebalanceService::build_sell_candidates`, then stores only one representative `account_id`.

That breaks account-level protection in all/portfolio scopes because holdings are aggregated earlier. `get_holdings_for_accounts` merges holdings by asset and stores real accounts in `source_account_ids`, while `account_id` becomes the aggregate scope id. As a result, protecting account `A` may not match anything, or the planner may be unable to remove only account `A`’s sellable shares when the same asset exists in multiple accounts.

This should be changed so sell candidates are built at holding/source-account granularity, or the aggregation layer needs to preserve enough per-account quantity/value information for sell filtering.

3. **Max turnover cannot be cleared once set**

The UI sends `maxTurnoverBps: null` when the input is emptied, but the service update path uses:

```rust
input.max_turnover_bps.or(existing.max_turnover_bps)
```

and the save-with-weights helper has the same preserve-existing behavior. That means a user can clear the field, save successfully, and still keep the old turnover cap.

This needs a way to distinguish “field omitted” from “explicitly set to null,” probably with an update DTO that uses `Option<Option<i32>>` or another explicit clear path.

4. **`0%` turnover is accepted but treated as no cap**

The migration allows `max_turnover_bps = 0`, and the UI input allows `min=0`. But planning filters zero out before passing it to the optimizer, so `0%` behaves like unlimited turnover.

Either disallow `0` in validation/UI, or treat `0` as a valid “sell nothing” cap. The current behavior is surprising and contradicts the DB constraint shape.

Jonjon-prog:
Thanks for the thorough review. All four points addressed in commit `6d9d3a427`:

**1. Route mismatch**
Renamed the Axum route from `/sell-constraints` to `/constraints` to match the web adapter. The web adapter URL was already correct — the server was the odd one out.

**2. Account-level sell protection**
Added `source_account_ids: Vec<String>` to `SellCandidate`. `build_sell_candidates` now collects all real account IDs from `HoldingAllocationContribution.source_account_ids` (or falls back to `account_id` for non-aggregated holdings). The constraint filter checks both the primary `account_id` and all `source_account_ids`, so blocking account A correctly suppresses sells across all holdings where A is a source — even when the same asset is held in multiple accounts.

**3. Max turnover clearable**
Removed the `.or(existing.max_turnover_bps)` fallback in both `update_target` and `build_target_from_input`. Since the frontend always sends the complete DTO on save, `None` from the payload now correctly clears the cap. The create path already had the right behavior (`input.max_turnover_bps` directly).

**4. Zero bps treated as "sell nothing"**
Removed the `bps > 0` filter in `rebalance_service` and the `> Decimal::ZERO` filter in the optimizer. `max_turnover_bps = 0` now produces `turnover_cap_value = 0`, which immediately triggers `TurnoverCapReached` and blocks all sells. The UI `min` attribute stays at `0` since this is a valid explicit choice (rather than silently disabling the cap). The DB check constraint already permits `0`.

PR #1202, opened 2026-06-30
Fix standalone tax charge handling

afadil:
## Summary

- Make standalone `TAX` activities read `tax` as the primary charge amount, with existing fallbacks to `fee` and `amount` preserved.
- Apply the same type-aware charge resolution to performance attribution so tax-only activities no longer persist without attribution impact.
- Add regression coverage for cash balance impact and attribution components.

## Root Cause

PR #1198 added a `tax` field to create/API surfaces, but standalone `TAX` charge handling still resolved charges from `fee` or `amount` only. A tax-only activity could therefore persist `tax` while producing no cash or performance tax effect.

## Validation

- `cargo fmt`
- `cargo check -p wealthfolio-core`
- `cargo clippy -p wealthfolio-core --all-targets -- -D warnings`
- `cargo test -p wealthfolio-core test_tax_activity_uses_tax_field_for_charge -- --nocapture`
- `cargo test -p wealthfolio-core activity_attribution_components_separate_income_fees_and_taxes -- --nocapture`
- `cargo test -p wealthfolio-core test_charge_activities_fee_and_tax -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/core/src/activities/activities_model.rs
+++ b/crates/core/src/activities/activities_model.rs
@@ -222,6 +222,19 @@ impl Activity {
         self.tax.unwrap_or(Decimal::ZERO).abs()
     }
 
+    /// Get the charge amount for standalone charge handling.
+    pub(crate) fn charge_amt_for(&self, activity_type: &ActivityType) -> Decimal {
+        if matches!(activity_type, ActivityType::Tax) && !self.tax_amt().is_zero() {
+            return self.tax_amt();
+        }
+
+        if !self.fee_amt().is_zero() {
+            return self.fee_amt();
+        }
+
+        self.amt()
+    }
+
     /// Get typed metadata value
     pub fn get_meta<T: serde::de::DeserializeOwned>(&self, key: &str) -> Option<T> {
         self.metadata
--- a/crates/core/src/portfolio/performance/performance_service.rs
+++ b/crates/core/src/portfolio/performance/performance_service.rs
@@ -2166,7 +2166,7 @@ impl PerformanceService {
             }
             ActivityType::Fee => (
                 Decimal::ZERO,
-                Self::activity_charge_amount(activity),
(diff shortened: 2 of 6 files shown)


afadil:
Updated this PR to address the duplication review note.

Changes:
- Moved the standalone charge precedence into `Activity::charge_amt_for` as the single source of truth.
- Updated both the holdings cash-flow path and performance attribution path to call that shared helper.
- Removed the duplicated private `activity_charge_amount` helpers.

Validation rerun:
- `cargo fmt`
- `cargo clippy -p wealthfolio-core --all-targets -- -D warnings`
- `cargo test -p wealthfolio-core test_tax_activity_uses_tax_field_for_charge -- --nocapture`
- `cargo test -p wealthfolio-core activity_attribution_components_separate_income_fees_and_taxes -- --nocapture`

afadil:
Follow-up on the inline `effective_type()` review thread: the comment was valid.

Update pushed in `6371b1c9f`:
- Snapshot scratch routing now parses `activity.effective_type()`.
- `process_single_activity` dispatch now parses `activity.effective_type()` instead of the stored `activity_type`.
- The trade amount helper now uses the effective type as well, so overridden BUY/SELL rows stay consistent inside handlers.
- Added a regression test for a stored `FEE` activity overridden to `TAX` with only `tax` populated.

Validation rerun:
- `cargo fmt`
- `cargo clippy -p wealthfolio-core --all-targets -- -D warnings`
- `cargo test -p wealthfolio-core test_tax_activity_override_uses_effective_type_for_charge -- --nocapture`
- `cargo test -p wealthfolio-core test_tax_activity_uses_tax_field_for_charge -- --nocapture`
- `cargo test -p wealthfolio-core activity_attribution_components_separate_income_fees_and_taxes -- --nocapture`
- `git diff --check`

PR #1210, opened 2026-07-02
Add Chinese localization support

southbaird:
## Summary
- add an English/Chinese language setting with persisted app settings and html lang updates
- localize key Wealthfolio UI surfaces, including dashboard, insights, performance, accounts, holdings, assets, spending, goals, settings, and shared date/interval controls
- add runtime fallback localization for remaining static UI strings, placeholders, aria labels, and common backend/status messages
- fix the web build target propagation so Vite builds with BUILD_TARGET=web
- keep Tauri-only authentication and scanner paths out of the browser web runtime

## Testing
- git diff --check
- corepack pnpm --filter frontend type-check
- corepack pnpm --filter frontend build

## Notes
- Build completed with existing Vite sourcemap and chunk-size warnings.


afadil:
i18 added in https://github.com/wealthfolio/wealthfolio/pull/1217

PR #1218, opened 2026-07-03
perf(db): cut holdings_snapshots bloat ~8×, tune SQLite, fix TWR headline

atsoc0ocsav:
### Summary
Three independent improvements to the local-first SQLite backend. Happy to split into separate PRs if you'd prefer to review them individually.
  
#### 1. SQLite performance tuning
- Per-connection PRAGMAs on the pooled read connections: `cache_size` (64MB), `mmap_size` (256MB), `temp_store = MEMORY` (previously SQLite defaults —  ~2MB cache, mmap disabled).
- New composite index `idx_activities_account_date` on `activities(account_id, activity_date)` for the main activity-list query (filters `account_id` + `activity_date` range, orders by `activity_date`).
- `ANALYZE` after migrations so the planner uses real statistics.

#### 2. TWR headline no longer nulled by dormant/dust days
When a single day's TWR denominator (`prev_value + inflow`) fell below one base-currency unit, the whole-period Time-Weighted Return was returned as `None` even though the daily chart still rendered — e.g. an account funded, fully sold, left as rounding dust for a stretch, then refunded. Those benign low-base days now pause compounding instead of nulling the headline; genuinely fatal reasons (unknown flow, unavailable coverage, negative value) still null it. Daily chart output is unchanged.

#### 3. holdings_snapshots bloat (~89% of DB size)
Each daily snapshot re-serialized the full per-position `lots` array into the `positions` JSON, duplicating the normalized `lots` table (observed ~343MB of a 420MB DB). Fixed in stages:
- Precompute per-position cost basis (base + account currency, at acquisition-date FX) into new `snapshot_positions` scalar columns, decoupling valuation from the embedded lots.
- Stop serializing `Position.lots`; holdings display now reads current lots from the `lots` table, and incremental-recalc seeds hydrate lots from the `lots` table so market-sync recalcs stay consistent.
- One-time startup migration that auto-migrates existing instances.
  
### Migration / upgrade behavior
- Additive schema migrations add the new columns/index.
- The one-time strip migration is destructive but **creates an automatic backup first and aborts if the backup fails**. It is guarded (runs only when embedded lots are present) and idempotent (a second run is a no-op). It backfills the cost-basis scalars from the existing embedded lots (identical to a write-time recompute), strips the embedded lots, and runs a single `VACUUM`.

### How to test
- `CONNECT_API_URL=http://test.local cargo test --workspace` — includes new regression tests: TWR dormant/dust; cost-basis parity (same- and cross-currency); incremental seed-lot hydration; migration parity/idempotency/backup-abort/local-FX-fallback.
- Upgrade path: run the server against a pre-change DB that still embeds lots; on startup it backs up, backfills scalars, strips lots, and VACUUMs. Verified on a real 420MB DB → **51MB**, with activities byte-identical and `cost_basis_base` / `net_contribution_base` identical across all valuation rows (0 mismatches) — value-for-value equal to a full rebuild.

### Notes
- Web/server (Docker) path validated end-to-end; Tauri desktop compiles (`cargo check --workspace`).
- No frontend changes.
- Existing instances that don't run the migration keep working via a fallback (old rows retain embedded lots + a lot-walk read path); the migration is what reclaims the disk.

## Checklist


afadil:
Thanks for the work here. The optimization is valuable.

I do think a few correctness/design issues should be addressed before merge.

### Main design direction: aggregate-only snapshots + current lots only for append-only updates

After this PR, snapshots should be treated as aggregate-only history. That is fine for display and valuation, but they should no longer be used as lot-book seeds for historical replay.

The current PR tries to preserve historical incremental recalculation by hydrating seed lots from the current `lots` table. I do not think that is valid. The current `lots` table represents the latest remaining open lots, not the lot book as of an arbitrary historical snapshot date.

Example:

```text
Jan 1: buy 100 shares
Feb 1: sell 40 shares
Today: lots table has remaining quantity = 60
```

If a recalculation needs to replay from Jan 15, the correct seed lot quantity is 100, not today’s 60. Hydrating from current open lots would replay future sells against the wrong lot state. It can also inject lots opened after the historical seed date.

Product-direction-wise, I think the simpler and safer model is:

```text
Snapshots = aggregate-only history for valuation/display.
Lots table = current lot book for the latest calculated state.
Strictly append-only new activity = can use current snapshot + current lots table.
Backdated add/edit/delete = rebuild the affected account from inception.
```

“New activity” here has to mean chronologically after the latest calculated state. A newly entered activity dated in the past is a historical mutation and should not use current lots as a historical seed.

That preserves the storage win without adding `lot_events` or re-embedding lots in every snapshot. Historical corrections may be slower, but they stay correct. If that later proves too slow for real portfolios, a separate minimal lot-book checkpoint table would be the next step.

### Migration direction: drop generated snapshots instead of rewriting them

I think we can avoid the complex “strip lots + backfill scalar cost basis” migration for generated snapshots entirely.

`CALCULATED` holdings snapshots and `daily_account_valuation` are derived read models, and existing migrations already reset them when generated schema semantics change. A simpler safer migration would be:

```sql
-- Defensive: preserve legacy HOLDINGS-mode source snapshots if any still say CALCULATED.
UPDATE holdings_snapshots
SET source = 'MANUAL_ENTRY'
WHERE source = 'CALCULATED'
  AND account_id IN (
    SELECT id FROM accounts WHERE tracking_mode = 'HOLDINGS'
  );

-- Generated read models. Rebuild from activities.
DELETE FROM holdings_snapshots
WHERE source = 'CALCULATED';

DELETE FROM daily_account_valuation;
```

If this PR changes lot-table semantics too, then `lots` / `lot_disposals` may also need to be cleared and rebuilt.

This avoids doing FX reconstruction inside a startup DB migration and avoids keeping aggregate-only snapshots with incomplete or half-trusted cost-basis scalars. The tradeoff is a one-time rebuild after upgrade, which seems acceptable for generated read models.

Important caveat: do not delete `MANUAL_ENTRY`, `CSV_IMPORT`, or `BROKER_IMPORTED` snapshots as part of this migration. They are not necessarily reconstructable from activities.

### Other correctness concern: account-currency FX is lost in persisted lots

Even with the simpler product direction above, the current `lots` table still needs to be a correct current lot book.

Right now `LotRecord` preserves `fx_rate_to_base`, but not `fx_rate_to_account` / `account_currency`, and `lot_record_to_snapshot_lot` sets those to `None`.

That means a full rebuild and an append-only incremental rebuild can diverge for multi-currency accounts. If the original activity had an explicit account FX rate, the in-memory lot book preserves it, but the persisted lot book loses it and later falls back to market FX. This can change `cost_basis_account` and downstream snapshot values.

So even if current lots are used only for append-only processing, they should persist enough account-FX context to be equivalent to the in-memory lot book.

### Lower-priority follow-ups

- I’d consider dropping the SQLite PRAGMA tuning from this PR unless there are benchmarks showing it is needed. For a single-user local app, `cache_size = -65536` with an 8-connection pool is aggressive because the cache budget is per connection, and `temp_store = MEMORY` can turn large sorts into memory spikes. Especially for the mobile app. The snapshot bloat reduction is already the main performance win here; I’d keep DB memory tuning separate, benchmarked, and platform-aware.
- `ANALYZE` appears to run even when no migration was applied. It should probably run only when migrations were actually applied.
- The new `activities(account_id, activity_date)` index should be verified with `EXPLAIN QUERY PLAN` against the intended hot queries, especially queries sorted by `activity_date, created_at`.
- The TWR denominator change may classify invalid negative denominator cases as benign. Near-zero positive denominators can be dormant/benign, but negative denominator states should probably remain data issues.

atsoc0ocsav:
Hi,

Definitely valid inputs, thank you for following up!

The main driver for optimization was the fact that the nested snapshots
were pilling up (was testing with an history of ~3k movements over a period
of ~5Y) and rendering was becoming slow, with a quite high RAM usage (1.5
to 2GB in peak) when updating the full history.

Let me take a more detailed look over your feedback, over the next few
days/hours, and re-iterate the PR!

Best regards,
Vasco Craveiro Costa

A sexta, 3/07/2026, 20:47, FADIL ***@***.***> escreveu:

atsoc0ocsav:
Hey @afadil ,

After taking a closer look at you feedback:

**1. Aggregate-only snapshots + current lots for append-only only**
Agreed. Hydrating historical seeds from the current `lots` table may not be the best approach, as the current table holds remaining open lots, not the as-of-date lot book - I'll drop that path and adopt the model you described:
- Snapshots = aggregate-only history (display/valuation).
- `lots` table = current lot book.
- Strictly append-only activity (dated after the latest calculated state) → seed from current snapshot + current lots.
- Backdated add/edit/delete → full rebuild from inception.

One question: is there an existing per-account "latest calculated state" high-water mark I should key the append-only check off, or should I derive it from the latest snapshot/valuation date?
  
**2. Migration: drop generated read models instead of rewriting**
Agreed. I'll remove the strip + scalar-backfill + in-migration FX reconstruction entirely, and replace it with:
- set legacy HOLDINGS-mode `CALCULATED` → `MANUAL_ENTRY`
- `DELETE` `CALCULATED` holdings_snapshots + all `daily_account_valuation`
- rebuild on next calc preserving `MANUAL_ENTRY` / `CSV_IMPORT` / `BROKER_IMPORTED`. Since this PR also changes `lots` semantics (account FX, below), I'll clear + rebuild `lots` / `lot_disposals` as well.

**3. Persist account FX in the lot book**
Good catch, I'll add `fx_rate_to_account` + `account_currency` to the `lots` table and carry them through the `LotRecord` ↔ `Lot` mapping, so the persisted lot book is equivalent to the in-memory one and append-only rebuilds match a full rebuild for multi-currency accounts.

**4. TWR negative denominators**
Did not thought of that, I'll restrict the dormant/benign classification to `0 <= denom < 1` and keep negative denominators as data issues.

**5. ANALYZE gating**
Agreed, will run it only when migrations were actually applied.

**6. PRAGMA tuning**
Agreed on the memory PRAGMAs. `cache_size`, `mmap_size`, and `temp_store` will be dropped from this PR (the per-connection × pool budget and mobile spike concerns are fair, and I don't have benchmarks to justify them). I'd like to keep the `activities(account_id, activity_date)` index, but per your feedback I will also take at `EXPLAIN QUERY PLAN` against the hot queries - including the `activity_date, created_at` sort - and extend to `(account_id, activity_date, created_at)` if the planner wants it. I'll post the plans here. Happy to pull the index into its own PR if you'd prefer this one to be purely the snapshot-bloat change.

atsoc0ocsav:
Hi again @afadil, 

I've pushed a revision addressing all points. Summary of the changes:

**1. Aggregate-only snapshots + current lots only for append-only.** Done.
Removed the historical seed-lot hydration. Recalc now gates at a single core choke point: a recalc seeds from the current snapshot + current `lots` only when it is strictly append-only (earliest affected activity is after the account's latest calculated state / high-water mark). Any backdated add/edit/delete is upgraded to a Full rebuild from inception (fails closed to Full). Snapshots are treated as aggregate-only; the `lots` table is the current lot book.

**2. Drop generated read models instead of rewriting.** Done.
Replaced the strip + scalar-backfill + in-migration FX migration with a plain reset: convert legacy HOLDINGS-mode `CALCULATED` snapshots to `MANUAL_ENTRY`, then `DELETE` `CALCULATED` snapshots, `daily_account_valuation`, `lot_disposals`, and `lots`. `MANUAL_ENTRY`/ `CSV_IMPORT`/ `BROKER_IMPORTED` are preserved. They rebuild on next launch via the existing `portfolio_history_backfill_needed` full recalc. No more FX reconstruction in a migration.

**3. Account-currency FX in persisted lots.** Done.
Added `fx_rate_to_account` + `account_currency` to the `lots` table, populated at persistence and mapped back in `lot_record_to_snapshot_lot`. A full rebuild and an append-only rebuild now match on `cost_basis_account` for multi-currency accounts (verified below).
  
**Lower-priority:**
- Dropped the `cache_size`/`mmap_size`/`temp_store` PRAGMAs.
- `ANALYZE` now runs only when migrations were actually applied.
- TWR: only `0 <= denom < 1` is treated as dormant/benign; a negative denominator stays a fatal data issue.
- Kept the `activities(account_id, activity_date)` index. `EXPLAIN QUERY PLAN` shows it converts the account_id + activity_date-range filter into an indexed range SEARCH. I did not extend it with `created_at`: the hot query sorts `activity_date DESC, created_at ASC` (mixed direction), which an all-ASC index can't satisfy - the TEMP B-TREE remains either way, so extending would only add write cost. Happy to drop the index entirely if you'd prefer.

**Validation:** ran the built server against a copy of a real 420 MB DB. It drops to ~50 MB, snapshots are aggregate-only, and after rebuild the deterministic values match the pre-change DB exactly - `cost_basis_base` and `net_contribution_base` identical across all 7079 valuation rows I tested with (0 mismatches), lot book consistent with positions, and cross-currency lots retain their account FX. Full `cargo test --workspace` is green.

banconev:
@atsoc0ocsav I ran into similar symptoms like yours (memory impact when looking at 5+ years, timeout/slowdowns). https://github.com/wealthfolio/wealthfolio/issues/1347 might interest you.

afadil:
Merging as-is. thanks @atsoc0ocsav for the quality work across both review rounds.

I did a final pass and found a few small issues (orphaned `snapshot_positions` rows from the reset migration, missing `VACUUM` so disk space isn't actually reclaimed, a TWR edge case on pre-chain negative denominators, minor cleanup). I'll land these myself in a follow-up PR right after this merges.

atsoc0ocsav:
Thank you for the follow up @afadil !
Can't wait to update my local deployment with the new merged version :-)

PR #1221, opened 2026-07-04
fix(addons): load web addons after login so sidebar items appear

r4mbo7:
### Problem
In web mode, installed add-ons don't show up in the sidebar after logging in. The only workaround is to open Settings and disable/re-enable an add-on, after which the Add-ons nav item appears.

### Root cause

Addon discovery in web mode **requires an authenticated session**, but **add-ons were loaded before login**:

1. main.tsx called `loadAllAddons()` at module-load time — before the user authenticates.
2. In web mode this hits list_installed_addons, which the server rejects with no auth cookie. `discoverAddons()` catches the error and returns [], so no sidebar items get registered.
3. After login, AppRoutes mount and the sidebar reads `getDynamicNavItems()` → empty. Nothing re-triggers the load.
4. Toggling an add-on in Settings runs `reloadAllAddons()` while authenticated → add-ons load → nav items register. That's why the workaround "fixes" it.

Desktop is unaffected: there's no auth gate and Tauri's list_installed_addons needs no auth, so the startup load already works there.

### Fix

- main.tsx — gate the eager startup load to isDesktop (desktop behavior unchanged).
- App.tsx — add a web-only AddonRuntimeLoader that calls loadAllAddons() once isAuthenticated becomes true. A useRef guard makes it load exactly once and keeps it StrictMode-safe.

Covers all cases:

- Desktop → loads at startup (via main.tsx).
- Web with password/OIDC → loads right after login.
- Web with no auth required → isAuthenticated is true immediately, so it loads on mount.

## Checklist

### DIFF
--- a/apps/frontend/src/App.tsx
+++ b/apps/frontend/src/App.tsx
@@ -1,15 +1,34 @@
 import { isWeb } from "@/adapters";
 import { setAddonQueryClient } from "@/addons/addons-runtime-context";
-import { AuthGate, AuthProvider } from "@/context/auth-context";
+import { loadAllAddons } from "@/addons/addons-loader";
+import { AuthGate, AuthProvider, useAuth } from "@/context/auth-context";
 import { EventDialogProvider } from "@/features/spending/components/event-dialog-provider";
 import { WealthfolioConnectProvider } from "@/features/wealthfolio-connect";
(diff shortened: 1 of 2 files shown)


afadil:
fixed in https://github.com/wealthfolio/wealthfolio/pull/1222

PR #1263, opened 2026-07-09
feat(rebalance): add transfer-only rebalance mode

ibalboteo:
## Description

Adds a **Transfer Only** rebalance mode — paired origin→destination fund movements without buy/sell operations. Designed for jurisdictions where fund transfers between same-category vehicles are tax-exempt (e.g. Spain's *traspasos*).

### Why

In Spain (and other EU countries), transferring between funds doesn't trigger capital gains tax. Users can rebalance by moving money between funds rather than selling and rebuying — preserving their tax-deferred status.

### Algorithm

- **Two-tier supply**: draws from overweight categories first, then taps in-band headroom only if needed. Never worsens at-target funds.
- **Tolerance-aware**: minimizes transfers to bring categories within the configured drift band — not to exact target.
- **Northwest-corner transport**: supply sorted DESC for minimum transfer pairs. Deterministic output.

### Changes

**Backend (Rust)**
- `ScenarioMode::TransferOnly` variant + `SuggestedTransfer`, `ResidualGap` types
- `TransferOptimizer` (new module) with 8 unit tests
- `RebalanceService` routing + integration test

**Frontend (React)**
- 4th mode button in `ModeSwitch` (responsive 2×2 grid on mobile)
- `TransferPairRow` cards showing origin → amount → destination
- `ResidualGapBanner` when plan is approximate (insufficient supply)
- CSV export and copy-to-clipboard adapted for transfer pairs (fully i18n)
- i18n keys for all 5 locales (EN, ES, DE, FR, ZH)

### Bug fix (included)

Fixed drift-driver sentence displaying actual allocation instead of deviation amount (e.g. "1% below" → "2% below").

## Checklist

### DIFF
--- a/apps/frontend/src/i18n/locales/de/allocation.json
+++ b/apps/frontend/src/i18n/locales/de/allocation.json
@@ -87,6 +87,9 @@
     "hybrid": "Hybrid",
     "hybridShort": "Hybrid",
     "hybridHint": "Bargeld + Verkäufe",
+    "transferOnly": "Nur Umschichtung",
+    "transferShort": "Umschichtung",
+    "transferHint": "Wert verschieben, keine Käufe/Verkäufe",
     "enableSellsTip": "Aktivieren Sie „Verkäufe zulassen“ für dieses Ziel, um diesen Modus zu verwenden"
   },
   "planner": {
@@ -102,7 +105,8 @@
     "driverSentenceUnder": "{{name}} liegt {{current}} % unter einem Ziel von {{target}} %.",
     "modeVerbSells": "Verkäufe und Käufe",
     "modeVerbHybrid": "Bargeld und Verkäufe",
-    "modeVerbCashFlow": "Cashflow-Käufe"
+    "modeVerbCashFlow": "Cashflow-Käufe",
+    "modeVerbTransfer": "Umschichtungen"
   },
   "narrative": {
     "liftAndShrink": "{{verb}} führen {{lifted}} näher an das Ziel und verringern die Übergewichtung von {{shrank}} von {{before}} auf {{after}}.",
@@ -192,6 +196,12 @@
     "tradesSummarySells": "{{buys}} · {{sells}} · {{cash}} neues Bargeld",
     "tradesSummaryDeployed": "{{buys}} · {{cash}} eingesetzt",
     "noTrades": "Keine Transaktionen — alle Positionen sind bereits im Band oder es wurden keine Bestände gefunden.",
+    "transferSummary_one": "{{count}} Umschichtung · {{amount}} bewegt",
+    "transferSummary_other": "{{count}} Umschichtungen · {{amount}} bewegt",
+    "proposedTransfers": "Vorgeschlagene Umschichtungen",
+    "noTransfers": "Keine Umschichtungen — alle Positionen sind bereits im Band.",
+    "residualGapTitle": "Angenäherter Plan — {{gap}} Lücke verbleibend",
+    "descriptionTransfer": "Rebalancing durch Verschiebung von Werten zwischen Positionen — keine Käufe oder Verkäufe erforderlich.",
     "calculatedFooter": "{{name}} · Berechnet am {{date}}",
     "copyAsText": "Als Text kopieren",
     "exportCsv": "CSV exportieren"
@@ -217,7 +227,11 @@
     "amount": "Betrag ({{currency}})",
     "shares": "Anteile",
     "lastPrice": "Letzter Preis ({{currency}})",
-    "reason": "Grund"
+    "reason": "Grund",
+    "from": "Von",
+    "fromName": "Name Quelle",
+    "to": "Nach",
+    "toName": "Name Ziel"
   },
   "copyText": {
     "header": "Rebalancing-Plan · {{date}}",
@@ -225,6 +239,8 @@
     "cashDeployed": "Eingesetztes Bargeld: {{used}} von {{available}}",
     "maxDrift": "Max. Abweichung: {{before}} → {{after}}",
     "proposedTrades": "VORGESCHLAGENE TRANSAKTIONEN",
+    "proposedTransfers": "VORGESCHLAGENE UMSCHICHTUNGEN",
+    "transfer": "UMSCHICHTUNG  {{from}} → {{to}}  {{amount}}",
     "account": "Konto {{account}}",
     "shares": "{{qty}} Anteile",
     "warnings_one": "{{count}} Warnung:",
--- a/apps/frontend/src/i18n/locales/en/allocation.json
+++ b/apps/frontend/src/i18n/locales/en/allocation.json
@@ -87,6 +87,9 @@
     "hybrid": "Hybrid",
     "hybridShort": "Hybrid",
     "hybridHint": "cash + sells",
+    "transferOnly": "Transfer only",
+    "transferShort": "Transfer",
+    "transferHint": "move value, no buys/sells",
     "enableSellsTip": "Enable 'Allow sells' on this target to use this mode"
   },
   "planner": {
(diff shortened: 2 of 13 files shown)


Jonjon-prog:
Hello @ibalboteo, 

I hope you are doing great. 

Thanks for putting this together. 
The transport-problem approach and the test coverage (determinism, multi-category exposure, min-trade pruning) are genuinely nice work.

**Quick context:** 
I've been the one driving the allocation targets / rebalance feature end to end (drift engine, cash-flow planner, sell-to-rebalance, hybrid bands, and most recently the generic constraints + turnover cap in
#1177). 
So I want to make sure this new mode lines up with where we've agreed the feature is headed before it goes further.

To give you a little bit of context on where we are currently: 
Phase 2 (generic do-not-sell/account constraints, turnover cap, plan explanations) just landed. The roadmap for what comes next was discussed with Afadil in #1070, and the agreed direction was explicit: keep core rebalancing smart, explainable, and simple, and push anything niche or jurisdiction-specific.
Tax wrappers, per-country rules — toward addons rather than baking it into core. Account-level tax wrappers specifically are Phase 3, still waiting on a global tax design doc, and the plan there is "lightweight tag + warnings in core, per-jurisdiction catalogue as an addon."

That's why this PR gives me pause: transfer-only rebalancing is really a tax-wrapper feature wearing a ScenarioMode costume. 
Spain's traspasos work because Spanish fund wrappers treat fund-to-fund moves as continuous
No cash leg, no taxable event. But that's a property of the *account wrapper*, not the asset, and it's not universal.

**Concrete counter-example from France** (where I am from):
Since it breaks the "any asset type eligible" assumption (R3) directly: a CTO (ordinary brokerage account) taxes every sale as a realized event and never allows a direct holding-to-holding move.
You're always forced through cash. A PEA, on the other hand, lets you sell and rebuy freely tax-free as long as it stays inside the envelope. 
Same country, same investor, two account types sitting side by side, completely opposite transfer semantics. A model that treats "transfer" as an asset-level concept will produce a plan that's simply wrong for one of these two, even though both are extremely common.

Given that, a couple of questions before I go deeper on the implementation:

1. Would it make more sense to hold this until the tax-wrapper design lands, so transfer eligibility gets gated by account wrapper type instead of being asset-type-blanket?

2. Or, given the whole premise is jurisdiction-specific, should this ship as an addon first — consistent with how we've been deferring tax-adjacent behavior elsewhere?

Not trying to block the idea. 
I just want to flag the overlap with the Phase 3 tax-wrapper plan before more work goes into the implementation. Happy to be wrong if this was already discussed with the maintainers separately. (aka @afadil).

ibalboteo:
Hey, thanks for the detailed review and for explaining the roadmap context. I appreciate it.

**About the "account wrapper vs asset type" thing:**

Your French CTO/PEA example makes sense, but in Spain (my use case and the reason for this PR) it works the other way around. Here, transfer eligibility depends on the instrument type, not the account. Mutual funds can be transferred tax-free within the same account or between different accounts and brokers (Trade Republic to Indexa, Indexa to MyInvestor, etc.), but ETFs or stocks in those same accounts cannot. The account type doesn't matter, what matters is whether the vehicle is a mutual fund.

So there's no single model: sometimes eligibility is account-level (France's PEA), sometimes instrument-level (Spain's mutual funds), sometimes both. This means that a purely account-wrapper-based design (Phase 3) wouldn't fully solve it either.

**Why this ended up in core instead of as an addon:**

I already had an addon doing this: [wealthfolio-rebalancer](https://github.com/ibalboteo/wealthfolio-rebalancer). The problem is that allocation targets are now managed in core and are not exposed through the addon API. Any addon that wants to compute rebalancing has to duplicate the user's target allocations on its own, which is bad UX and hard to maintain.

I moved it to core not because I think it *must* live here, but because right now there's no way for an addon to access the data it needs.

**Either path works for me:**

1. Keep it in core and scope it better once the tax-wrapper design lands.
2. Move it back to an addon, but then expose allocation targets (and ideally drift/holding state) through the API so addons can use them.

Happy to go with whatever you and @afadil prefer. But if the answer is "addon", I think opening up the allocations API should come first. Otherwise addon authors can't really solve these problems without the right inputs.

ibalboteo:
Also, to clarify what this PR actually introduces: it's a different way of presenting the rebalancing operations by pairing them into a single item with a from/to view, instead of showing them as separate movements. So the user can see the full transfer as one thing.

Jonjon-prog:
Thanks for the clarification on the presentation model — makes sense that it's pairing existing movements into a single from/to item rather than introducing new trade semantics.

That said, I think this actually sharpens the concern rather than resolving it.
 If "Transfer" is a UI relabeling of what's otherwise a sell + buy pair, the label itself carries an implicit promise: "this is a tax-neutral movement, not two taxable events." 
 That promise is only true for specific account/instrument combinations (Spanish mutual funds, French PEA swaps) —
as far as I know, a French CTO doesn't support direct holding-to-holding transfers at all: every sale is a taxable event and you're always forced through cash. 
There may be broker-specific exceptions I'm not aware of, but at least in the general case, showing that pair as a single "Transfer" card would misrepresent it to the user. So the eligibility question doesn't go away with the presentation framing — if anything it matters more, since we'd be making a tax-adjacent claim in the UI itself.

Given that, I'd lean toward a more robust foundation than gating this by ScenarioMode: 
a generic account/instrument tag (e.g. extending the constraint model from #1177, or a lightweight `tax_wrapper_type` / `transfer_eligible` tag on Account/Instrument) that core reads at plan time to decide, case by case, whether a pair can be safely shown as "Transfer" vs must stay "Sell + Buy". That covers both your Spain case (instrument-level) and the French PEA/CTO case (account-level) with the same mechanism, without hardcoding per-country rules into core.
And it's consistent with the "lightweight tag + warnings" shape already planned for Phase 3's tax-wrapper
work.

@afadil — tagging you since this touches both the addon-API question and the Phase 2/3 roadmap split; would love your read on whether the tag-based approach above is the right shape, or if you see this differently.

ibalboteo:
That makes sense, the tag-based approach sounds like a good way to handle both account-level and instrument-level eligibility with the same mechanism. I'm on board with that direction.

That said, I still think it would be valuable to expose target allocations through the API, at least as read-only. That way addon developers working on rebalancing or similar features can access them without having to duplicate the user's configuration on their side. @afadil

afadil:
Hi guys,

Sorry, I haven’t had the chance to analyse the PR and your discussion yet, but I agree the add-ons API should expose target allocations. I’m working on an improved add-on system from a security standpoint, and I’ll add the new API asap.

For this PR, please give me more time to think about what could go to the core and what generic shape could be implemented.

Jonjon-prog:
Thanks both. 

 I took everything we discussed and turned it into a design draft — opened as **#1271** (docs only, no code!) so we've got something concrete to poke at together.

A few things from above made it straight in:

- **@ibalboteo, your read-only API ask** is basically the "Brick A" of the whole thing — exposing targets/accounts/holdings read-only so an addon never has to duplicate the user's config. And since @afadil already gave the thumbs up on exposing target allocations, that part feels mostly settled 🎉 — the doc just frames it as riding the improved addon-security work he's building, rather than a parallel surface.
- **The transfer-eligibility worry** (a French CTO pair really shouldn't be dressed up as a tax-neutral "Transfer") is now a first-class rule: the "Transfer" label is *earned* per account/instrument at plan time, otherwise it politely stays a plain "Sell + Buy".
- **The tag-based idea** (a persisted `tax_wrapper_type` / `transfer_eligible` on Account/Instrument, extending #1177) is very much still on the table — I kept it as an explicit open decision next to a transient, non-persisted alternative. The doc leans one way but honestly I don't think we should lock that without @afadil's call.

The bigger picture in #1271: keep core as the one smart, explainable rebalance engine, and let per-country addons hand it *declarative tax rules* (account tax character, capital-gains params, transfer-equivalence classes) that core reads — no second rebalancer hiding in an addon, and nothing writing back into core. There are worked examples for 🇪🇸/🇺🇸/🇫🇷, plus a 🇬🇧 ISA "limit case" I threw in on purpose to show where the vocabulary starts to creak 😅

@afadil — zero rush on this PR! #1271 is really just meant to feed the exact "what belongs in core / what's the generic shape" question you raised. Whenever you get a moment, your read on the α-vs-β and eligibility-source decisions (§14) would be super helpful 🙏

PR #1279, opened 2026-07-11
fix(spending): show savings in the "Where it went" dashboard widget

Jonjon-prog:
Hey @afadil !

 Small UX fix for the spending dashboard.

Money set aside (savings) was invisible in the "Where it went" widget on the dashboard — you'd only see it if you clicked "view all" and scrolled down to the separate cashflow section. That felt counter-intuitive: users want to see at a glance how much they saved, not just where they spent.

**What changed:**
- The "Where it went" widget (both map and list views) now shows a distinct "Saving" row alongside spending categories, sized proportionally (`total = spending + saved`).
- Works in flat mode and in the budget-grouped mode (gets its own group bucket instead of falling into "Other").
- Clicking the row links to the existing cashflow insight anchor (same target as the dashboard's income/spending/saving KPI strip).
- Row-building logic extracted into `buildWhereItWentRows` (`lib/category-rollup.ts`) so it's unit-tested without mounting the full page — added `category-rollup.test.ts` (4 tests: row appears/disappears based on `totalSaved`, sorts correctly among spending categories, delta vs prior period).

**Not in scope:** the "What changed" insights page has the same gap (spending-only movers) — separate, larger fix since its headline copy is spending-specific.

## Test plan
- [ ] `pnpm test` — full suite green (1055 tests)
- [ ] Manual: verified in dev with an account that has savings this month — shows in both list and map mode

---

## Checklist

### DIFF
--- a/apps/frontend/src/features/spending/components/spending-tab-content.tsx
+++ b/apps/frontend/src/features/spending/components/spending-tab-content.tsx
@@ -34,7 +34,7 @@ import { useCategorizationRules } from "../hooks/use-categorization-rules";
 import { useCashActivities, useUncategorizedCount } from "../hooks/use-cash-activities";
 import { useSpendingReport } from "../hooks/use-spending-report";
 import { useSpendingSettings } from "../hooks/use-spending-settings";
-import { topCategoryId } from "../lib/category-rollup";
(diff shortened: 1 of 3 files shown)


afadil:
@Jonjon-prog Merci chef!

Jonjon-prog:
> @Jonjon-prog Merci chef!

Derien 😄 Un autre petit fix arrive aussi sur le budget

PR #1288, opened 2026-07-12
feat(spending): track savings goals in the budget

### LINKED ISSUE #1: Does not run on macos intel based chips
Thank you for you work! I would like to try this app, but unfortunately the binary does not work for intel based macs.
Can you do a build for intel based macs or is there an issue that this is not possible?

Thank you.

Jonjon-prog:
Hey @afadil ! 

This PR addresses a fairly frustrating gap: I could track spending and income in the budget, but there was **no way to set a savings goal** or see progress toward it. 
Neither on the Budget page nor on the dashboard. While wiring up the dashboard widget properly, I ran into two real backend bugs that were silently zeroing out savings "actuals", details below.

## What changed

**Budget page** (`/spending/budget`):
- New "Saving" panel (next to "Income Sources") — a $ target per top-level savings category, same mechanics as income (no groups/rollover, just target vs actual).

**Spending dashboard**:
- New "Saving" widget — a big number + status (reached/not reached), a cumulative day-by-day SVG chart with a dashed line at the goal (same visual style as "Monthly budget"), and one ring per savings category with a target set (matching the "By category" pattern).

## Backend bugs found and fixed

While wiring the new widget's day-by-day chart to real data, I hit two independent bugs — both in the same "savings treated as second-class" vein:

1. **`budget::service::actuals_by_month`** called the plain `classify_activity` classifier, which has **no `Saving` branch at all** (a CASH transfer-out always classifies as `Expense` there). As a result, savings actuals computed in the budget always came out to zero, even with a correctly configured category and target. Fix: use `classify_activity_for_aggregation` (the classifier that's aware of transfer pairs), the same one `insight::service` already uses for the dashboard.

2. **`analytics::service::add_report_breakdown_allocations`** — the function that builds the day-by-day-by-category series (`report.byDayByCategory`) — had day-bucket emission **hardcoded to `SPENDING_TAXONOMY`** only, regardless of the flag passed in. The monthly savings total (`savings_breakdown`) was correct, but the daily breakdown was always empty for savings — the new widget's chart stayed flat even with real transactions that month.

Both bugs could affect **any savings transfer** in the existing calculations — nobody had noticed because no UI surfaced these values before this PR.

## Architecture limitation worth flagging

`Saving` classification only triggers through **one path**: a CASH outbound transfer whose other leg falls outside the spending-account universe (`activity_classification.rs`, `classify_activity_for_aggregation`). There's no other path — in particular, if a user manually categorizes a non-transfer activity under a savings category, it will **never** count toward savings actuals, even though the category tag is explicit. This is existing structural behavior (not introduced by this PR), but worth documenting — might be worth a separate product discussion on whether "Saving" should stay purely account-movement-based or also accept manual categorization as a signal.

## Unrelated bug found while testing

`quick-categorize-popover.tsx`: the category picker used by the rule-creation flow (`scope="both"`) silently excluded savings categories from the list — an asymmetric condition (`if (scope === "saving")` instead of `scope === "saving" || scope === "both"`) present since the component was first written. It was impossible to create an auto-categorization rule for savings. 
One-line fix included in this PR (small enough that it didn't warrant a separate one).

## Tests

- `native_amounts_for_actuals_classifies_external_transfer_out_as_saving` + `_treats_within_spending_transfer_as_neutral` (Rust, budget) — guards fix #1.
- `report_breakdown_includes_savings_in_daily_category_buckets_when_enabled` (Rust, analytics) — guards fix #2.
- `rollUpToTopLevel` (3 tests) + `sumByDayForTaxonomy` (2 tests) on the TS side — pure logic extracted from the dashboard widget to make it testable.
- Manually reproduced against a copy of a real user DB before/after each fix.

## Test plan
- [ ] `cargo test -p wealthfolio-spending --lib` — 102/102
- [ ] `pnpm test` (frontend) — 1060/1060
- [ ] `cargo clippy --workspace --lib --bins` — 0 warnings
- [ ] Manual: target, actual, and chart are correct on both the dashboard widget and the Budget panel

---

## Checklist


Jonjon-prog:
Hey @afadil , 

Just a quick heads up to have a little look into this when you got time.

afadil:
@Jonjon-prog Thanks for working on this.

We already have a Savings budget group and long-term saving goals under Investments, so this introduces overlapping new concept.

I suggest using the the existing Savings group and savings categories. The group could show monthly targets and actual categorized transfers for Emergency Fund, Retirement, Investments, etc. This keeps Budget focused on monthly cash flow while Investment Goals remain for long-term objectives. The backend classification fixes are still valuable and could be separated from the UI changes.

PR #1294, opened 2026-07-12
fix(addons,dev): let dev-server addons use the network broker

markrickert:
## Description

Thank you for this great project! 

I was working on a custom addon and realized that dev-server addons (pnpm dev:server) couldn't use the brokered network API. The broker looks up an addon's manifest on disk, but dev addons are hot-loaded in-memory and never installed, so every request failed with: `"Addon not found"`.

This fixes it with an in-memory `DevAddonRegistry`: the frontend registers a dev addon's manifest with the backend on load and clears it on unload, auto-approving its declared hosts (dev mode has no install-time approval dialog). The network broker checks this registry before falling back to the disk-backed lookup. Desktop only. Added tests covering the manifest-bypass path (host approval, secrets-permission checks, audit logging) and the frontend register/unregister sync.

I also made sure to use the `#[cfg(debug_assertions)]` directive so that release versions are completely precluded from using this functionality.

## Checklist


markrickert:
Added a fix that moved the two desktop-only dev-addon network commands into platform-specific adapters so the web tests would pass. Rebased on `main` and pushed.

PR #1307, opened 2026-07-13
feat(addon-sdk): Allow addons to access and modify transfer link data

markrickert:
## Description

I was working on an addon and would love to have the ability to work with transfer linking. The transfer matching logic already exists in the backend the addon SDK's ActivitiesAPI just didn't have any way to read or write transfer pairs.

This PR exposes that existing functionality to addons instead of building anything new. Added `getTransferPair`, `findTransferMatchCandidates`, `saveTransferPair`, `linkTransfer`, and `unlinkTransfer` to the SDK, wired through the permission system. Docs updated.

Added tests for both sides: one making sure the SDK bridge routes each new call to the right internal function and blocks it without the activities permission, and one making sure the auto-detection scanner correctly flags an addon that calls these functions.

I have the addon published here if you want to see example usage: https://github.com/markrickert/wealthfolio-transfer-finder

## Checklist


markrickert:
Update: published an addon that can use this feature when ready: https://github.com/markrickert/wealthfolio-transfer-finder

9DA2E696-961F-47C2-A5D0-60FF7AFDF1DA-86369-0000368CB1526FDF

afadil:
@markrickert, Thanks for putting this together! The permission, iframe bridge, documentation, and test coverage are appreciated.

Two small public SDK contract issues on the latest PR head:

1. `getTransferPair` is described as returning a pair “if any,” but the backend rejects when the activity is unpaired. Please either return `InternalTransferPairResponse | null` or document and expose a stable, distinguishable not-paired error.

2. `InternalTransferPairRequest` makes `transferOutId` and `transferInId` independently optional, while the backend enters update mode if either is present and then requires both. Please model create and update as separate request shapes so partial updates are rejected by TypeScript.

Since these public contracts would be harder to correct after release, I’m requesting changes. Thanks again for the contribution!

markrickert:
@afadil Thanks for the feedback and sorry it took me so long to get back to this! I completely understand and have updated this PR to address your comment. 

I chose to make it a `null` return instead of a not-paired error because an unpaired activity is a routine state, not an exceptional failure. Returning null mirrors ``getHolding(accountId, assetId)` in the SDK and saves add-on authors from wrapping standard lookups in `try { ... } catch` like i have in my example addon.

Ready for another review! Should be all there and rebased on `main`.

PR #1321, opened 2026-07-15
fix: Addon sandbox fails to load on WebKit (iOS) due to CSP `'self'` in sandboxed iframe

ibalboteo:
## Description

Addons fail to load on iOS browsers (Chrome/Safari) with a timeout error. The addon sandbox iframe never finishes bootstrapping because WebKit blocks all scripts and stylesheets.

**Console errors on iOS:**
```
Refused to load https://example.com/assets/addon-sandbox-m9pemPyl.js because it does not appear in the script-src directive of the Content Security Policy.
Refused to load https://example.com/addon-sandbox.css because it does not appear in the style-src directive of the Content Security Policy.
Failed to load addon rebalancer: Error: Timed out loading addon 'rebalancer' during sandbox document loaded
```

**Works on:** Desktop Chrome (Blink)  
**Broken on:** iOS Chrome, iOS Safari, any WebKit-based browser

### Root Cause

The addon sandbox iframe uses `sandbox="allow-scripts"` without `allow-same-origin`, which assigns the iframe an **opaque (null) origin**. The CSP for the sandbox used the `'self'` keyword in `script-src` and `style-src`.

**Blink (Chrome desktop)** resolves `'self'` against the document's URL even in sandboxed iframes — so it matches the server origin and loads assets fine.

**WebKit (Safari/iOS)** strictly resolves `'self'` against the document's effective origin, which is opaque/null for sandboxed iframes. This means `'self'` matches nothing, and all same-origin script/style loads are blocked.

### Fix

**1. api.rs** — Dynamic CSP for addon sandbox

Replaced the static `ADDON_SANDBOX_CSP` constant with a function that constructs the CSP using the actual request origin (from `Host` + `X-Forwarded-Proto` headers) instead of relying on `'self'`:

```rust
fn addon_sandbox_csp(origin: &str) -> String {
    format!(
        "default-src 'none'; script-src {} 'unsafe-inline' blob:; ...",
        origin, origin, origin
    )
}
```

This produces `script-src https://wealthfolio.example.com 'unsafe-inline' blob:` — which works correctly regardless of the browser's interpretation of `'self'` in sandboxed contexts.

**2. addon-sandbox.html** — Broadened meta tag CSP

The HTML meta tag CSP (used in Tauri mode where no HTTP server sets headers) was replaced:

```diff
- script-src 'self' 'unsafe-inline' blob: http://localhost:* http://127.0.0.1:* http://tauri.localhost https://tauri.localhost tauri: asset:
+ script-src 'unsafe-inline' blob: https: http: tauri: asset:
```

Removed `'self'` (broken on WebKit in sandboxed iframes) and the enumerated localhost origins. Replaced with scheme sources (`https:`, `http:`). This is safe because:

- In **web mode**: the server's HTTP header CSP (specific origin) intersects with the meta tag CSP (broad). Effective policy = the server's strict CSP.
- In **Tauri mode**: only the meta tag applies, but the sandbox still has `connect-src 'none'` (no network), `sandbox="allow-scripts"` only (no DOM/storage access), and addon code is loaded from local files.

### Security Impact

- `connect-src 'none'` remains — addons cannot make network requests
- `sandbox="allow-scripts"` without `allow-same-origin` remains — no access to parent cookies, localStorage, or DOM
- `form-action 'none'`, `object-src 'none'`, `base-uri 'none'` remain
- The effective CSP in web deployments is unchanged (server header is the binding constraint)

### Testing

Verified addon loading on iOS Safari (via Safari Web Inspector) — addon sandbox bootstraps and renders correctly. Desktop Chrome continues to work as before.

## Checklist

### DIFF
--- a/apps/frontend/addon-sandbox.html
+++ b/apps/frontend/addon-sandbox.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'none'; script-src 'self' 'unsafe-inline' blob: http://localhost:* http://127.0.0.1:* http://tauri.localhost https://tauri.localhost tauri: asset:; style-src 'self' 'unsafe-inline' http://localhost:* http://127.0.0.1:* http://tauri.localhost https://tauri.localhost tauri: asset:; img-src 'self' data: blob: http://localhost:* http://127.0.0.1:* http://tauri.localhost https://tauri.localhost tauri: asset:; font-src 'self' data: blob: http://localhost:* http://127.0.0.1:* http://tauri.localhost https://tauri.localhost tauri: asset:; connect-src 'none'; object-src 'none'; base-uri 'none'; form-action 'none'"
+      content="default-src 'none'; script-src 'unsafe-inline' blob: https: http: tauri: asset:; style-src 'unsafe-inline' https: http: tauri: asset:; img-src data: blob: https: http: tauri: asset:; font-src data: blob: https: http: tauri: asset:; connect-src 'none'; object-src 'none'; base-uri 'none'; form-action 'none'"
     />
     <script>
       (function () {
--- a/apps/server/src/api.rs
+++ b/apps/server/src/api.rs
@@ -85,21 +85,48 @@ pub async fn readyz() -> &'static str {
 pub struct ApiDoc;
 
 const SERVER_CSP: &str = "default-src 'self'; script-src 'self' blob:; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob: https:; font-src 'self' data:; connect-src 'self' https://wealthfolio.app https://auth.wealthfolio.app https://connect.wealthfolio.app https://connect-staging.wealthfolio.app; frame-src 'self' blob: about:; child-src 'self' blob: about:; object-src 'none'; base-uri 'self'; form-action 'self'; frame-ancestors 'none'; worker-src 'self' blob:";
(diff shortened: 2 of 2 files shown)


ibalboteo:
@afadil

ibalboteo:
Could any maintainer please take a look at this PR?

This is a regression in the software, and addons are no longer working as they did before. If there are any concerns with the proposed solution, I’m happy to discuss them and consider a different approach.

However, I believe it’s important to get a patch out for this issue as soon as possible, since it is currently affecting existing functionality.

@afadil

afadil:
@ibalboteo Thanks for investigating this. 

Working on a new architecture direction for the addon sandbox architecture for the upcoming 3.7 release. so we won’t use the origin-specific CSP approach from this PR.
Runtime JS/CSS will instead be transferred into the opaque sandbox as Blobs through an authenticated message channel. Packaged images, fonts, and other resources will use a private per-addon ctx.assets registry. This keeps the sandbox origin-independent and network-free across Safari/iOS, Chromium, Firefox, and Tauri.

The replacement is nearly ready and will be released this week. 

Thanks again for identifying the WebKit issue and proposing a fix.

afadil:
Resolved by #1431, which has now been merged. It fixes the WebKit/iOS opaque-origin sandbox loading regression with an origin-independent Blob bootstrap and private per-addon asset registry, superseding the origin-specific CSP approach proposed here. Thank you again for identifying the issue and contributing the investigation.

PR #1330, opened 2026-07-17
feat(activities): Add in-kind asset exchange support

blastik:
imagen

## Description

Some countries let you switch directly between funds without it counting as a taxable sale — Spain's _traspasos_ are the best-known example. Move from Fund A to Fund B and no capital gain is realized until you actually cash out, and even then it's measured against Fund A's _original_ cost basis, not the switch-day value.

Wealthfolio had no way to represent that. The only building blocks were `TRANSFER_IN`/`TRANSFER_OUT` (same asset, different accounts) or `SELL`+`BUY` (different asset, but a real disposal). This came up directly while building the [MyInvestor importer addon](https://github.com/blastik/myinvestor-importer-addon): it currently books traspasos as `SELL`+`BUY`, which keeps net worth correct but shows a phantom realized gain on every switch, because Wealthfolio has no way to know the gain is deferred.

This PR adds a proper primitive for it: a paired **EXCHANGE_OUT / EXCHANGE_IN** activity (an `ADJUSTMENT` subtype pair, same account, sharing a `source_group_id` the same way transfers already do). The closing leg removes the old asset's lots via FIFO; the opening leg carries that cost basis over to the new asset — prorated per source lot, preserving original acquisition dates, converted to the new asset's currency if needed. No cash moves, `net_contribution` is untouched, and no realized gain gets recorded.

It isn't Spain-specific on purpose — the same primitive fits any tax-deferred exchange (share-class conversions, fund mergers, etc.), Spain is just the common case that motivated it.

### What's included

- New "Exchange" entry in the Add Activity picker, with create and edit support (editing mirrors the existing internal-transfer edit flow)
- Cost-basis carryover with lot-level fidelity (dates, currencies)
- Health check for orphaned/mismatched exchange legs
- `sourceGroupId` on the addon-sdk bulk `import()` call, so addons can emit paired rows directly instead of two `create()` calls

### Testing

Backend: unit + integration tests for lot proration, currency conversion, pairing validation, and bulk import (`cargo test --workspace` and clippy clean). Frontend: component/hook tests for the new form and edit flow (`vitest` clean). E2E: new Playwright case covering the create flow end to end.

## Checklist


Jonjon-prog:
Hello @blastik 

There is already a discussion about that here:  #1263 
It's a more complexe thing, I will let you read 😄

blastik:
Hey @Jonjon-prog, thanks for the pointer! Read through #1263 — good stuff, but I think we're actually solving different layers of the same itch, not overlapping.

#1263 is a planner: it tells you *what* traspaso to make to get back on target, then you go do it at your broker. It doesn't touch the activity ledger at all.

This PR is the other half: once a traspaso has actually happened (whether you got there via #1263's suggestion or just did it yourself), there's finally a correct way to *record* it — cost basis carries over from the old fund to the new one, no phantom realized gain, net worth unaffected. Right now the only option is logging it as SELL+BUY, which is what got me here in the first place (built [an importer addon](https://github.com/blastik/myinvestor-importer-addon) for MyInvestor and ran into this).

So honestly they'd pair nicely — #1263 could suggest the move, and this gives you somewhere accurate to log the result. Happy to chat if it'd help align the two.

victor-marino:
Hey there @blastik.

So happy to see someone already digging into Spanish-style "traspasos" for Wealthfolio :) I really hope the maintainers accept your contribution as there are almost no portfolio tracking apps that support this today!

After trying out your branch, I found a small issue: your current implementation only accepts one single date for each "exchange pair".

While this can be fine for "switch"-type exchanges that execute the sale and the purchase on the same date (such as those allowed between some Vanguard funds), in most "traspasos" you're going to have a 2-3 day gap between the origin sale and the destination purchase.

And I think it's important to account for this gap, because it will actually affect the return calculations (your money will remain outside the market and will be unaffected by market movements during those 2-3 days).

Do you think it would be feasible to add a second "date" field so the sale and the purchase can be correctly registered?

Also, as far as I can see, you can select any asset for the "closing" position, even if you never had it in your account in the first place. While it's ultimately the user's responsibiliy to enter valid data, would it make sense to have a filter in place so that you can only "exchange out" assets you already had in that account, and limit the transaction to the number of shares you had?

blastik:
@victor-marino good catch. ive added another commit to tackle your comments. can you give it a try ? lets hope that we reach to an agreement with this feature

victor-marino:
Wow, that was fast :)

Yep, I think it's perfect right now! I would love to have this available in the production version already.

@Jonjon-prog, hope you can seriously consider this PR. To be clear, and like @blastik pointed out, this is a completely separate (and independent) feature from the rebalancing calculations #1263.

What this does is add a new activity type that allows to register transfers/exchanges between two different funds, preserving the original cost basis and thus not realizing any capital gains. This is the only way to properly track unrealized gains/losses in countries where transfers between different funds are allowed (such as Spain), because registering an actual sale+purchase obviously clears the tax history of the sold shares.

#1263 can be used to suggest the user when to make fund transfers for rebalancing purposoes, but won't help with the main problem: tracking them.

Also keep in mind fund transfers go way beyond mere rebalancing acts: it's very frequent in Spain for people to freely move money from a more expensive to a cheaper index fund, or temporarily move part of their equity portfolio to a money market fund in preparation for a planned expense, etc. Rebalancing between stocks and bonds is one of the main use cases, but not the only one.

kryofin:
I would love to see this or something like this get merged. I have very similar US transactions where Vanguard mutual funds were converted into the corresponding ETF (VTSAX -> VTI, for example) within an account. Not supporting this is blocking TWR and MWR calculations.

markitosgv:
Really useful for spanish investors ;-) +1

cculbertson81:
> I would love to see this or something like this get merged. I have very similar US transactions where Vanguard mutual funds were converted into the corresponding ETF (VTSAX -> VTI, for example) within an account. Not supporting this is blocking TWR and MWR calculations.

I have a similar use case for conversion from Vanguard Investor Shares to Vanguard Admiral Shares, for example, VTSMX -> VTSAX. The EXCHANGE_OUT / EXCHANGE_IN  subtype nomenclature threw me for a loop briefly, since Vanguard has an Exchange event type which is a [potentially] taxable, cost basis-altering event (redistributions, etc, easily representable as SELL / BUY), but the description of the functionality lines up with my share conversion case.

PR #1353, opened 2026-07-24
feat(spending): Let users set a categorization rule's account scope

### LINKED ISSUE #1350: Spending rule scopes aren't visible in the UI
The Spending and Budget [docs](https://wealthfolio.app/docs/guide/spending-budgets/#rules) list Scope ("All accounts or a specific account") as a setting but the settings rule editor but `.../settings/spending/rules` never offers an option to set it when creating or editing. It does look like the backend has this implement and the only missing gap is the UI.

kryofin:
### Description

Fixes #1350

The [docs](https://wealthfolio.app/docs/guide/spending-budgets/#rules) describe a rule **Scope** ("All accounts or a specific account"), and the whole backend supports it — but the rule editor never rendered a control, so the feature was unreachable. `RuleFormValues` carried `isGlobal` as a phantom field (hardcoded `true` on create, echoed back on edit) and never sent `accountId`. Result: every UI-created rule was global, and account-scoped rules created via the AI assistant were invisible in the list and silently un-editable.

The data layer was already complete and tested: `is_global` + `account_id` columns with a FK and a CHECK constraint, `validate_rule_scope` on create/update, matcher enforcement, and pass-through API/IPC. This PR is **frontend-only** — no schema, migration, or backend changes.

### Changes

- **`rule-form.tsx`** — replace the phantom `isGlobal` with a single nullable  `accountId` field and an "Applies To" select. `isGlobal` is derived at the  payload boundary (`accountId === null`), so the two fields can't disagree
- A rule pointing at an untracked/archived account keeps a synthetic option so its id survives an unrelated edit instead of the trigger going blank.
- **`spending-rules-page.tsx`** — feed the form tracked spending accounts only (a rerun walks just those, so a rule scoped elsewhere could never fire), and send `accountId` explicitly on both create and update.
- **`rule-item.tsx`** — show an account-name pill on scoped rules; global rules render nothing.
- **`rule-edit-modal.tsx`** — thread the account options through.

## Checklist

### DIFF
--- a/apps/frontend/src/features/spending/components/rule-edit-modal.tsx
+++ b/apps/frontend/src/features/spending/components/rule-edit-modal.tsx
@@ -8,7 +8,12 @@ import {
   DialogTitle,
 } from "@wealthfolio/ui";
 
-import { RuleForm, type RuleFormCategoryOption, type RuleFormValues } from "./rule-form";
+import {
+  RuleForm,
+  type RuleFormAccountOption,
+  type RuleFormCategoryOption,
+  type RuleFormValues,
+} from "./rule-form";
 import type { CategorizationRule } from "../types/rule";
 
 interface RuleEditModalProps {
@@ -16,6 +21,7 @@ interface RuleEditModalProps {
   onClose: () => void;
   rule?: CategorizationRule;
   categoryOptions: RuleFormCategoryOption[];
+  accountOptions: RuleFormAccountOption[];
   onSave: (values: RuleFormValues) => void;
   isLoading?: boolean;
 }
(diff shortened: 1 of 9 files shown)


kryofin:
@afadil Any thoughts on this PR? It just exposes existing functionality and makes it easier for Actual Budget users to move to Wealthfolio. There are a couple of additional gaps for rules (amount being a big one) that I can work on if PRs like this are acceptable to you.

afadil:
@kryofin Thanks for the contribution! Sorry for the delay, This looks good. The spending module is still pretty new, so please feel free to submit more improvements.

kryofin:
@afadil Sounds good. I already have a PR open for using amounts but I need to rebase on this one and resolve conflicts. I will update it after another round of testing and add some nuances for you to consider.

PR #1354, opened 2026-07-24
feat(custom-provider): support POST requests and format more date placeholders

charlie890414:
Extends the custom market-data provider so sources can fetch data from APIs that require POST requests, and adds extra formatted-date placeholders for building request URLs and bodies.

What changed:
image

Ref issue: https://github.com/wealthfolio/wealthfolio/issues/812


afadil:
@charlie890414 Thanks for working on this! few issues to address before merging:

- **Fixed-URL POST providers can be skipped.** Provider selection only checks the URL for `{SYMBOL}` or `{ISIN}`. If the placeholder is only in the POST body, the provider isn’t selected unless explicitly assigned to the asset. Check both URL and body:`crates/core/src/quotes/custom_scraper_provider.rs`

- **Body placeholders aren’t tested correctly.** The live tester only looks for placeholders in the URL, so inputs for body-only `{ISIN}` and `{MIC}` aren’t shown or passed correctly. Detect placeholders in both URL and body: `apps/frontend/src/pages/settings/market-data/use-source-runtime.ts`

- **Existing headers can be removed.** Opening a GET provider with `Content-Type: application/json` can silently remove that header. It would be safer to set the default content type when sending a POST request instead of modifying saved headers: `apps/frontend/src/pages/settings/market-data/source-config-panel.tsx`

- **Request building is duplicated.** Test and normal quote requests build POST requests separately, which can make their behavior drift. A shared backend helper would keep placeholder expansion and headers consistent.

Thanks again!

bfg100k:
Can we add support for good old standard HTML forms too? i.e. application/x-www-form-urlencoded: Encodes data as key-value pairs separated by ampersands (e.g., key1=value1&key2=value2)

charlie890414:
@afadil
Thanks for the review! I’m working through the points you mentioned.

@bfg100k
I’ll also try to add support for `application/x-www-form-urlencoded`, either in this PR or in a follow-up PR.

afadil (changes requested):
Thanks for addressing the earlier feedback. A few things detected by codex and still need attention:

- **[P2] Invalid date formats can panic the backend**  
  `crates/core/src/custom_provider/model.rs:104`  
  I reproduced this with `{FROM:%Q}`; the process exits with a Rust panic instead of returning a validation error. Please validate the Chrono format before calling `.to_string()`.

 - **[P2] Quick-start templates keep the previous POST method and body**  
  `apps/frontend/src/pages/settings/market-data/use-source-runtime.ts:383`  
  `applyTemplate` replaces the URL and mappings but not `method` or `body`. Selecting a JSON template while editing a POST source can therefore send the template URL using the old POST body. Reset these to `GET` and an empty body, or include them in `ProviderTemplate`.

 - **[P2] Formatted date placeholders are not expanded in preview or copied values**  
`apps/frontend/src/pages/settings/market-data/live-preview-pane.tsx:92`  
`apps/frontend/src/pages/settings/market-data/use-source-runtime.ts:159`  
 The backend expands `{FROM:%Y%m%d}`, but the UI preview and Copy button leave it unchanged. A shared frontend placeholder-expansion helper would keep the preview consistent with the actual request.

- **[P3] The new UI text is not fully localized**  
 `apps/frontend/src/pages/settings/market-data/source-config-panel.tsx:519`  
 `apps/frontend/src/pages/settings/market-data/live-preview-pane.tsx:739`  
 Several labels are hardcoded in English, and the new translation keys only exist in `en/settings.json`. German, French, and Spanish currently fall back to English. Please add translation keys and update the locale files.

 The `application/x-www-form-urlencoded` request from the comments looks like an additional feature, not a regression in this PR, so handling it in a follow-up seems reasonable. Thanks again!

charlie890414:
@afadil Thanks for the detailed review. I’ve addressed all four points.
Could you please take another look? Thanks!

afadil (approved):
@charlie890414 Thanks. merging now

PR #1369, opened 2026-07-29
fix(activity): reject cash dividend/interest edits over hidden quantity/price fields

Baykenn:
## Summary

Editing an existing Cash-type DIVIDEND or INTEREST activity failed to save with "Quantity: Received quantity must be greater than 0" and "Price: FMV per unit must be greater than 0", even though those fields are hidden in Cash mode.

Root cause: `quantity`/`unitPrice` were defined with a top-level `z.coerce.number().positive().optional()`, so `.positive()` ran unconditionally. `.optional()` only skips validation for `undefined`, not `0`, and existing cash-dividend/interest records store `quantity=0`/`unitPrice=0` in the DB, so the stale zero values always failed validation.

Fix: move the `positive()` checks into the existing `superRefine` block (already gated on `isAssetBacked` for dividend-form.tsx / `isStakingReward` for interest-form.tsx), so they only apply when the fields are actually shown (DRIP/DIVIDEND_IN_KIND for dividends, STAKING_REWARD for interest).

## Test plan

- [x] `vitest run` on `dividend-split-forms.test.tsx` (21 tests passing)
- [x] `tsc --noEmit` clean
- [x] Manually edit an existing Cash-type dividend/interest activity in the app and confirm it saves without spurious errors

### DIFF
--- a/apps/frontend/src/pages/activity/components/forms/dividend-form.tsx
+++ b/apps/frontend/src/pages/activity/components/forms/dividend-form.tsx
@@ -97,9 +97,6 @@ export const createDividendFormSchema = (t?: TFunction) =>
             "FMV per unit must be a number.",
           ),
         })
-        .positive({
-          message: msg(t, "activity:form.err_fmv_gt_zero", "FMV per unit must be greater than 0."),
-        })
         .optional(),
       quantity: z.coerce
         .number({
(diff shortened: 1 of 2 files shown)


afadil:
fixed in v3.6.3 in bot mobile and web

PR #1370, opened 2026-07-29
Show average cost line on holding history chart

Baykenn:
## Summary
- Add a dashed average-cost (PRU) reference line to the symbol/holding history chart (`var(--foreground)`, adapts to light/dark theme), rendered via a new optional `averageCost` prop on `HistoryChart`
- Fold this into the chart's existing buy/sell activity marker toggle so a single button shows both together, instead of adding a second control
- Persist the toggle choice locally so it survives reloads
- Wire `averageCost` from the already-computed holding average price into both the desktop and mobile `AssetHistoryCard` render paths; the line is only shown when the position has shares (skipped for fully closed positions)

## Test plan
- [x] Open a holding with buy/sell activity and a non-zero average cost; toggle the History icon on the chart and confirm both the buy/sell markers and the dashed average-cost line appear together
- [x] Reload the page/app and confirm the toggle state persists
- [x] Verify the average-cost line is not shown for holdings with zero shares (e.g. fully closed positions)
- [x] Check appearance in both light and dark theme

image

### DIFF
--- a/apps/frontend/src/components/history-chart-symbol.tsx
+++ b/apps/frontend/src/components/history-chart-symbol.tsx
@@ -8,6 +8,7 @@ import {
   Area,
   AreaChart,
   ReferenceDot,
+  ReferenceLine,
   ResponsiveContainer,
   Tooltip,
   XAxis,
@@ -65,6 +66,7 @@ export default function HistoryChart({
   interval,
   activityMarkers = [],
   onActivityMarkerClick,
+  averageCost,
   height = 350,
 }: {
   data: HistoryChartData[];
@@ -72,6 +74,7 @@ export default function HistoryChart({
   height?: number;
   activityMarkers?: HistoryChartActivityMarker[];
   onActivityMarkerClick?: (marker: HistoryChartActivityMarker) => void;
+  averageCost?: number;
 }) {
   const [hoveredMarker, setHoveredMarker] = useState(false);
   const markerByTimestamp = useMemo(() => {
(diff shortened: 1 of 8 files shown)


afadil:
@Baykenn Thanks for the PR, looks great in the screenshot! Two things before merging:

1. **Balance privacy:** average cost is treated as private elsewhere — `AssetDetailCard` hides `averagePrice` when balance privacy is on. The chart label prints it unconditionally. Could you skip the line (or at least the label) when `isBalanceHidden` is true? It's already available in `AssetHistoryCard`.

2. **Toggle label:** the button's tooltip and `aria-label` still say "show/hide activity markers", but it now also controls the average-cost line. Please update the i18n strings to mention both.

One thing to double-check (not blocking): with `ifOverflow="extendDomain"`, a holding whose price has moved far from its average cost will get its price curve compressed when the toggle is on. Worth a quick look to confirm it's acceptable.

Baykenn:
Thanks for the review! Pushed a fix for both:

1.Balance privacy: the average-cost line (and its label) is now skipped entirely when isBalanceHidden is true, matching how AssetDetailCard already hides averagePrice.

2.Toggle strings: updated the tooltip and aria-label (all locales) to say "activity markers and average cost line" since the button now controls both.

On the ifOverflow="extendDomain" point: kept it as-is intentionally.
It's what keeps the reference line visible at all when the average cost falls outside the price's natural range  without it, the line just clips off-chart for any holding with a large gain/loss. The resulting curve compression is the visible tradeoff of showing "how far price has moved from cost," so it reads as intentional rather than a bug to me, but happy to revisit if you'd rather it behave differently.

PR #1387, opened 2026-08-02
fix(storage): reclaim disk and clean orphans in the derived read model reset

afadil:
Follow-up to #1218, as noted when merging it. Four fixes found in a final review pass, plus the cleanup that PR left behind.

**No release should be cut between #1218 and this PR.** The reset migration is one-shot per database: anyone upgrading on #1218 alone runs it without the VACUUM and permanently strands the orphaned `snapshot_positions` rows, which a later fix cannot re-run.

## 1. Reset migration leaves orphaned `snapshot_positions`

`run_migrations` sets `PRAGMA foreign_keys = OFF`, so `DELETE FROM holdings_snapshots WHERE source = 'CALCULATED'` never fires the `ON DELETE CASCADE` into `snapshot_positions`. Keyframes are sparse, so a rebuilt snapshot does not necessarily reuse a deleted row's `snapshot_id`, and the delete-then-insert on write never reclaims them. The 2026-05-22 and 2026-06-21 migrations already delete these rows explicitly for the same reason.

Verified on a real 356 MiB pre-migration database: 14,408 position rows → 22 (matching the 16 preserved source snapshots), 0 orphans.

## 2. No `VACUUM`, so the disk space is never reclaimed

The earlier revision of #1218 vacuumed; the reset migration that replaced it does not, and nothing else in the codebase does. Deleting ~90% of rows releases pages to SQLite's freelist but leaves the file the same size on disk.

`VACUUM` cannot run inside a transaction and Diesel wraps migrations in one, so it lives in its own migration with `metadata.toml` (`run_in_transaction = false`). Keeping `000001` transactional preserves its atomicity — it contains `ALTER TABLE ... ADD COLUMN` statements that would fail on a re-run if a crash left it recorded as unapplied — while `VACUUM` is idempotent and safe to re-run.

Measured on the same real database: **356 MiB → 79 MiB**, freelist 0, `integrity_check` ok, activities byte-identical.

## 3. TWR: pre-chain negative denominators are a behavior regression

The negative-denominator check added in #1218 runs *before* the pre-chain branch, so it also fires on days that were previously skipped silently: those before the chain starts whose opening value is not positive. That is junk history from before an account was first funded, and nulling the headline for it reintroduces the exact "dormant junk history poisons the headline" failure mode the low-base exclusion was written to remove.

Gated on `chain_started || prev_value > 0`. A negative denominator after the chain starts, or on a pre-chain day with a positive opening value, stays fatal — the test added in #1218 for that case still passes unchanged.

## 4. Cleanup

- `backup_database_from_path` was extracted for the in-app strip migration that the reset replaced and has had no callers since — inlined back.
- Stale "two callers" doc on `compute_position_cost_basis_from_lots` naming the removed startup backfill.
- `lot_record_to_snapshot_lot` silently turned an unparseable `open_date` into `Utc::now()` and unparseable decimals into zero — each now warns and names the field.
- Reverted the two blank-line-only hunks in the server/tauri context providers.

## Testing

- 1,661 core + 232 storage-sqlite tests pass; `cargo fmt --check` clean.
- New test applies the **full embedded migration chain** to a temp database. It fails with `cannot VACUUM from within a transaction` if `metadata.toml` is missing or not honored by `embed_migrations!` — verified by deleting the file and watching it fail, so it genuinely guards the assumption rather than passing vacuously.
- Migration test extended with a `snapshot_positions` fixture (carrying the inert `CASCADE` clause, mirroring the real schema under `foreign_keys = OFF`) asserting orphans are removed while converted and preserved snapshots keep their rows.
- End-to-end on a real 356 MiB database predating both migrations, results above.

## Note on the ANALYZE concern

I had initially flagged the gated `ANALYZE` as baking empty-table statistics after the reset. That turned out to be wrong and **no change was made**: `ANALYZE` on an emptied table *deletes* its `sqlite_stat1` rows rather than recording zeros, so the planner falls back to its built-in defaults. `activities` — the table the new index is on — is not emptied by the reset and gets accurate statistics; confirmed present after migration.


afadil:
Good catch from Codex — fixed in 96326cf77.

The parse-failure diagnostics I added interpolated the raw column value, and since the shared closure covers `remaining_quantity`, `original_quantity`, `remaining_cost_basis`, `cost_per_unit`, `fee_allocated` and `tax_allocated`, that put financial data in the logs — against the `AGENTS.md` rule. All three warnings (including the adjacent `open_date` and `split_ratio` ones) now name only the lot id and the offending column, which is enough to locate the row without emitting any stored value.

afadil:
Also fixed the same class of leak in `check_lot_quantity_consistency` (756f68946), which is pre-existing rather than introduced here — it interpolated the extracted lot total and the position quantity into its CRITICAL error.

Rather than dropping both numbers and leaving a mismatch that says nothing, it now logs the account, the asset and the failure *class*: an empty lot book (points at extraction or seeding) versus a non-empty disagreement (points at drift between the two representations). That is the distinction worth acting on, and neither case needs the amounts.

PR #1399, opened 2026-08-04
feat: add configurable server URL for self-hosting

smitpatel316:
## Summary
Currently the mobile/desktop Tauri app forces users to use Wealthfolio's hosted server (https://api.wealthfolio.app) and there is no way to point to a self-hosted server. This is a limitation for self-hosters running apps/server.

This PR adds a configurable Server URL setting so users can point the app to their own self-hosted instance.

## Changes
- **Core**: Add server_url field to Settings model (Settings + SettingsUpdate) with empty default = use hosted server
- **Storage**: Handle server_url key in SettingsRepository (get/update)
- **Tauri service** (connect_service.rs):
  - Add global cache for custom URL set via UI
  - cloud_api_base_url() now checks in order:
    1. Custom URL from settings UI (self-hosting)
    2. Runtime env vars CONNECT_API_URL / WEALTHFOLIO_SERVER_URL (Docker/env override)
    3. Compile-time CONNECT_API_URL (backwards compat)
    4. Default https://api.wealthfolio.app
  - Add set_custom_server_url() / get_custom_server_url() helpers
  - Auth URL and publishable key also check runtime env var first
- **Context**: Initialize custom URL cache from settings at startup
- **Settings command**: Update cache when server_url is changed via UI
- **Server**: Support WEALTHFOLIO_SERVER_URL alias in apps/server/src/features.rs
- **Frontend**:
  - Add serverUrl to Settings TS type and settings provider
  - New component ServerUrlSettings in General settings: input with validation, save/reset, current effective URL display
  - Included in GeneralSettingsPage

## UI
Field appears under Settings -> General -> Server URL
- Empty = use default hosted server
- Valid URL (http/https) = use self-hosted server
- Shows current effective URL and restart note

## Minimal & Safe
- Only necessary changes, no breaking changes
- Empty string = existing behavior preserved
- Validation ensures http/https scheme
- Backwards compatible with existing env-var flows

## Testing
- Manual: Set custom URL via UI, verify it persists and cloud_api_base_url() returns it
- Manual: Reset to empty, verify it falls back to default
- Manual: Env var override still works
- Should be covered by existing settings tests

Closes self-hosting limitation for mobile app.

### DIFF
--- a/apps/frontend/src/lib/settings-provider.tsx
+++ b/apps/frontend/src/lib/settings-provider.tsx
@@ -21,6 +21,7 @@ interface ExtendedSettingsContextType extends SettingsContextType {
         | "onboardingCompleted"
         | "menuBarVisible"
         | "syncEnabled"
+        | "serverUrl"
       >
     >,
   ) => Promise<void>;
@@ -55,6 +56,7 @@ export function SettingsProvider({ children }: { children: ReactNode }) {
         | "onboardingCompleted"
         | "menuBarVisible"
         | "syncEnabled"
+        | "serverUrl"
       >
     >,
   ) => {
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -858,6 +858,7 @@ export interface Settings {
   autoUpdateCheckEnabled: boolean;
   menuBarVisible: boolean;
   syncEnabled: boolean;
+  serverUrl: string;
 }
 
 export interface SettingsContextType {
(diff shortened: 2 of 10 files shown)


rsnodgrass:
@smitpatel316 thank you!

afadil:
Thanks for the contribution. I’m closing this PR because the proposed approach is technically incorrect and does not solve the requested issue.

Today, each Wealthfolio native app is self-contained, with its own embedded backend and local database.

`https://api.wealthfolio.app` is the API for Wealthfolio Connect, a separate optional service that provides broker and device synchronization between independent installations. The self-hosted `apps/server` runtime is a different system, with different endpoints, responsibilities, and authentication.

This PR redirects the Connect client to `apps/server`. That does not make the native app use the self-hosted backend: normal Tauri commands would continue using the app’s local database, while Connect and device-sync requests would be sent to endpoints that `apps/server` does not implement.

The PR also does not currently compile because `set_custom_server_url` is called through `crate::services` without being exported.

Supporting native apps backed directly by a self-hosted server requires a separate architectural design covering the remote API adapter, authentication, offline behavior, and local-versus-remote data ownership. This should be analyzed and implemented as a dedicated feature rather than as a Connect URL override.

For self-hosted users who want mobile access today, Wealthfolio can be installed as a PWA from the browser and added to the phone’s home screen. Because it uses the same frontend as the native app, it provides largely the same core experience while connecting directly to the self-hosted server.

Thanks again for exploring this use case.

PR #1424, opened 2026-08-09
fix(connect,core): keep one asset identity per instrument (NEO MIC, share-class suffixes)

rwgs:
## Description

A broker-connected instrument could be filed under two different asset identities —
once from the activity path and once from the holdings path — so transaction-derived
holdings and broker-reported positions could never reconcile, and the instrument was
double-counted across the pair.

Two independent causes, both found against a real brokerage connection:

**1. Exchange MIC disagreed between the two paths.** Activities took the MIC straight
from the provider's exchange object, preferring `mic_code`. Holdings read the suffix
off the symbol instead. On NEO Exchange these disagree: the provider sends
`mic_code: "NEOE"`, which is not an ISO MIC, while the `.NE` suffix and `code: "NEO"`
both point at `XNEO` (Cboe Canada) correctly.

**2. A share-class suffix was mistaken for a venue suffix.** `parse_symbol_with_exchange_suffix`
cannot tell an exchange suffix from a ticker that merely ends in one — the symbol alone
carries no venue to check against. `.F` is Frankfurt in the Yahoo table, but BMO's
`ZAAA.F` on Cboe Canada is the *currency-hedged unit class*. Stripping it files the
holding under the unhedged listing — a different security that Yahoo carries separately:

```
ZAAA.NE    BMO AAA CLO ETF                 quoted 30.62  <- what we stored
ZAAA-F.NE  BMO AAA CLO ETF Hedged          quoted 29.48  <- what was traded
ZBBZ.NE    BMO BBB CLO ETF - CAD UNITS     quoted 29.96  <- what we stored
ZBBZ-F.NE  BMO BBB CLO ETF (Hedged Units)  quoted 29.05  <- what was traded
```

Out by 3.7% and 3.1%, with nothing able to flag it: the shortened ticker resolves a
perfectly valid quote, for the wrong fund. The provider is blameless — its description
is "BMO AAA CLO ETF" for both, so the ticker is the only field separating them. 33
activity records in one real account were affected (26 `ZAAA.F`, 7 `ZBBZ.F`).

### What changed

- **Both paths resolve identity from the API `symbol` first**, falling back to
  `raw_symbol`. The API symbol is the decorated form, so stripping one known suffix
  leaves the ticker intact (`ZAAA.F.NE` → `ZAAA.F`). The `raw_symbol` fallback is kept
  and still pinned by the existing `VOD.L` test.
- **New `parse_symbol_with_known_exchange`** in `crates/core/src/assets/asset_id.rs`:
  a suffix that names a market the instrument cannot be trading on was never a venue
  marker, so it stays on the ticker. Two guards keep the rule from over-reaching,
  because a merely *different* MIC is not evidence of a share class:
  - **The known venue must be a MIC the registry recognises.** Brokers send codes that
    are not MICs — this connection calls Cboe Canada `NEOE` — and `sync_holdings` passes
    a broker's raw exchange `code` through when the symbol carries no suffix. Junk must
    not silence a good suffix.
  - **The two venues must trade in different currencies.** `XETR` (`.DE`) and `XFRA`
    (`.F`) are two registry entries for the same German listing, so a caller naming one
    while the symbol carries the other is choosing a venue, not spelling a share class.
    Keeping `.DE` there would store `BMW.DE` on XFRA and resolve the provider symbol
    `BMW-DE.F`, which does not exist. A different currency is a different market, and
    `.F` on a CAD venue cannot be Frankfurt.

  Behaviour is otherwise unchanged: with no known venue, with a venue the suffix agrees
  with, and for a share class that is not a known venue suffix at all (`BRK.B`).
- The rule is applied at **all seven sites that decide a stored ticker**: identity
  canonicalization, import resolution, the activity save and update paths, the bulk
  import/sync asset-spec builder, the import preview check, and the broker mapper's
  symbol fallback. That count is load-bearing — fixing only the first two changed
  nothing observable, because for a synced activity the decision that reaches an asset
  row is made in the save path.

Symbols with no suffix — the majority — keep exactly their present behaviour.

### Migration

No migration ships with this. An installation that already synced a `.F` holding keeps
its `ZAAA` row and would file new activity under `ZAAA.F`, splitting that history.
Flagging it for maintainer judgement rather than deciding it here.

For the MIC half, existing **activity** rows self-heal: the bulk upsert matches on
provider identity (`source_system`, `account_id`, `source_record_id`) before falling
back to the semantic idempotency key, so a re-sync repoints them in place rather than
inserting duplicates. The **asset** row does not — it is keyed by instrument identity,
so a re-sync files the instrument under `XNEO` and leaves the old `NEOE` asset row
behind with whatever quote history it accumulated. Harmless but untidy, and the same
maintainer call as above.

### Testing

- `cargo test --workspace` — **2865 passed, 0 failed, 16 ignored** across 37 suites.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets
  --all-features -- -D warnings` clean.
- **18 new tests.** The broker paths: NEO-listed ETF → `XNEO`, suffixless symbol still
  takes the broker's MIC, `ZAAA.F.NE` → `XNEO` with the share class retained, `BRK.B`
  surviving in both fields. The identity rule directly: no venue, agreeing venue,
  different-currency venue, sibling venue (`BMW.DE` + `XFRA` → `BMW`), a venue that is
  not a MIC (`NEOE`, `NASDAQ`), unsuffixed symbols. And the rule through the paths that
  store identity: canonicalization, import resolution, and the activity save path.
- **The save-path test is a real guard, not a snapshot.** Reverting just that one call
  site back to `parse_symbol_with_exchange_suffix` makes
  `test_create_keeps_a_share_class_the_supplied_venue_contradicts` bind the activity to
  `SEC:ZAAA:XNEO` — the unhedged fund — instead of `SEC:ZAAA.F:XNEO`.
- Verified live against a real brokerage connection: after the change the store holds
  `ZAAA.F`/`ZBBZ.F` on `XNEO`, named "BMO AAA CLO ETF Hedged" and "BMO BBB CLO ETF
  (Hedged Units)" and quoted from those listings. The names come from the market-data
  provider, so they confirm which security was matched. That sync predates the two
  guards above, which only narrow the rule and cannot change this case — `XNEO` trades
  in CAD and `XFRA` in EUR, so the contradiction still holds.
- A round-trip audit asserting the stored ticker is the provider's ticker with exactly
  its venue suffix removed failed 2 of 63 assets before this change and passes 63 of 63
  after — so the guard is proven against the defect, not only after it. (That audit
  tooling is local and not included in this PR.)
- Not run: Playwright e2e.

### Known and not addressed

**Search still strips unqualified.** `local_search_identity` in
`crates/core/src/quotes/service.rs` and the agent-tools taxonomy lookup resolve a query
with no venue to check against, so a local search for `ZAAA.F` strips to `ZAAA` and will
not match the asset now stored as `ZAAA.F` on `XNEO`. That is a search-surface gap rather
than an identity one — nothing mis-files — but it wants its own change, probably by
trying the unstripped query as an additional candidate.

**Sibling-venue disagreements are resolved in favour of the suffix.** The currency guard
above draws the line where I could defend it: same currency means same market, so the
suffix wins and the ticker is stripped as before. If a maintainer would rather the
explicit MIC always win in that case, it is a one-line change to the guard, and the
`BMW.DE` + `XFRA` test says which way it currently goes.

## Checklist


afadil (changes requested):
@rwgs Thanks for the detailed fix. The issue is real.
a few changes I think are needed before merging:

- **NEOE** is the official ISO MIC for **Cboe Canada**. could you please update the `.NE` mapping to `NEOE` instead of treating it as invalid or converting it to `XNEO`.

- Activities and holdings should use the same normalization function with `symbol`, `raw_symbol`, and exchange metadata. to prevent this kind of situation in the futur.
- also if you can add a test for `raw_symbol: "ZAAA.F"` with no decorated symbol. It currently resolves to Frankfurt (`XFRA`) instead of Cboe Canada.

thanks again

rwgs:
@afadil committed requested changes.

PR #1428, opened 2026-08-10
fix(dashboard): count hidden accounts in group stats, keep groups of one

### LINKED ISSUE #1340: Dashboard group stats exclude hidden (not archived) accounts
The dashboard's account list builds its performance scopes from a list that filters out hidden accounts. This contradicts docs and expected behavior.

A related secondary problem is that an account group disappears if there is only one visible member which makes the stats not representative.

## Expected behavior

Per the [accounts docs](https://wealthfolio.app/docs/guide/accounts/#account-status), a **hidden** account is removed from account lists but is still counted in Total and history. Only **archived** accounts aren't a part of calculations.

It looks like the dashboard header total correctly includes hidden accounts. The account rows below it do not, so the same screen reports all-time performance over two different data sets.

## Steps to reproduce

1. Create a group with two or more accounts, at least one of which has closed activity history (contributes gain but is now at zero value).
2. Note the group row's total gain and return on the dashboard.
3. Hide (do not archive) one of the accounts in that group.
4. The group row's gain and return change, despite no change in current value.
5. Hide members until one remains visible: the group row disappears entirely, replaced by a single account row, and group-wide stats are no longer shown.

The header stats seem to stay steady throughout, which maps to the documented behavior.

--- 
**The following sections are generated by Claude Opus**. I can submit a patch under the CLA that implements the fixes, if needed.

--- 

## Actual behavior

`useAccounts` defaults to `filterActive: true` and drops hidden accounts client-side (`apps/frontend/src/hooks/use-accounts.ts`). The dashboard's accounts section calls it without overriding that default (`apps/frontend/src/pages/dashboard/accounts-summary.tsx`), then derives its performance scopes from the filtered result. One list drives both what is displayed and what is calculated, so "hidden from lists" silently becomes "removed from the group's performance scope".

A closed account at zero current value still contributes a full gain history and still determines when the group's compounding clock starts, so hiding one moves both the group's all-time gain and its TWR. Different hidden subsets produce different group numbers.

Separately, the group-vs-account decision is made on the count of _visible_ members, in two independently duplicated places — once when building scopes and once when rendering. A group with one visible member is flattened into a plain account row, which removes the group-wide value and return entirely.

## Suggested fix

Frontend-only, contained to `apps/frontend/src/pages/dashboard/accounts-summary.tsx`:

- Fetch with `filterActive: false` **at this call site only**. The hook default should stay `true` — 15+ other screens (account pickers, import flows, holdings) rely on it, and surfacing hidden accounts there would be wrong.
- Carry `isActive` onto the display data so group math and row rendering can diverge: hidden members feed both the group's performance scope and its value sum, but never render a row.
- Include hidden members in the group's value sum as well as its scope, so value and gain are computed over the same set of accounts.
- Drop the single-visible-member collapse at both the scope-building and rendering sites. Render a group row whenever the group has at least one visible member.

Two follow-on details worth deciding as product questions:

- A group's "N accounts" subtext: visible count is the natural choice, since it describes what the user can see rather than what is being summed.
- The dashboard empty state keys off the combined account list, which becomes non-empty when every account is hidden. It should key off whether any account is _visible_, or an all-hidden portfolio renders a blank list instead of the empty state.

## Known trade-off

With this fix, an expanded group can show a header total that its visible child rows do not sum to, because hidden members contribute to the header but render no row. This is correct per the docs, but it is a real presentation wrinkle and may deserve a hint in the UI — worth a maintainer's opinion.

## Notes

- Existing dashboard tests mock `useAccounts` wholesale, so no current test exercises `isActive`; hidden-account behavior is uncovered.
- The existing test asserting the grouped scope array encodes the single-account-collapse behavior and needs updating alongside this change.
- Unrelated pre-existing quirk noticed nearby: an account whose group is _literally_ named "Uncategorized" is treated as ungrouped, in both the scope and render paths.

kryofin:
## Description

Fixes #1340

Hidden accounts (`isActive: false`, not archived) are documented as still counting toward
totals and history, and the backend honors that everywhere — `{type: "all"}` scope
resolution, snapshot recalculation, and `account_in_portfolio_scope` all exclude only
archived accounts. The dashboard's account summary did not: `useAccounts` defaults to
`filterActive: true`, and both the performance scopes and the group value sums were built
from that filtered list.

Consequences:

- Hiding an account silently removed its entire gain history from its group's TWR/PnL,
  even at zero current value, and shifted the group's TWR start date.
- The group rows disagreed with the dashboard headline (which is hidden-inclusive), by
  triple-digit percentage points on all-time in real data.
- Hiding all but one member dissolved the group entirely: the `length === 1` collapse saw
  only the filtered list, so the group and its stats vanished.

Short windows (YTD/3M) mask all of this because hidden accounts are usually dormant —
the divergence concentrates in all-time and long ranges.

## Fix

Two commits:

**`fix(dashboard): count hidden accounts in group stats`**

- Fetch with `filterActive: false` at this one call site. The hook default is left
  alone — account pickers and import flows still want active-only.
- Carry `isActive` onto the display data.
- Include hidden members in both the group's performance scope and the group's value sum,
  so value and gain are computed over the same account set — and match the headline.
- Measure group membership over all non-archived accounts, so hiding members no longer
  dissolves a group.
- Only active accounts get a rendered row; per-account scopes on expand are visible-only.

**`fix(dashboard): render single-member groups as groups`**

- Drop the `length === 1` collapse at both the scope and render sites, so a genuine
  group-of-one renders with its group header instead of a bare account row. The numbers
  are identical either way (a one-account group scope has the same scope key and value as
  the account itself); this is presentation only, split into its own commit so it can be
  dropped independently.

## Behavior note

A group's header value includes hidden members' current value, so expanded visible rows
may not sum to the header. This is intentional: the alternative (visible-only value,
hidden-inclusive gain) would compute the two numbers over different sets, and the header
would disagree with the portfolio total above it.

## Test plan

- `accounts-summary.test.tsx`: new tests — hidden accounts kept in group scope and totals
  without rendered rows (collapsed and expanded), group survives hiding all but one
  member, single-member group renders as a group, and the component actually opts out of
  the hook's active-only filter (the hook is mocked, so this pins the
  `filterActive: false` argument).
- Verified against a live dataset over HTTP: group scopes with hidden members now match
  the hidden-inclusive headline scope; YTD/3M unchanged.
- `pnpm type-check` clean.

## Checklist

### DIFF
--- a/apps/frontend/src/pages/dashboard/accounts-summary.test.tsx
+++ b/apps/frontend/src/pages/dashboard/accounts-summary.test.tsx
@@ -429,6 +429,62 @@ describe("AccountsSummary", () => {
     ]);
   });
 
+  it("counts hidden accounts in group totals without rendering them", async () => {
+    const user = userEvent.setup();
+
+    renderAccountsSummary({
+      accounts: [
+        createAccount({ id: "visible", name: "Visible Account", group: "Brokerage" }),
+        createAccount({
+          id: "hidden",
+          name: "Hidden Account",
+          group: "Brokerage",
+          isActive: false,
+        }),
+      ],
+      valuations: [
+        createValuation({ accountId: "visible", totalValue: 100, totalValueBase: 100 }),
+        createValuation({ accountId: "hidden", totalValue: 25, totalValueBase: 25 }),
+      ],
+      performanceByScopeKey: {
+        "accounts:hidden,visible": { pnl: 40, returnValue: 0.4 },
+      },
+    });
+
+    // The component must opt out of the hook's default active-only filtering, or hidden
+    // accounts never reach the grouping logic at all.
+    expect(mockUseAccounts).toHaveBeenCalledWith(expect.objectContaining({ filterActive: false }));
+
+    // Hidden accounts stay in the group's scope, so their gain history is not dropped.
+    expect(getLastPerformanceScopes()).toContainEqual({ accountIds: ["visible", "hidden"] });
+
+    // Group value spans both members: 100 visible + 25 hidden. Only the group row can show
+    // this total, since the sole visible account is worth 100.
+    expect(screen.getByText("Brokerage")).toBeInTheDocument();
+    expect(screen.getByText("value:USD:125")).toBeInTheDocument();
+
+    // But the hidden account never appears as a row, expanded or not.
+    expect(screen.queryByText("Hidden Account")).not.toBeInTheDocument();
+    await user.click(screen.getByText("Brokerage"));
+    expect(screen.queryByText("Hidden Account")).not.toBeInTheDocument();
+    expect(screen.getByText("Visible Account")).toBeInTheDocument();
+  });
+
+  it("keeps a group visible when all but one member is hidden", () => {
+    renderAccountsSummary({
+      accounts: [
+        createAccount({ id: "left", name: "Still Visible", group: "Brokerage" }),
+        createAccount({ id: "gone", name: "Now Hidden", group: "Brokerage", isActive: false }),
+      ],
+      valuations: [
+        createValuation({ accountId: "left", totalValue: 100 }),
+        createValuation({ accountId: "gone", totalValue: 0 }),
+      ],
+    });
+
+    expect(screen.getByText("Brokerage")).toBeInTheDocument();
+  });
+
   it("keeps standalone account values visible while performance is loading", () => {
     renderAccountsSummary({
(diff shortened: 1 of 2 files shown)


kryofin:
@afadil Any feedback on this PR? It's been open for a bit and it makes a big impact on the correctness of the numbers presented in the main dashboard.

kryofin:
@afadil Ping 😀

PR #1429, opened 2026-08-10
feat(spending): add amount conditions to categorization rules

### LINKED ISSUE #1427: FEATURE: Support spending rules matching on amount
Spending rules currently match on the transaction name (plus an optional
activity type and account scope). That works until two transactions from the
same payee mean different things at different sizes — and then there's no way to
tell them apart.

A few examples that come up quickly with real data:

- Third-party payment services (ACH, Apple Pay) don't always send along
  accurate payee information but regular charges can filter based on amounts.
- `TRANSFER TO SAVINGS` for $50 is pocket money; for $5,000 it's a real savings
  contribution that deserves its own category.
- A recurring $49.99 charge from a payee is my gym membership; a one-off $200
  charge from the same payee is something else.
- Tiny card transactions under a few dollars are noise I'd like to sweep into a
  catch-all without dragging larger purchases along.

## Proposal

Add an optional amount condition to rules: an operator (`exactly`,
`greater than`, `greater than or equal`, `less than`, `less than or equal`,
`between`) plus a value — two values for `between`, inclusive on both ends. The
condition is AND-ed with the existing pattern/type/scope checks.

In the editor it's one extra row: an "Amount (optional)" selector that defaults
to **Any amount**, showing a single value field for most operators and From/To
fields only when you pick Between. The rules list spells the condition out in
words ("Amount greater than 1,000") so a rule's behavior is obvious at a glance.

Related: #1350

kryofin:
Closes https://github.com/wealthfolio/wealthfolio/issues/1427

Rules can now optionally match on the transaction amount, alongside the existing
name pattern, activity type, and account scope. This makes it possible to split
same-payee transactions by size, critical for certain uses cases described in 
#1427. Similar functionality is a critical part of other projects such as
Actual Budget.

## How it works

A rule stores an operator (`exactly`, `greater than`, `greater than or equal`,
`less than`, `less than or equal`, `between`) and a value — two for `between`,
inclusive on both ends. The condition is AND-ed with the other gates and
compares against the activity's **unsigned** amount; direction stays the job of
the activity-type filter. That's a deliberate departure from Actual Budget's
design: no "amount (inflow)/(outflow)" pseudo-fields (unnecessary here since
amounts aren't signed) and no approximate matching (a hidden tolerance causes
more confusion than it saves).

In the editor, it's one extra row — an "Amount (optional)" selector defaulting
to **Any amount**. Most operators show a single value field; Between swaps in
From/To. The rules list spells the condition out in words ("Amount greater than
1,000") so behavior is obvious at a glance.

Semantics pinned down explicitly:

- Activities with no amount never match a rule that has an amount condition.
- Zero compares like any other value.
- Swapped Between bounds are rejected in the form (and again server-side) rather
  than silently reordered.
- The server normalizes: no operator ⇒ no stored values; a non-between operator
  drops a stale upper value.

## Changes

- **Migration** — three nullable columns on `spending_categorization_rules`
  (`amount_op` with a CHECK, `amount_value`, `amount_value2`; Decimal-as-TEXT
  like every other money column).
- **Matcher** — per-operator amount gate in `match_compiled`, evaluated with the
  account/type gates before the text match.
- **Service** — validation matrix (value required, non-negative, from ≤ to) plus
  the normalization above, on both create and update; update supports explicitly
  clearing the condition.
- **Frontend** — amount section in the rule form with inline validation, worded
  condition chip in the rules list, i18n for en/de/es/fr/zh.
- Preset rules are unchanged. Device sync carries the new fields; **older app
  versions will silently drop them when syncing**, which seems worth a
  release-note line.

## Screenshots

|                              |                                     |
| ---------------------------- | ----------------------------------- |
| Rules list with amount chips | 01-rules-list-amount-chips  |
| Operator options             |  02-amount-operator-options |
| Single-value operator        |  03-amount-greater-than |
| Swapped-bounds validation    | 04-between-validation-error |
| Between with valid bounds    | 05-amount-between |

## Testing

- Rust: new matcher tests cover every operator's boundary from both sides
  (strict vs. inclusive, Between's ends, zero, missing amounts, unsigned
  comparison of negative inputs, interaction with priority and the other gates);
  service tests cover the validation matrix, normalization, and a rerun
  integration case; storage tests cover round-trips and the clear-condition
  patch. `cargo test -p wealthfolio-spending -p wealthfolio-storage-sqlite` —
  359 passing.
- Frontend: vitest suites for the zod schema (positive/negative per operator),
  payload mapping (including explicit-null clearing on update), and chip
  rendering. Full suite passes (1,145 tests), plus `type-check` and `lint`.
- Manual: built the server image from this branch, seeded activities, and walked
  the full flow in the browser — create rules with conditions, trigger the
  validation error, re-run rules (only in-range activities categorized), edit a
  rule back to "Any amount", and confirm a re-run then picks up the previously
  excluded activity.

## Checklist


kryofin:
@afadil This is the next PR for improving spending rule classification (with screenshots). The one nuance to point out is that this only deals with absolute amounts in the rules classification and we don't have access to activity type. This might limit the rules that can be set up but I believe this change should capture a majority of the use cases. We can always iterate on more functionality later as we get more user feedback.

Please also do look at the DB migration code. As I am mostly doing fresh imports, fixing gaps, and then redoing the import, I haven't experienced too many DB migrations with a long-running system.

kryofin:
@afadil Any chance you might be able to look at this and give feedback? I have a few more spending improvements/paper cuts I want to look at after this.

kryofin:
@afadil Thank you for catching the serde issue! 😬

PR #1437, opened 2026-08-13
Fix TWR after pre-funding negative balance

### LINKED ISSUE #1425: A one-day negative cash balance on the first activity day nulls TWR for the entire period
## Version

3.6.3 (self-hosted, Docker `afadil/wealthfolio:3.6.3`). This is a regression against 3.6.1
for one of my accounts.

## Summary

In a T+2 settlement market (Taiwan), a trade executed on day D is only paid for on D+2, so
it is normal and correct for the cash deposit covering it to be recorded *after* the trade
date. Wealthfolio books cash on the trade date (`settlement_date` is never populated by my
importer), so the account's first valuation day shows a small negative cash balance.

3.6.3 treats that as a fatal condition and returns `twr: null` and `valueReturn: null` for
**the whole period**, so the account shows no return at all. 3.6.1 returned a normal figure
for the same data.

## The data

Account opened 2026-06-24. Full activity list for the first two days:

| Date | Type | Asset | Amount | Fee |
|---|---|---|---:|---:|
| 2026-06-24 | BUY ×10 | a warrant, 1000 units @ 0.01 | 10 each | 1 each |
| 2026-06-25 | BUY ×5 | an ETF | ~38,000 each | 15 each |
| 2026-06-25 | DEPOSIT | — | 1,000,000 | — |

Total outlay on 2026-06-24 is **110** in account currency (about USD 3.40). Settlement for
those trades falls on 2026-06-26, one day *after* the deposit, so nothing is actually
overdrawn in reality.

Resulting valuation rows:

| valuation_date | total_value | cash_balance | external_flow_source |
|---|---:|---:|---|
| 2026-06-24 | **-110** | **-110** | NO_FLOW |
| 2026-06-25 | 999,525 | 808,625 | CASH_AMOUNT |
| 2026-06-26 | 968,216 | 264,016 | NO_FLOW |

## Behaviour

`POST /api/v1/performance/summary` with `itemType: "account"`, any period covering the
account's history:

**3.6.1**

```
period.startDate = 2026-06-25
twr             = -0.1356443    (percentStatus: complete)
amount          = -456,934
```

**3.6.3**

```
period.startDate = 2026-06-24
twr             = null          (percentStatus: unavailable)
valueReturn     = null
amount          = -457,299      (amountStatus: complete)

notApplicableReasons:
  - "TWR unavailable for 2026-06-25 because portfolio value is negative.
     Review the underlying transactions, prices, and cash balances."
  - "Value return unavailable for transaction-mode scope because starting value
     is zero or negative."
```

The message names 2026-06-25 because that day's return is computed against the previous
day's closing value, which is the -110 row.

Two things changed together: 3.6.3's rebuild now emits a valuation row for 2026-06-24 (the
first activity date) where 3.6.1's timeline began at 2026-06-25, and the negative value on
that row is classified as fatal rather than skippable.

Restricting the query to a later window works fine — the same account returns -11.04% for
2026-07-10 → 2026-08-09 on 3.6.3.

## Expected

A transient sub-unit negative balance on the first day, before any external flow has been
recorded, should not void the entire period. The 3.6.3 note for #1218 says dormant days
with a sub-unit denominator now pause compounding instead of nulling the result; the same
treatment seems right here — either skip days before the first day with a positive value
when establishing the chain start, or pause compounding across them.

For scale: 110 out of a 1,000,000 account (0.011%) currently removes the return figure for
a period in which the account traded normally for six weeks.

## Environment

- Wealthfolio 3.6.3, Docker, SQLite backend, base currency TWD
- Activities imported via API from a Taiwanese broker; `settlement_date` is null on all rows
- Verified by restoring the same pre-upgrade database into a 3.6.1 container and a 3.6.3
  container side by side, with a forced full recalculation on both

Neallin-917:
## Summary

Fixes #1425.

- Skip a negative opening window only before the return chain starts, when it has no valid return base.
- Preserve the existing fatal result for negative values after compounding begins or when the opening value is positive.
- Add a regression test covering a trade-date debit followed by the deposit that funds it.

## Validation

- `cargo fmt --check`
- `cargo test -p wealthfolio-core`
- `cargo clippy -p wealthfolio-core -- -D warnings`
- `git diff --check`

## Checklist

- [x] I have read and agree to the Contributor License Agreement (CLA).
- [x] I have added tests that cover the change.

### DIFF
--- a/crates/core/src/portfolio/performance/performance_service.rs
+++ b/crates/core/src/portfolio/performance/performance_service.rs
@@ -615,7 +615,15 @@ impl PerformanceService {
                 warned_partial_value_coverage = true;
             }
 
-            if prev_value.is_sign_negative() || curr_value.is_sign_negative() {
+            // Before the return chain starts, a negative opening balance can be
+            // historical pre-funding bookkeeping (for example, a trade-date
+            // debit that is covered by a later settlement deposit). It has no
+            // valid return base, so skip that window and let the first funded
+            // window establish the chain. Once the opening value is positive
+            // or compounding has started, a negative value remains fatal.
+            if (prev_value.is_sign_negative() || curr_value.is_sign_negative())
+                && (chain_started || prev_value > Decimal::ZERO)
+            {
                 not_applicable_reasons.push(format!(
                     "TWR unavailable for {} because portfolio value is negative. Review the underlying transactions, prices, and cash balances.",
                     curr_point.valuation_date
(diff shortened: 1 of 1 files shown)


afadil:
fixed in #1443

PR #1445, opened 2026-08-13
fix(allocation): keep the rebalance planner reachable when allocations are in-band

### LINKED ISSUE #1444: [BUG] Cash-flow rebalance planner is unreachable while all categories are inside the tolerance band
**Describe the problem**

The only path to the rebalance planner is the "Review rebalance →" button on the "Largest gaps" card, and that button only renders when at least one category is outside its drift tolerance band (`isOutOfBand()` gating in `drift-drivers-card.tsx`). Drift is computed on invested value only — cash is not part of the target base — so depositing fresh cash never moves a category out of band.

The result: after a deposit, the card says "All categories inside target. No action required." and there is no way to reach the cash-flow-only planner — precisely the moment it is most useful ("I added cash, how should I split the buys to stay on target?").

**Steps to reproduce**

1. Create an allocation target with weights and a tolerance band; hold positions that sit inside the band.
2. Deposit cash into the account — even a substantial deposit leaves the displayed drift unchanged, since cash is not in the target base.
3. Open the target's detail view: "All categories inside target. No action required." — no "Review rebalance →" button anywhere, so the planner (including its cash-flow-only mode) cannot be reached.

**Expected**

The "Review rebalance →" entry is available whenever its parent view wires the rebalance action (the Calculate step already handles the in-band case fine and produces a cash-flow plan that deploys the idle cash toward the target weights).

**Proposed solution**

Render the "Review rebalance →" CTA whenever the parent wires the handler, independent of the out-of-band check, keeping the in-band status text as is. I have this running on my instance and will open a PR referencing this issue.

fmaass:
## Description

Closes #1444.

One-line change: the "Review rebalance →" CTA on the "Largest gaps" card now renders whenever the parent wires `onRebalanceClick`, instead of only when a category is out of band.

Rationale: drift is computed on invested value only, so idle cash never produces drift — after a deposit the planner was unreachable exactly when a cash-flow-only calculation is most useful. The planner itself already handles the in-band case correctly (it accepts the report and produces a plan that deploys available cash toward the target weights).

Nothing else changes: `isOutOfBand()`, the drift math, the suggested-moves list and the in-band status text ("All categories inside target. No action required.") behave exactly as before — the status text and the button now coexist. That coexistence is intentional rather than contradictory: the status describes tolerance-band drift, while the button also exposes cash-deployment planning, which is useful precisely when drift is zero but idle cash is not.

Tested on my self-hosted instance: with all categories in-band and idle cash on the account, the CTA renders, "Cash-flow only" mode calculates a plan whose buys sum exactly to the deployable cash, and with zero drift and zero cash the Calculate step stays disabled as before. `pnpm` frontend build, type-check and lint are clean.

## Checklist

- [x] I have read and agree to the [Contributor License Agreement](../blob/main/CLA.md)

### DIFF
--- a/apps/frontend/src/pages/allocation-targets/components/drift-drivers-card.tsx
+++ b/apps/frontend/src/pages/allocation-targets/components/drift-drivers-card.tsx
@@ -78,7 +78,7 @@ export function DriftDriversCard({
     .sort((a, b) => Math.abs(b.driftBps) - Math.abs(a.driftBps));
   const visibleRows = oobRows.slice(0, 3);
   const remainingRows = oobRows.slice(3);
-  const showRebalanceCta = oobRows.length > 0 && onRebalanceClick;
+  const showRebalanceCta = Boolean(onRebalanceClick);
 
   return (
     <Card className="flex h-full flex-col">


afadil:
Thank you!

PR #1446, opened 2026-08-13
feat(auth): auto-redirect to SSO when it is the only login method

### LINKED ISSUE #1343: [FEATURE REQUEST] Skip login landing page if only OIDC configured for self hosted web
I recently switched to OIDC authentication and noticed that there is now an additional screen (the login landing page. see screenshot below) that I have to click through before I get to the dashboard. i.e. the complete login flow is now 
User enter URL -> Wealthfolio login landing page -> idP login page -> Wealthfolio dashboard

Can we have a setting in the app that disables the login landing page if the installation only has OIDC configured so the login flow becomes
User enter URL -> idP login page -> Wealthfolio dashboard

Image

fmaass:
## Description

Closes #1343.

When the server reports `oidcEnabled: true` and `requiresPassword: false`, the login page currently renders a single "Sign in with SSO" button — a click with no decision behind it. With this change the page navigates straight to `/api/v1/auth/oidc/login` once auth status resolves, so a fresh visit lands in the app (or at the identity provider) with zero clicks.

A one-shot per-tab guard keeps the redirect from looping or hiding a failure:

- **One automatic attempt until a session is confirmed.** A `sessionStorage` guard (`wf.sso-redirect-guard`) is armed immediately before the automatic redirect and on logout, and cleared only once `/auth/me` confirms a session. While it is armed, automatic redirects are suppressed — a callback that returns without establishing a session (cookie unavailable, `/auth/me` failing) lands on the login page with the manual button instead of restarting OIDC.
- **Failed OIDC round-trips stay visible.** The redirect is also suppressed while `loginError` is set. The `?oidc_error=` query param is consumed into state and stripped via `history.replaceState` before auth status resolves, so the persisted state — not the URL — is the reliable signal.
- **The manual button always navigates.** A deliberate click on the SSO button is never blocked by the guard; it stays the recovery path after a failed automatic attempt.
- **Arming fails closed.** The guard is armed with a write + read-back; if `sessionStorage` is blocked or silently dropping writes, the automatic redirect is suppressed — the worst case in restricted browsers is exactly today's behavior, the interstitial with the button.

The automatic redirect uses `window.location.replace()` so the transient login page stays out of back-button history, and the button remains rendered as a fallback while the redirect happens.

Verification: the original submission was tested live on my self-hosted instance behind Authelia — a fresh visit with a valid IdP session reaches the dashboard with zero clicks (without one, the redirect lands on the IdP's login as expected), `/?oidc_error=access_denied` shows the error and does not redirect, and after logout the login page renders without redirecting while one SSO click signs back in. The guard rework on top is covered by the new `login-page.test.tsx` — the failed-session callback case (armed guard → no automatic redirect, manual button rendered), the arm-and-redirect happy path, manual navigation while the guard is armed, and error suppression — with type-check and lint clean.

## Checklist

- [x] I have read and agree to the [Contributor License Agreement](../blob/main/CLA.md)

### DIFF
--- a/apps/frontend/src/context/auth-context.tsx
+++ b/apps/frontend/src/context/auth-context.tsx
@@ -47,10 +47,22 @@ function resolveOidcError(t: TFunction, code: string): string {
 }
 
 /**
- * Set on logout so the login page offers the SSO button instead of bouncing the
- * user straight back into the IdP, which would make signing out impossible.
+ * One-shot per-tab guard for the automatic SSO redirect. Armed before every
+ * automatic redirect and on logout; cleared only once `/auth/me` confirms a
+ * session. While armed, the login page suppresses further automatic redirects
+ * (the manual SSO button is unaffected), so a callback that fails to establish
+ * a session — no cookie, `/auth/me` rejecting — lands on the login page once
+ * instead of looping through the IdP.
  */
-export const MANUAL_LOGIN_STORAGE_KEY = "wf.manual-login";
+export const SSO_REDIRECT_GUARD_STORAGE_KEY = "wf.sso-redirect-guard";
+
+function clearSsoRedirectGuard() {
+  try {
+    window.sessionStorage.removeItem(SSO_REDIRECT_GUARD_STORAGE_KEY);
+  } catch {
+    // noop – sessionStorage may be unavailable
+  }
+}
 
 const AuthContext = createContext<AuthContextValue | undefined>(undefined);
 
@@ -100,6 +112,7 @@ export function AuthProvider({ children }: { children: React.ReactNode }) {
               credentials: "same-origin",
             });
             if (meRes.ok && !cancelled) {
+              clearSsoRedirectGuard();
               setCookieSession(true);
             }
           } catch {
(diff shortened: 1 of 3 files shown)


afadil:
thanks for this PR. The overall approach looks good, but automatic SSO needs a one-shot per-tab redirect guard. If the callback succeeds but the session cookie is unavailable, or `/auth/me` fails, the page could go on infinite loop and repeatedly restart OIDC.

Please use a single `sessionStorage` guard:
- Set it before automatic SSO and on logout.
- Suppress subsequent automatic redirects while set.
- Let the manual SSO button navigate regardless.
- Clear it only after `/auth/me` confirms authentication.
- Prefer `window.location.replace()` for the automatic redirect.

 A focused test for the failed-session callback case would be helpful.

fmaass:
Thanks for the review — implemented in abb5f89, and the PR description is updated to match.

- A single one-shot per-tab guard (`wf.sso-redirect-guard`) is armed immediately before the automatic redirect and on logout, and cleared only once `/auth/me` confirms a session. While it is armed, automatic redirects are suppressed, so a callback that returns without establishing a session lands on the login page with the manual button instead of restarting OIDC.
- The manual button navigates regardless of the guard.
- The automatic redirect now uses `window.location.replace()`.
- Arming does a write + read-back, so an unusable `sessionStorage` fails closed to the manual button rather than risking a loop.
- Added `login-page.test.tsx` with the failed-session callback case (armed guard → no automatic redirect, manual button rendered) plus the arm-and-redirect happy path, manual navigation while the guard is armed, and error suppression.

One nuance to confirm: since the manual button navigates regardless, a manual click does not arm the guard — so if a manually started round-trip fails silently, the next page load fires one automatic redirect, which arms the guard and then settles on the manual button. That read of "navigate regardless" seemed closest to the intent, but I can arm on the manual click too if you'd rather keep it strictly one attempt per tab.

afadil:
Thanks for addressing the feedback. The current manual-button behavior is fine. Merging

PR #1458, opened 2026-08-16
fix(net-worth): stop dividing the Δ column by a non-baseline (#1456)

### LINKED ISSUE #1456: Net worth Δ column shows absurd percentages (or reports "0.0%" for new categories)
## Summary

On the Net Worth page with the **ALL** interval, the Δ column in the breakdown table renders percentages with six or more digits, while a category that grew from nothing renders as `0.0%`.

Both come from `deriveChange` in `apps/frontend/src/pages/net-worth/components/utils.ts`, which divides the range change by the value at the **first history point** with no guard on an immaterial or zero denominator.

## Steps to reproduce

1. Open the Net Worth page on a portfolio whose history starts many years before the current positions.
2. Select the **ALL** interval.
3. Look at the Δ ALL column in the breakdown table.

## Minimal reproduction

Any history shaped like the following triggers it. `GET /api/v1/net-worth/history` returns forward-filled daily points; the values below are illustrative:

```jsonc
// first point — small opening balance, no property yet
{ "date": "2015-01-05", "netWorth": 500.0,
  "breakdown": { "cash": 2e-8, "investments": 500.0 } }     // no "properties" key

// last point
{ "date": "2026-08-15", "netWorth": 1675000.0,
  "breakdown": { "cash": 25000.0, "investments": 1250000.0, "properties": 400000.0 } }
```

| Row                | first → last    | rendered               | problem                                                                    |
| ------------------ | --------------- | ---------------------- | -------------------------------------------------------------------------- |
| Investments        | 500 → 1,250,000 | `249,900%`             | arithmetically correct, unreadable                                         |
| Properties         | 0 → 400,000     | `0.0%`                 | **wrong** — reads as "no change" for a row that went from nothing to $400K  |
| Cash               | 2e-8 → 25,000   | `124,999,999,999,900%` | dividing by a float residue                                                |
| Net worth (footer) | 500 → 1,675,000 | `334,900%`             | same                                                                       |

The `cash` residue is not hypothetical: a rounding artifact in an early history point gets forward-filled for the life of the portfolio, so an effectively empty starting balance becomes the denominator.

## Root cause

`apps/frontend/src/pages/net-worth/components/utils.ts:144`

```ts
export function deriveChange(series: number[], isLiability: boolean): Change {
  if (series.length < 2) return { amount: 0, percent: 0 };
  const first = series[0];
  const last = series[series.length - 1];
  const amount = isLiability ? first - last : last - first;
  const base = Math.abs(first);
  return { amount, percent: base > 0 ? amount / base : 0 };
}
```

Two distinct defects:

1. **No materiality floor on `base`.** Any non-zero starting value is accepted, including float residue. Even a legitimate small baseline from a decade ago yields a number no one can read.
2. **`percent: 0` as the fallback for an undefined ratio.** When `base === 0` the percentage is undefined (or infinite), but `0` is returned — which the UI renders as `0.0%`, indistinguishable from a genuinely flat row.

The headline gain percentage has the same flaw independently, at `apps/frontend/src/pages/net-worth/net-worth-content.tsx:151`:

```ts
const base = first !== 0 ? Math.abs(first) : 1;   // falls back to dividing by 1
return { gainLossAmount: change, gainLossPercent: change / base };
```

## Expected

- A row whose starting value is zero or immaterial should not claim a percentage at all — and must never read as `0.0%`.
- Very large but genuine ratios should render in a form that fits the column and can be read at a glance.

## Proposed fix

Make `Change.percent` nullable and cap the display. Contained to `utils.ts` plus its two call sites; no layout change (the percent column is `w-12`, and both new forms fit).

```ts
const BASELINE_FLOOR = 1; // a sub-$1 starting balance is not a baseline

export interface Change {
  amount: number;
  percent: number | null; // null = no meaningful baseline
}

export function deriveChange(series: number[], isLiability: boolean): Change {
  // ...
  const base = Math.abs(first);
  return { amount, percent: base >= BASELINE_FLOOR ? amount / base : null };
}

export function formatChangePercent(percent: number | null, amount: number): string {
  if (percent === null) return amount > 0 ? "New" : "—";
  const abs = Math.abs(percent);
  if (abs >= 10) return `${formatCompactNumber(1 + abs)}×`; // 249,900% → "2.5k×"
  return formatPercent(abs, { digits: abs >= 1 ? 0 : 1, signDisplay: "never" });
}
```

Rendered result for the table above: `2.5k×`, `New`, `New`, `3.4k×`. Sign and color already derive from `change.amount`, so `ChangeCell` is otherwise untouched. The exact percentage can go in a `title` tooltip so it stays recoverable.

kryofin:
Fixes #1456

## What

Over the ALL range the Δ column showed percentages with six or more digits, and a category that grew from nothing showed `0.0%`. Both came from `deriveChange` dividing the range change by the first history point with no guard on the denominator.

`Change.percent` is now nullable. A starting value under one currency unit is treated as **no baseline** rather than a divisor, and `formatChangePercent` renders those rows as `New` (or `—` when the row shrank to nothing). Genuine but unreadable ratios past 10× collapse to a multiple.

## Before / after

Using the reproduction from #1456 — a history whose first point predates most positions:

| Row                                | before                 | after   |
| ---------------------------------- | ---------------------- | ------- |
| Investments (500 → 1,250,000)      | `249,900%`             | `2.5K×` |
| Properties (0 → 400,000)           | `0.0%`                 | `New`   |
| Cash (2e-8 → 25,000)               | `124,999,999,999,900%` | `New`   |
| Net worth header (500 → 1,675,000) | `334,900%`             | `3.4K×` |

Shorter ranges are unchanged: any row whose starting value is a real balance still renders a plain percentage, animated in the header exactly as before. Only ratios past 10× and rows with no baseline take the new paths.

## Design notes

- **Why `New` rather than `∞%` or a hidden cell.** `0.0%` was actively misleading — the point is to stop claiming a ratio that does not exist, while still telling the reader why. `—` covers the mirror case (a row that ended at nothing).
- **Why a multiple past 10×.** `2.5K×` fits the existing `w-12` percent column and is how people read growth at that scale. Nothing in the layout moved.
- **`BASELINE_FLOOR` is one unit of the display currency**, not a ratio-relative threshold. A ratio-relative rule would suppress percentages for legitimately explosive-but-small positions; an absolute floor only rejects balances that were never real.

## Tests

Four cases added to `components/utils.test.ts`: zero baseline, float-residue baseline, a genuine small baseline that must still produce a percentage, and the formatter's percent/multiple/`New`/`—` branches, plus `isPlainPercent` boundaries.

## Checklist

### DIFF
--- a/apps/frontend/src/i18n/locales/de/insights.json
+++ b/apps/frontend/src/i18n/locales/de/insights.json
@@ -92,6 +92,7 @@
       "category": "Kategorie",
       "value": "Wert",
       "delta_period": "Δ {{period}}",
+      "new": "Neu",
       "liabilities": "Verbindlichkeiten",
       "net_worth": "Nettovermögen"
     },
--- a/apps/frontend/src/i18n/locales/en/insights.json
+++ b/apps/frontend/src/i18n/locales/en/insights.json
@@ -92,6 +92,7 @@
       "category": "Category",
       "value": "Value",
       "delta_period": "Δ {{period}}",
+      "new": "New",
       "liabilities": "Liabilities",
       "net_worth": "Net Worth"
     },
--- a/apps/frontend/src/i18n/locales/es/insights.json
+++ b/apps/frontend/src/i18n/locales/es/insights.json
@@ -92,6 +92,7 @@
       "category": "Categoría",
       "value": "Valor",
       "delta_period": "Δ {{period}}",
+      "new": "Nuevo",
       "liabilities": "Pasivos",
       "net_worth": "Patrimonio neto"
     },
--- a/apps/frontend/src/i18n/locales/fr/insights.json
+++ b/apps/frontend/src/i18n/locales/fr/insights.json
@@ -92,6 +92,7 @@
       "category": "Catégorie",
       "value": "Valeur",
       "delta_period": "Δ {{period}}",
+      "new": "Nouveau",
       "liabilities": "Passifs",
       "net_worth": "Valeur nette"
     },
--- a/apps/frontend/src/i18n/locales/ja/insights.json
+++ b/apps/frontend/src/i18n/locales/ja/insights.json
@@ -91,6 +91,7 @@
       "category": "カテゴリ",
       "value": "評価額",
       "delta_period": "Δ {{period}}",
+      "new": "新規",
       "liabilities": "負債",
       "net_worth": "純資産"
     },
--- a/apps/frontend/src/i18n/locales/ko/insights.json
+++ b/apps/frontend/src/i18n/locales/ko/insights.json
@@ -91,6 +91,7 @@
       "category": "카테고리",
       "value": "가치",
       "delta_period": "Δ {{period}}",
+      "new": "신규",
       "liabilities": "부채",
       "net_worth": "순자산"
     },
--- a/apps/frontend/src/i18n/locales/zh/insights.json
+++ b/apps/frontend/src/i18n/locales/zh/insights.json
@@ -91,6 +91,7 @@
       "category": "类别",
       "value": "价值",
       "delta_period": "Δ {{period}}",
+      "new": "新增",
       "liabilities": "负债",
       "net_worth": "净资产"
     },
--- a/apps/frontend/src/pages/net-worth/components/breakdown-table.tsx
+++ b/apps/frontend/src/pages/net-worth/components/breakdown-table.tsx
@@ -29,6 +29,7 @@ const ROW_GRID =
 
 function ChangeCell({ change, currency }: { change: Change; currency: string }) {
   const formatting = useNumberFormatting();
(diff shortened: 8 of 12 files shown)


afadil:
@kryofin , Thanks for the fix and solid test coverage! One liability edge case remains: `deriveChange([0, 400_000], true)` returns `{ amount: -400_000, percent: null }`, so `formatChangePercent` renders `—` instead of `New`. This happens because liability amounts are sign-inverted.

Could we determine “New” from the raw first/last values (or store an explicit `isNew` flag) and add a liability test? Otherwise a newly added mortgage appears as `-$400K —` rather than `-$400K New`.

kryofin:
@afadil Done. Separate commit for easier review.

PR #1461, opened 2026-08-16
Fix net worth breakdown categories not being translated

### LINKED ISSUE #1453: [BUG] Net worth menu -> not translated categories
###  Description

Desktop application used in French, but the categories in the Net Worth menu have not been translated. 

The same issue occurs for the other languages, even though translations are available in the i18n files.

Image

###  Root cause

Separation of responsibilities between the backend and frontend.

The Rust backend returns objects with two distinct fields:

category — a stable machine key (“investments,” “properties,” “cash,” etc.)
name — a human-readable label in English, generated on the Rust side

In breakdown-table.tsx, the component was directly displaying item.name (the English string from the backend) without going through i18n

kryofin:
Fixes #1453

The backend returns each breakdown row with a stable machine key (`category`:
`"investments"`, `"preciousMetals"`, …) and an English display name generated in
`net_worth_service.rs` (`name`: `"Investments"`, `"Precious Metals"`, …). The
frontend rendered `name` directly, so category labels on the Net Worth page
stayed in English in every non-English locale.

## Fix

`breakdownLabel(t, entry)` in `pages/net-worth/components/utils.ts` resolves the
stable `category` key through a lookup map instead of trusting the backend's
English string. Applied in the two places that render category names:

- `breakdown-table.tsx` — asset rows, plus the name passed to the detail sheet
- `composition-bar.tsx` — stacked-bar segment tooltips

Rows carrying an `assetId` are individual **user-named** assets (liabilities,
category children), so their name passes through untranslated. Unmapped
categories fall back to the backend name rather than rendering a raw key.

## Translations

New `networth.categories.*` keys in the `insights` namespace — the one the
net-worth page already uses — for all seven locales (en, fr, de, es, zh, ja, ko),
with values matching the equivalent `holdings:group_*` labels so wording stays
consistent with the Holdings page.

## Not included

`pages/holdings/components/net-worth-widget.tsx` has the same bug, but it's
exported and never rendered anywhere in the app. Happy to fold in the fix if
preferred.

## Checklist

### DIFF
--- a/apps/frontend/src/i18n/locales/de/insights.json
+++ b/apps/frontend/src/i18n/locales/de/insights.json
@@ -93,6 +93,16 @@
       "liabilities": "Verbindlichkeiten",
       "net_worth": "Nettovermögen"
     },
+    "categories": {
+      "cash": "Bargeld",
+      "investments": "Investitionen",
+      "properties": "Immobilien",
+      "vehicles": "Fahrzeuge",
+      "collectibles": "Sammlerstücke",
+      "precious_metals": "Edelmetalle",
+      "other_assets": "Sonstige Vermögenswerte",
+      "liabilities": "Verbindlichkeiten"
+    },
     "category_sheet": {
       "over_period": "über {{period}}",
       "trend_period": "Trend · {{period}}",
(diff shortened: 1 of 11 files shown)


kryofin:
I believe there are more comprehensive fixes being worked on that would address this. Happy to reopen if not.

PR #1476, opened 2026-08-17
fix: convert retirement/goal money fields to the account's base currency

### LINKED ISSUE #1475: [BUG] Retirement planner (and other goal money fields) hardcode USD-scale amounts, ignoring the account's base currency
## Problem

Money fields on the Retirement Planner (and other `GoalLeverRow`-based goal fields) are built around fixed USD-scale numbers that never adapt to the account's base currency. For a currency like IDR (~15,000/USD), this makes the retirement planner effectively unusable — defaults and caps land in ranges that are meaningless in the selected currency.

Three compounding issues, all in the same money-input plumbing:

1. **Default plan values are raw USD numbers relabeled with the wrong currency.** `DEFAULT_RETIREMENT_PLAN` in `plan-adapter.ts` seeds `monthlyContribution: 1000` and expense presets (`Living: 3000`, `Healthcare: 300`). When a goal is created (`goal-new-page.tsx`), only the currency *label* is swapped to the account's base currency — the numbers aren't converted. Result: a new IDR plan shows "IDR 1,000/mo contribution" (≈ $0.065), silently seeding a plan that funds nothing.

2. **Typed input was clamped to a tiny USD-scale ceiling.** `GoalLeverRow` (`goal-lever-row.tsx`) capped manually-typed money values to the same ceiling used for the slider's practical drag range — topping out at 200,000 (raw units) for the living-expense field. Users on high-denomination currencies couldn't type a realistic value at all.

3. **The slider's own range had the same raw-unit ceiling**, independent of (1) and (2) — `sidebar-configurator.tsx` computed slider ranges via `sliderMaxFor(value, baseMax, increment)` with hardcoded USD `baseMax`/`increment` args, so even after fixing the text input, dragging the slider back to max would still snap to the unconverted ~200,000 ceiling.

## Steps to reproduce

1. Set base currency to IDR (Settings → General).
2. Create a new Retirement goal.
3. Open the plan sidebar → Expenses → Living.

## Expected vs actual

- **Actual:** Contribution/expense defaults show nonsensical tiny IDR amounts; typing a realistic monthly figure (e.g. 15,000,000 IDR) gets silently clamped; the slider caps at the same tiny value regardless of typed input.
- **Expected:** Defaults and input ceilings scale to the IDR equivalent of the original USD-scale amounts (e.g. $1,000/mo contribution → ~15,000,000 IDR), using an already-tracked exchange rate.

## Environment

- Wealthfolio 3.7.0, desktop (Tauri)

nofanto:
## Summary

Fixes #1475.

Retirement planner (and other `GoalLeverRow`-based goal) money fields were built around fixed USD-scale numbers that never adapted to the account's base currency:

- `DEFAULT_RETIREMENT_PLAN`'s contribution/expense presets (`monthlyContribution: 1000`, `Living: 3000`, `Healthcare: 300`) were raw USD numbers just relabeled with the account's currency at goal creation, not converted.
- `GoalLeverRow`'s money inputs clamped typed values to the same tiny ceiling used for the slider's practical drag range (~200,000 raw units), so users on high-denomination currencies (e.g. IDR) couldn't type a realistic value.
- The slider's own range had the same raw-unit ceiling, computed via `sliderMaxFor(value, baseMax, increment)` with hardcoded USD `baseMax`/`increment` — so even with the input fixed, dragging the slider still snapped back to the unconverted ceiling.

## Changes

- `goal-lever-row.tsx`: money-kind text inputs are no longer clamped to the slider's practical-range ceiling by default (only an explicit `inputMax` clamps them).
- `hooks/use-usd-to-currency-rate.ts` (new): resolves an already-tracked USD → currency exchange rate, plus a small `usdEquivalent` conversion helper.
- `plan-adapter.ts`: adds `scaleRetirementPlanAmounts` to scale a plan's money fields by a rate.
- `goal-new-page.tsx`: seeds new retirement plans with currency-converted defaults instead of raw USD numbers.
- `sidebar-configurator.tsx`: scales the USD `baseMax`/`increment` args passed to `sliderMaxFor` so both the slider's range and its ceiling reflect the real currency, not raw USD units.

Falls back to the original raw USD-scale behavior when no exchange rate is tracked yet for the currency pair, since there's no way to guess an appropriate scale without one.

## Test plan

- [x] `pnpm --filter frontend exec vitest run src/features/goals` — 12 passed (added 3 new tests for `scaleRetirementPlanAmounts`: scaling, rounding, immutability/non-monetary fields untouched)
- [x] `pnpm type-check` on touched files — clean
- [x] `pnpm lint` on touched files — clean (pre-existing warnings only)
- [x] Manually verified in `pnpm tauri dev` with base currency set to IDR: retirement plan defaults now seed sensible IDR amounts, typing a large monthly expense is no longer clamped, and dragging the expense/contribution/income sliders down and back up now reaches the correctly-scaled ceiling instead of snapping back to the raw ~200,000 cap.


afadil (changes requested):
thanks for this PR. Please change the solution direction.

All goal money values use the app base currency. All goal calculations use the same base currency. Portfolio values use `totalValueBase` before they enter the goal calculation.

The value `DEFAULT_RETIREMENT_PLAN.investment.monthlyContribution = 1000` means 1,000 units of the selected base currency. It does not mean USD 1,000. The `currency: "USD"` value in the template is a placeholder. Goal creation replaces it with the app base currency. The code must not convert these default values with an exchange rate.

The input limit is the real problem. The `max` value defines a useful slider range. It is not a business limit for the money value. The current code uses the same value for both purposes. As a result, the app can reduce a typed value. For example, it can reduce IDR 15,000,000 to IDR 200,000.

Could you please make these changes:

- Keep the change that separates the money input limit from the slider range.
- Accept the typed money value unless the caller supplies an explicit `inputMax`.
- Make the slider range include the current typed value.
- Do not let the slider reduce the stored value.
- Remove `useUsdToCurrencyRate`.
- Remove `scaleRetirementPlanAmounts`.
- Do not convert the default plan values.
- Use a function such as `createDefaultRetirementPlan(baseCurrency)` to set the correct currency in the default plan.

This solution fixes the input and slider problem. It also keeps all goal values in the app base currency and does not add an unnecessary exchange-rate dependency.

nofanto:
Thanks for the detailed feedback — you're right, and I've reworked the PR accordingly (pushed in `15566c83`).

Dropped the exchange-rate approach entirely:
- Removed `scaleRetirementPlanAmounts` and `useUsdToCurrencyRate`.
- Added `createDefaultRetirementPlan(baseCurrency)`, which only relabels the currency on `DEFAULT_RETIREMENT_PLAN` — amounts are untouched.

For the actual input/slider bug, it turned out `sliderMaxFor` was the root cause, not just the input clamp:

```ts
// before
const steppedMax = Math.max(baseMax, Math.ceil(value / increment) * increment + increment);
return Math.min(baseMax * maxGrowthMultiple, steppedMax); // re-clips to 10x baseMax

// after
const valueInclusiveMax = Math.ceil((value + increment) / increment) * increment;
return Math.max(baseMax, valueInclusiveMax); // uncapped, always covers `value`
```

It already grew with the typed value, but then re-clipped to `baseMax * 10` — for USD that ceiling ($200k) is generous enough nobody hits it, but for IDR (`baseMax` = 20,000 IDR) the same 10x cap is ~200,000 IDR, so typing a real monthly figure and dragging the slider snapped back to a ceiling that's orders of magnitude too small. Removing the cap makes the slider range monotonic in the current value — it always includes whatever's typed, in any currency, with no FX dependency. Money inputs stay unclamped unless an explicit `inputMax` is passed, as you suggested.

Also widened the money input boxes so larger IDR-scale figures with the `/mo` suffix aren't visually cramped.

Added a test file (`goal-lever-row.test.ts`) covering `sliderMaxFor`'s value-inclusive, uncapped growth. Let me know if this direction looks right.

PR #1486, opened 2026-08-18
Rebalance: allow selecting eligible holdings in Cash Flow Only mode

### LINKED ISSUE #1141: Rebalance – Allow selecting target holdings in "Cash Flow Only" mode
When using "Cash Flow Only" mode (purchases only), I'd like to be able to select which holdings are eligible for new purchases, rather than having all holdings considered.

marcoscale98:
## Description

Adds an **Eligible holdings** selector to the Cash Flow Only rebalance workflow.

- selects all unique non-cash holdings by default and supports search, grouping, individual selection, Select all, and Clear
- keeps exclusions session-only and resets them when the planning context changes
- sends a canonical asset allowlist through web and Tauri only when the selection is partial
- restricts Cash Flow Only buy candidates without changing current allocation, drift, sell behavior, or other rebalance modes
- exposes canonical instrument types on holdings and adds localized selector copy
- adds frontend, adapter parity, and core service coverage

Closes #1141.

### Verification

- `pnpm lint` (0 errors; existing warnings remain)
- focused frontend tests: 44 passed
- `cargo test -p wealthfolio-core portfolio::allocation_targets`: 123 passed
- focused Prettier check

## Checklist

### DIFF
--- a/apps/frontend/src/adapters/adapter-command-parity.test.ts
+++ b/apps/frontend/src/adapters/adapter-command-parity.test.ts
@@ -3,7 +3,14 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
+import { calculateRebalancePlan as calculateTauriRebalancePlan } from "./tauri";
 import { COMMANDS, invoke } from "./web/core";
+
+const { platformInvokeMock } = vi.hoisted(() => ({
+  platformInvokeMock: vi.fn(),
+}));
+
+vi.mock("#platform", () => ({ invoke: platformInvokeMock }));
 
 const currentDir = path.dirname(fileURLToPath(import.meta.url));
 const frontendSrcDir = path.resolve(currentDir, "..");
@@ -106,6 +113,38 @@ function collectNamedReexports(
 
   return { hasStar, names };
 }
+
+describe("rebalance eligibility transport", () => {
+  it("canonicalizes the Tauri shared request before web transport", async () => {
+    const mock = stubFetch({});
+    platformInvokeMock
+      .mockReset()
+      .mockImplementation((command, payload) => invoke(command, payload));
+
+    await calculateTauriRebalancePlan("target-1", 100, { type: "all" }, "cash_flow_only", [
+      "asset-z",
+      "asset-a",
+      "asset-z",
+    ]);
+
+    const { body } = lastCall(mock);
+    const parsed = JSON.parse(body as string) as { eligibleAssetIds?: unknown };
+    expect(parsed.eligibleAssetIds).toEqual(["asset-a", "asset-z"]);
+  });
+
+  it("omits the allowlist when no restriction is supplied", async () => {
+    const mock = stubFetch({});
+    await invoke("calculate_rebalance_plan", {
+      targetId: "target-1",
+      availableCash: 100,
+      filter: { type: "all" },
+      scenarioMode: "cash_flow_only",
+    });
+
+    const { body } = lastCall(mock);
+    expect(JSON.parse(body as string)).not.toHaveProperty("eligibleAssetIds");
+  });
+});
 
 describe("adapter command parity", () => {
   it("registers every command reachable from the web adapter", () => {
--- a/apps/frontend/src/adapters/shared/allocation-targets.ts
+++ b/apps/frontend/src/adapters/shared/allocation-targets.ts
@@ -105,16 +105,33 @@ export const saveTargetConstraints = async (
 
 // ── Rebalance ─────────────────────────────────────────────────────────────────
 
+export function canonicalizeEligibleAssetIds(
+  eligibleAssetIds?: readonly string[],
+): string[] | undefined {
+  if (eligibleAssetIds === undefined) return undefined;
+  return [...new Set(eligibleAssetIds)].sort();
+}
+
 export const calculateRebalancePlan = async (
   targetId: string,
   availableCash: number,
   filter: AccountScope,
   scenarioMode: ScenarioMode = "cash_flow_only",
+  eligibleAssetIds?: readonly string[],
 ): Promise<RebalancePlan> => {
-  return invoke<RebalancePlan>("calculate_rebalance_plan", {
+  const payload: {
+    targetId: string;
+    availableCash: number;
+    filter: AccountScope;
+    scenarioMode: ScenarioMode;
+    eligibleAssetIds?: string[];
+  } = {
     targetId,
     availableCash,
     filter,
     scenarioMode,
-  });
+  };
+  const canonicalIds = canonicalizeEligibleAssetIds(eligibleAssetIds);
+  if (canonicalIds !== undefined) payload.eligibleAssetIds = canonicalIds;
+  return invoke<RebalancePlan>("calculate_rebalance_plan", payload);
 };
--- a/apps/frontend/src/adapters/web/core.ts
+++ b/apps/frontend/src/adapters/web/core.ts
@@ -2005,17 +2005,19 @@ export const invoke = async <T>(command: string, payload?: Record<string, unknow
       break;
     }
     case "calculate_rebalance_plan": {
-      const { targetId, availableCash, filter, scenarioMode } = payload as {
+      const { targetId, availableCash, filter, scenarioMode, eligibleAssetIds } = payload as {
         targetId: string;
         availableCash: number;
         filter: unknown;
         scenarioMode: string;
+        eligibleAssetIds?: string[];
       };
       body = JSON.stringify({
         targetId,
         availableCash,
         filter,
         scenarioMode,
+        ...(eligibleAssetIds === undefined ? {} : { eligibleAssetIds }),
       });
       break;
     }
--- a/apps/frontend/src/adapters/web/index.ts
+++ b/apps/frontend/src/adapters/web/index.ts
@@ -356,6 +356,7 @@ export {
 export {
   archiveAllocationTarget,
   calculateRebalancePlan,
+  canonicalizeEligibleAssetIds,
   createAllocationTarget,
   deleteAllocationTarget,
   getAllocationTargetDrift,
--- a/apps/frontend/src/i18n/locales/de/allocation.json
+++ b/apps/frontend/src/i18n/locales/de/allocation.json
@@ -113,7 +113,7 @@
   "eligibleHoldings": {
     "label": "Geeignete Bestände",
     "triggerLabel": "{{label}}: {{summary}}",
-    "rowLabel": "{{symbol}} {{name}} — {{state}}",
+    "rowLabel": "{{symbol}} {{name}} {{details}} — {{state}}",
     "selected": "ausgewählt",
     "notSelected": "nicht ausgewählt",
     "allSelected": "Alle Bestände ausgewählt",
--- a/apps/frontend/src/i18n/locales/de/allocation.json
+++ b/apps/frontend/src/i18n/locales/de/allocation.json
@@ -110,6 +110,29 @@
     "modeVerbHybrid": "Bargeld und Verkäufe",
     "modeVerbCashFlow": "Cashflow-Käufe"
   },
+  "eligibleHoldings": {
+    "label": "Geeignete Bestände",
+    "triggerLabel": "{{label}}: {{summary}}",
+    "rowLabel": "{{symbol}} {{name}} — {{state}}",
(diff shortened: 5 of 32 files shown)


afadil:
Hi @marcoscale98 ,

After some research and a legal-risk assessment around regulated investment advice and personalized recommendations. My conclusion is that automatically generating security-level trades from a user’s portfolio could become a serious regulatory and liability problem across multiple jurisdictions.
So I'm working on a new direction for the rebalancing that preserves most of this feature value but moves Wealthfolio toward a simplier, safer, self-directed model:

- The user manually chooses every position, account, direction, and adjustment.
- Wealthfolio calculates and visualizes the resulting allocation impact.
- No optimizer-generated trades, security selection, ranking, or suitability claims.
- Cash and account scope remain useful worksheet inputs.
- Results stay descriptive: current, projected, target, and differences.

The attached screenshot shows the proposed direction: a more visual position-based worksheet where the user creates the changes and Wealthfolio provides the arithmetic and portfolio preview.

CleanShot 2026-08-11 at 09 13 03@2x


This is intended as risk mitigation. It definitely degrades the experience, but puts Wealthfolio in a better, safer position. 
Interested in you feedback on this.

Thanks

marcoscale98:
This surprises me for a few reasons.

1. Banks offer “smart investment plans” that perform rebalancing similar to what Wealthfolio does.

2. I do not see these as trades or personalized recommendations. They are mathematical calculations based on an allocation target explicitly chosen by the user. Removing allocation presets during target creation could make sense, since those might be interpreted as advice. But beyond that, the strategy is entirely user-defined.

The user also explicitly accepts that rebalancing will use instruments already held in their portfolio. With my proposed feature, they would have even more control, since they could directly choose which assets are allowed to participate in the rebalancing process.

It is unfortunate, but if you have done thorough research I trust your judgement. I will probably build an add-on for my personal use and attach it to Wealthfolio.

Jonjon-prog:
Hello @afadil and @marcoscale98 

I hope you are doing both great, sorry I couldn't take some time recently to advance on the allocation and rebalance part. 

@afadil I saw that you put the same message one\ an other PR of mine, and I will second @marcoscale98 on a little surprise too about that. 
Cause we neither say to the user to follow one thing as a true ground truth. The user here is the only one to decide the allocation or the only one to choose which type of preset of allocation he wants, after it's just a mathematical operation to suggest something. 

And I think it will be more even true, when we open the holdings targets level (which we don't have yet). 

We could put a line / text somewhere to back up the product, and tell to the user that it's not a financial advice, and that financial activity got some risks etc (a more law proof line but something like that). 

Cause allocation and rebalancing suggestions are in a lot fo software without even nocitying me that it can have risks once. 

Happy to continue to talk about that, cause for personal users and specially, after it's an important tool to gain some time and not do the calculation if we just want to follow our targets. (for passive investors, which is the target user of wealthfolio).

afadil:
@marcoscale98 @Jonjon-prog , thanks for the feedback, these are fair points, and I’m still thinking about the right direction here. The goal is to reduce legal risk without removing too much value from the feature.

My concern is that the old version did more than show allocation math. It selected specific securities, accounts, directions, and amounts for the user’s real portfolio. That could be seen as a personalized recommendation in some countries.

Banks offer similar tools, but they normally operate through registered entities with legal and compliance teams. Automated advice is covered by guidance in the 
[US](https://www.sec.gov/newsroom/press-releases/2017-52),  [Canada](https://lautorite.qc.ca/fileadmin/lautorite/reglementation/valeurs-mobilieres/0-avis-acvm-staff/2015/2015sept24-31-342-avis-acvm-en.pdf), and the [UK](https://handbook.fca.org.uk/handbook/perg8/perg8s41). A “not financial advice” message may not be enough by itself.

The current idea is to let the user choose every security, account, direction, and amount, while Wealthfolio only calculates and previews the result. This is a cautious approach, but it may be too conservative and could make the experience less useful.

I’d appreciate your toughs here. Do you think there is a safer middle ground that would keep more automation while making it clear that the user is making the decisions?

afadil:
What do you think about this direction:

1. Merge #1486 so users can choose the eligible holdings.
2. Keep the existing drift, cash, bands, turnover, minimum size, and whole-share calculations.
3. Let the user choose how adjustments are split between eligible holdings.
4. Start with one simple rule: **keep current proportions**.
5. Show the result as **calculated adjustments**, not “proposed trades.”
6. Keep every adjustment editable and require user review.
7. Do not rank securities or decide which investment is best for the user.

This would keep most of the useful automation while making the user responsible for the target, eligible holdings, and allocation rule.

@Jonjon-prog do you want to lead a design document for the whole feature roadmap so we can refine together before implementation?

Jonjon-prog:
Hey @afadil @marcoscale98, thanks both for your inputs on that. 

@afadil your 7-point direction works for me just some clarification and adjustments. 

I spent the last hours in the actual texts (FCA PERG, MiFID II + the ESMA 2023 briefing, Advisers Act, NI 31-103) and it lines up almost exactly with where you landed. 
Below is what I found, split into what I think we're safe on and what we're not, plus a fix for each, keeping the UX simple, because that's the whole point of the feature.


## The line that actually matters

The clearest formulation is FCA PERG 8.30A on filtering tools: a filter based on **factual criteria** is not regulated advice; a filter incorporating **opinion, judgment or ranking** is.

So the question is never "does the tool output numbers". It's **who supplied the judgment**. Everything below follows from that.

## Allocation targets,  I think we keep all of it

Targets, drift, bands, the donuts: the target is 100% authored by the user and the output is at asset-class level. UK RAO art 53 requires a **particular** investment; MiFID (DR 2017/565 art 9) requires a **specific financial instrument**. Class-level output sits outside both. This is the least exposed part of the app.

**Except one thing: named presets** (60/40, All Weather, …). 
That's the single place where Wealthfolio supplies judgment instead of arithmetic, @marcoscale98 called this early and I think he's right. 
It's also the cheapest fix in the whole discussion.

*Fix:* drop the strategy names, or keep them as unnamed starting layouts with no performance figures, no backtest, no attribution. No forward-looking claims anywhere ("this reduces your risk" etc.).
That's what turns a template into a recommendation.

## Rebalancing, the engine is fine, three narrow things are not

Drift, cash, bands, turnover cap, minimum trade size, whole-share rounding: pure arithmetic on rules the user wrote. Nothing there needs to go.

What needs to change:

**1. The optimizer choosing *which* security inside a sleeve.** 
This is the real exposure, and it's exactly your point 7. 
*Fix:* your ladder, eligible holdings (#1486) → keep current proportions → optionally a user-set instrument per sleeve, configured once at target creation. One dropdown, once. UX cost is basically nil and the security choice becomes the user's structurally, not by disclaimer.

**2. Automatic account assignment.** 
Smaller point, and I want to be precise because it sounds pedantic: picking *which account* to buy in is asset location, which is a tax decision (PEA/CTO, TFSA/RRSP, ISA/SIPP, IRA/taxable), and tax advice is regulated separately in several countries. 
*Fix:* not removal: justcmake the rule mechanical, visible and overridable: "buys go to the account that already holds this security, otherwise the one holding the cash." 
That's simpler than a hidden optimizer, not harder.

**3. Register.** 
"Proposed trades", share quantities, a CSV that reads like an order ticket. 
*Fix:* your "calculated adjustments" wording, amounts rather than share counts as the primary figure, and show the arithmetic (`target% × total − current value = gap`) so it visibly reads as a calculator. 
FINRA Rule 2214's disclosure wording for investment analysis tools is a decent model to borrow.

## One UX idea worth considering

Rather than the screen handing the user a plan, have it ask what they're considering, with a **"fill from drift"** button. 
Same numbers, every line editable, but the user pulls the figure instead of receiving it. 
It's your worksheet without the loss of value, and it costs one button.

## Two things missing from the thread

**Passiv.** 
Canadian, same jurisdiction, and it does *more* than us: user targets, rebalancing math, and one-click order routing to the brokerage. 
Not registered as an adviser. That's a much stronger comparable than "banks do it",  banks cut against us, since they're registered entities.

**We're local-first and never say so.** 
The calculation runs on the user's machine; the portfolio never reaches us. 
"Advising others" presupposes an *other* who receives the data. A spreadsheet does this exact math and nobody worries.
Add no execution, no custody, no discretion, no order routing, and AGPL source anyone can audit, that's a solid structural position and we state it nowhere.

## And a note on holdings-level targets

@afadil this is worth flagging: letting users set targets per holding **reduces** this risk rather than adding to it. If the user writes "15% VWCE" themselves, there is nothing left for us to choose.
It's the answer to your concern, not a new instance of it.

## Two corrections, one to myself

1. I leaned on a disclaimer earlier.  
I was wrong. ESMA and the FCA both assess  substance over form; a "not financial advice" line doesn't requalify something that is otherwise a personal recommendation. Worth having, can't be  load-bearing. 
You were right.

2. CSA 31-342 is guidance for **already-registered portfolio managers** running **discretionary** online advice (KYC, suitability). 
With no discretion, no custody and no execution, I don't think it reaches us. 
Same for the SEC 2017 release,  it's addressed to registered robo-advisers, not a statement that calculators must register.

## Design doc

Yes, happy to lead it. Proposed scope: goals and non-goals, the split-rule ladder, sleeve-level vs instrument-level output, the account rule, copy and terminology, presets, export format, holdings-level targets, and how all of this interacts with the AI assistant,  which is the biggest amplifier of everything above. 
I'll open a draft in the week if I can and tag you both.

(If we ever want a sanity check on the above, a short consult with a securities lawyer would settle it quickly,  but I don't think it blocks the design work.)

afadil:
@Jonjon-prog @marcoscale98 , 
Thanks, this is very, personally agree on these four decisions:

1. The first supported allocation rule is Keep current proportions.
2. The user explicitly chooses eligible securities and accounts.
3. The app performs no ranking, suitability assessment, or hidden selection.
4. The result is editable arithmetic, not a recommended plan.

this direction makes sense. Please go ahead with the design doc.

For #1486, do you think it can be merged as-is and reused for this direction, or should we adjust it after agreeing on the design?

Jonjon-prog:
Thanks @afadil, and good to have the four decisions settled — I have added them to the design doc as agreed constraints.

On #1486, my read is that it can be merged as-is, but since it is @marcoscale98's PR the call should be his too.

From the design side it needs no rework: the eligible-holdings selector is exactly the second rung of the split ladder (eligible holdings → keep current proportions → optional instrument targets per sleeve).
It is also inert unless the user narrows the selection, since the allowlist is only sent when the selection is partial, and it leaves drift, sells and the other scenarios untouched — so there is nothing to unwind if we refine the design later.

Three follow-ups I would keep out of it and do afterwards, all additive:

1. **Scenario scope.** The allowlist applies in Cash-flow only today and is dropped for Sell to rebalance and Hybrid, both in the hook and in `rebalance_service`. Once the proportional split rule lands, the eligible set becomes the denominator of that split in every scenario, so the guard widens then.
2. **Persistence.** Exclusions are session-only. For a plan to be reproducible, and for the export header to state "eligible holdings (n of m)", the set needs to live on the target profile. Cheapest path is probably `allocation_target_constraints` from #1177 with an asset-scoped exclude-from-buys action, rather than a new table.
3. **Buys only.** I would keep the eligible set gating buys, and leave sells to the do-not-sell constraints from #1177 — two overlapping mechanisms for the same decision would be confusing. "I don't want to add to this position" and "I don't want to sell this position" are genuinely different intents.

@marcoscale98 does that match how you see it, and are you happy to merge as-is with those as follow-ups?

One clarification on your decision 2, @afadil ("the user explicitly chooses eligible securities and accounts").
For securities that is the ladder.
For accounts, the doc currently reads it as a mechanical rule that is displayed and overridable on every line: buys are attributed to the account that already holds the instrument, otherwise to the account holding the cash.
My preference is that reading, because it costs a single-account user zero clicks and an overridable rule shown in full is not a hidden selection.
The stricter reading — no default, the user picks on every line — is a one-line change if you would rather have it, so tell me which one you want and I will lock it in.

The design doc is now open as #1565.

marcoscale98:
I have rebased #1486 onto the latest `main` and resolved the conflict, preserving the PR’s existing scope. It is now mergeable.

I have now installed the worktree dependencies and rerun the relevant checks locally: package type build, formatting, lint (existing warnings only), type check, 47 focused frontend tests, `cargo fmt --all -- --check`, and the focused allocation-target Rust tests (`123 passed`).

This completes my contribution to this PR. Please merge it if it remains appropriate for M0. I will not be taking on further changes related to the broader rebalancing redesign; any follow-up work can proceed independently of #1486.

afadil:
@marcoscale98 thanks for this contribution

PR #1487, opened 2026-08-18
Add closed positions and cash to holdings

### LINKED ISSUE #1417: Issue : Cash should be included in 'Holdings' Tab and show % of account
breakdown by currency as well
Thanks Afadil
Sent you 25$ for a coffee

afadil:
## Summary

- add a responsive holdings visibility facet filter for Open, Closed, and Cash, defaulting to Open
- expose closed positions through the desktop and web APIs and preserve explicit lifecycle state during multi-account aggregation
- keep cash balances first and present only cash-relevant values without the manual badge
- show cash portfolio weight on desktop and mobile
- align the mobile search, filter, and actions controls in one row while preserving account scope controls for empty accounts
- calculate and display realized gain for closed lots, including option-expiry disposals

## Why

Closed positions were excluded before reaching the UI, cash could not be included from the holdings view, and a zero aggregate quantity was not a reliable indicator that a position was closed. Closed lots also lacked a useful realized-gain value, and option expiries were incorrectly filtered out of realized results.

## User impact

Users can focus on open positions, closed positions, cash, or any combination from the same familiar faceted-filter pattern on desktop and mobile. Open remains the default. Closed positions remain distinct from open positions for the same asset, even when open long and short positions net to zero. Mobile users retain account selection when the chosen account is empty.

CleanShot 2026-08-18 at 11 05 12@2x

## Validation

- `pnpm --filter @wealthfolio/ui type-check`
- `pnpm --filter frontend test` — 149 files, 1,241 tests passed
- focused mobile holdings tests — 2 files, 7 tests passed
- focused option-expiry realized-gain regression test
- `cargo check --workspace`
- `cargo test -p wealthfolio-core closed_positions --lib`
- Prettier, targeted ESLint, and `git diff --check`

Fixes #1417

### DIFF
--- a/apps/frontend/src/adapters/adapter-command-parity.test.ts
+++ b/apps/frontend/src/adapters/adapter-command-parity.test.ts
@@ -267,6 +267,17 @@ describe("scope-based routing — get_holdings_list", () => {
     expect(method).toBe("GET");
   });
 
+  it("account with closed positions → GET with includeClosed", async () => {
+    const mock = stubFetch();
+    await invoke("get_holdings_list", {
+      filter: { type: "account", accountId: "acc_1" },
+      includeClosed: true,
+    });
+    const { url, method } = lastCall(mock);
+    expect(url).toBe("/api/v1/holdings/list?accountId=acc_1&includeClosed=true");
+    expect(method).toBe("GET");
+  });
+
   it("portfolio → POST /holdings/list/query", async () => {
     const mock = stubFetch();
     await invoke("get_holdings_list", { filter: { type: "portfolio", portfolioId: "pf_1" } });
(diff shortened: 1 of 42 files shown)


kryofin:
@afadil Shouldn't cash be a type (ETF/Stock/MF) vs. a view? Cash looks like it is in the wrong thematic location.

afadil:
@kryofin good catch, I'll fix it in a follow-up.
thanks

kryofin:
> @kryofin good catch, I'll fix it in a follow-up. thanks

Would you like me to open an issue so that we don't lose track or do you already have it?

afadil:
I'm on it , will push soon

PR #1493, opened 2026-08-19
fix: preserve no-flow TWR provenance

### LINKED ISSUE #1485: Spurious TWR warning due to NO_FLOW days
Quiet `NO_FLOW` days are relabeled `NetContributionFallback`, so the "flows were inferred from net contribution deltas" note shows on essentially every portfolio.

Observed on `main` at `1194bb54` (v3.6.3-83); reproduced on a server build against real data.

## Summary

The performance insights tab shows this calculation note on any scope whose
history contains at least one quiet day — which is every real portfolio:

> External cash flows were inferred from net contribution deltas for part of
> this period because gross daily flow data was unavailable; same-day deposits
> and withdrawals may be netted.

On the data that triggered it, **nothing was inferred**: every day that took
the fallback path had a net-contribution delta of exactly zero. The note is
fired by quiet days, not by days with genuinely unavailable gross flow data.

## Root cause

`ExternalFlowSource::NoFlow` is documented as the neutral identity — *not*
degraded, available for returns, "the absence of a flow, not an unvaluable
one" (`crates/core/src/portfolio/valuation/valuation_model.rs`):

```rust
/// No external flow occurred. Neutral identity for provenance merging and the
/// default for freshly-constructed rows and aggregation fillers. Available for
/// returns and not degraded — it represents the *absence* of a flow, not an
/// unvaluable one.
#[default]
NoFlow,
```

But `daily_external_flows` in
`crates/core/src/portfolio/performance/performance_service.rs` discards that
classification. In the `ExternalFlowBasis::BaseCurrency` arm, a `NoFlow` row
falls through every guard — it is not `UnknownBoundaryTransfer`, not
`is_unavailable_for_returns()`, not `is_explicit_gross()`, and its stored
gross columns are zero — and lands on the final fallback:

```rust
curr_point.net_contribution_base - prev_point.net_contribution_base
```

which is stamped `ValuationExternalFlowSource::NetContributionFallback`
unconditionally. (The `ExternalFlowBasis::AccountCurrency` arm stamps *every*
window `NetContributionFallback`, so single-account scopes computed in account
currency are affected on every single day.)

`external_flow_quality_warnings` then emits the note if **any** day in the
series carries that source:

```rust
let used_net_fallback = daily_flows
    .iter()
    .any(|flow| flow.source == ValuationExternalFlowSource::NetContributionFallback);
```

So one weekend in the period is enough to display the warning, regardless of
data quality.

## Why it matters

- The note claims "gross daily flow data was unavailable" when it was
  available and explicitly zero. Users auditing a migration (or any long
  history) chase a data problem that does not exist.
- Because it fires on effectively every portfolio, the note carries no
  signal — the one time gross flow data really *is* missing, the warning looks
  identical to the permanent background noise.
- It also forces `dataQuality.status` to `partial` on healthy data.

## Suggested fix

Either (or both):

1. In `daily_external_flows`, early-return for `NoFlow` rows with zero stored
   gross flows, preserving `NoFlow` as the source — matching the enum's
   documented contract ("not degraded, absence of a flow").
2. In `external_flow_quality_warnings`, only count a `NetContributionFallback`
   day as `used_net_fallback` when its inferred flow is nonzero — a zero delta
   inferred nothing and netted nothing.

Option 1 is the more principled: the quiet day never belonged on the fallback
path, and it keeps the `AccountCurrency` arm's behavior an explicit choice
rather than an accident of the shared stamp.

Neallin-917:
## Description

Closes #1485.

Preserves a zero-valued `NoFlow` valuation row as neutral provenance instead of relabeling it as `NetContributionFallback`. This prevents healthy quiet days from showing the warning that cash flows were inferred from unavailable gross data.

Adds regression coverage for both flow bases and the user-visible performance warning.

## Verification

- `cargo test -p wealthfolio-core`
- `cargo clippy -p wealthfolio-core --all-targets --all-features -- -D warnings`

## Checklist

### DIFF
--- a/crates/core/src/portfolio/performance/performance_service.rs
+++ b/crates/core/src/portfolio/performance/performance_service.rs
@@ -751,6 +751,18 @@ impl PerformanceService {
         flow_basis: ExternalFlowBasis,
     ) -> DailyExternalFlow {
         let date = curr_point.valuation_date;
+        if curr_point.external_flow_source == ValuationExternalFlowSource::NoFlow
+            && curr_point.external_inflow_base.is_zero()
+            && curr_point.external_outflow_base.is_zero()
+        {
+            return DailyExternalFlow {
+                date,
+                inflow: Decimal::ZERO,
+                outflow: Decimal::ZERO,
+                source: ValuationExternalFlowSource::NoFlow,
+            };
+        }
+
         let cash_flow = match flow_basis {
             ExternalFlowBasis::AccountCurrency => {
                 curr_point.net_contribution - prev_point.net_contribution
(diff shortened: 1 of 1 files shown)


kryofin:
Thank you for the fix @Neallin-917. I think it should work and the CI failure is spurious (rebase/force push) should retrigger. I hope it gets merged soon.

Neallin-917:
Thanks for spotting that. I rebased the branch onto the latest main and retriggered the PR workflow; both checks are queued now. The focused NoFlow regressions and strict Clippy checks also pass on the rebased commit.

Neallin-917:
Friendly follow-up: this focused NoFlow provenance fix remains clean with all three PR checks passing. It fixes the incorrect degraded-data warning while preserving the existing non-zero fallback behavior. When convenient, could a maintainer review whether it is ready to merge? I’m happy to make any requested adjustment. Thank you.

PR #1498, opened 2026-08-19
fix(allocation): Account for the unclassified remainder in category drill-downs

### LINKED ISSUE #1497: Drilling into an asset class shows a sub-class as 100% of it
Clicking a class with sub-classes shows only the sub-classified part, and
percentages are computed against that subset — so a lone sub-class always reads
100%, even when it's a fraction of the class.

**Repro** (Insights → Classes, All Accounts)

1. Classify one holding as `FI_MUNICIPAL` (worth 1,200) and another as
   `FIXED_INCOME` only, with no sub-class (worth 800).
2. Fixed Income correctly totals 2,000 in the breakdown list.
3. Click Fixed Income in the Classes donut.

**Expected:** Municipal Bonds ≈ 60%, with the remaining 800 accounted for.
**Actual:** Municipal Bonds $1,200 (100.00%). The 800 is invisible.

**Cause:** `aggregate_by_taxonomy` emits a child row only for sub-category
assignments, so holdings classified at the top level alone add to the parent's
value but produce no child. Consumers then divide by the children they received.

Three surfaces under-report independently: the drill-down donut, the Insights
breakdown tree, and the allocation detail sheet (which the net-worth Investments
drawer reuses).

kryofin:
Fixes #1497 .

A rolled-up category only emitted children for sub-category assignments, so
holdings classified at the top level alone (e.g. "Fixed Income" with no bond
type) added to the parent's value but produced no child row. Every drill-down
then divided by the children it received: a class worth 2,000 with a single
1,200 sub-class reported that sub-class as 100%.

### Changes

- `aggregate_by_taxonomy` appends a residual child, `<parentId>:__residual__`,
  holding the difference — children now always account for their parent.
- `get_holdings_by_allocation` resolves that id by matching shares assigned to
  the parent with no sub-category to roll up, so the row addresses real holdings
  instead of falling back to the whole class.
- The backend names the row in English for consumers that don't know the marker;
  the UI substitutes a localized label (`common:allocation_other_in_category`,
  all 7 locales).

Fixes the donut drill-down, the Insights breakdown tree, and the allocation
detail sheet.

### Why in the backend

The omission is silent, and three consumers hit it independently. A frontend
helper would have to be remembered by every future consumer; a residual emitted
once cannot be missed. It also lets the holdings query answer the residual id
exactly — a frontend-only fix has to open the parent, listing sub-classified
holdings that don't belong to the row clicked.

### Testing

- 4 new Rust tests (residual added / not added when children cover the parent /
  leaf categories / residual holdings query returns only top-level-only holdings)
- Frontend: `type-check` clean, full suite passes, 4 new tests for the label helper
- Verified end-to-end against a running server: parent = sub-class + residual,
  donut drill-down reads 59.10% instead of 100.00%, and the residual row lists
  only the top-level-classified holdings

## Checklist

### DIFF
--- a/apps/frontend/src/i18n/locales/de/common.json
+++ b/apps/frontend/src/i18n/locales/de/common.json
@@ -364,5 +364,6 @@
   "navStyle": {
     "sidebar": "Seitenleiste",
     "floating": "Schwebeleiste"
-  }
+  },
+  "allocation_other_in_category": "Andere {{category}}"
 }
--- a/apps/frontend/src/i18n/locales/de/common.json
+++ b/apps/frontend/src/i18n/locales/de/common.json
@@ -365,5 +365,5 @@
     "sidebar": "Seitenleiste",
     "floating": "Schwebeleiste"
   },
-  "allocation_other_in_category": "Andere {{category}}"
+  "allocation_other_in_category": "Sonstige ({{category}})"
 }
--- a/apps/frontend/src/i18n/locales/de/common.json
+++ b/apps/frontend/src/i18n/locales/de/common.json
@@ -363,5 +363,6 @@
   "navStyle": {
     "sidebar": "Seitenleiste",
     "floating": "Schwebeleiste"
-  }
+  },
+  "allocation_other_in_category": "Andere {{category}}"
 }
--- a/apps/frontend/src/i18n/locales/en/common.json
+++ b/apps/frontend/src/i18n/locales/en/common.json
@@ -364,5 +364,6 @@
   "navStyle": {
     "sidebar": "Sidebar",
     "floating": "Floating Bar"
-  }
+  },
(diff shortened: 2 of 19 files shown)


kryofin:
@afadil Anything else needed after your fix commit?

kryofin:
@afadil Can we get this merged? You have already done a pass and I just added support for newly-added languages to ensure a clean merge.

afadil:
@kryofin, thank you. Sorry for the delay; a lot of work to do :)

kryofin:
> @kryofin, thank you. Sorry for the delay; a lot of work to do :)

@afadil No worries at all! You have a lot on your plate. LMK if there is anything else I can help with. Otherwise, I will keep on squashing bugs I see.

PR #1499, opened 2026-08-19
fix(core): flag an import trade row whose total contradicts quantity x unit price

rwgs:
## The problem

A BUY/SELL import row can supply `quantity`, `unitPrice` **and** `amount`, and
nothing checks that the three agree.

`gross_trade_amount` in
`crates/core/src/portfolio/snapshot/holdings_calculator/economics.rs` books a
plain trade from `quantity * unit price`; `should_use_activity_amount` returns
`false` as soon as a non-bond BUY/SELL has both a quantity and a unit price, so
the broker total is never read. A total that disagrees is therefore a cost basis
that is silently not the amount that left the account — and neither
`check_activities_import` nor `validateDraft` in
`apps/frontend/src/pages/activity/import/utils/draft-utils.ts` looks at it.

The shape that makes this routine rather than rare is a **foreign-listed trade**.
Brokers commonly print the execution price in the instrument's listing currency
while the total and the row currency are the account's. An activity carries one
`currency`, so mapping such a row naively produces something *internally
consistent and wrong*:

| column | value |
| --- | --- |
| quantity | 10 |
| unitPrice | 50.00 (listing currency) |
| amount | 675.00 |
| currency | CAD |

`10 x 50.00 = 500.00`, so the position is booked with 500.00 CAD of cost basis
instead of the 675.00 CAD that was actually paid — understated by the FX factor,
on every foreign-listed position, with nothing in the app able to detect it. The
lot side is worse when the asset's quote currency differs from the row currency,
because `convert_to_position_currency` then converts a figure that was never in
the row currency to begin with.

## The change

`check_activities_import` reconciles the three figures on BUY/SELL rows and
attaches a warning to `unitPrice` when they cannot agree:

- **Both ends of the commission band reconcile.** Brokers state the total either
  gross or net of commission, so `amount` is accepted anywhere within
  `fee + tax` of `quantity * unit price`.
- **A 1% ratio absorbs price rounding** — a price printed to fewer decimals than
  it filled at — and stays well below any currency or unit mismatch.
- **When a stated `fxRate` explains the gap**, the message says so and names the
  rate, since that is the actionable case: the price is quoted in another
  currency and needs restating in the row's.
- **Bonds and options are exempt**, because the calculator already prefers the
  broker total for bonds (quoted as a percentage of par) and scales the price by
  the contract multiplier for options, so a disagreement there is expected.

### Warn, not reject

A broker that states the trade in one currency and settlement in another
produces exactly the same disagreement from a row that imports *correctly* — the
price and row currency agree and only the `amount` column is denominated
elsewhere, which the calculator already ignores. Rejecting would block those
imports, so which figure is authoritative stays the user's call.

No frontend change is needed: backend warnings are merged into the draft by
`import-context.tsx` and keys matching a column id are already rendered by
`import-review-grid.tsx`, so `unitPrice` highlights the cell, shows the message
on hover, and the row counts toward the review step's warning badge and filter.

## Tests

Five tests in `activities_service_tests.rs` drive `check_activities_import`
end to end:

- a total that only reconciles through the stated FX rate is flagged, and the
  message names the rate, the booked figure and the stated total
- the same disagreement with no FX rate mapped is still flagged, without
  claiming a rate the file did not supply
- a total that includes the commission is accepted
- a total off by price rounding is accepted
- a bond quoted as a percentage of par is skipped

`cargo test -p wealthfolio-core --lib` passes (the three `addons::tests` failures
on this machine reproduce on a clean `main` and are unrelated to this change).
`cargo clippy -p wealthfolio-core --all-targets` is clean.

## Checklist


afadil:
@rwgs, Thanks for this PR. Codex flagged these two edge cases worth addressing before merge:

1. **[P2] Contract-multiplier trades receive false warnings.**  
   [`gross` is calculated as only `quantity * unit_price`](https://github.com/wealthfolio/wealthfolio/blob/1b8beb6d7237299be9428aa70b7d05c469870fbb/crates/core/src/activities/activities_service.rs#L1434-L1439), while the holdings calculator also applies `asset_info.contract_multiplier`. Options are exempted, but Wealthfolio explicitly supports metadata-based multipliers on equity-like assets such as futures. A valid future trade—quantity `1`, price `100`, multiplier `50`, amount `5000`—is therefore warned as though `100` were the booked value. My diagnostic test reproduced this false warning. The validator should use the resolved asset’s multiplier and mirror `gross_trade_amount`.

2. **[P2] A tiny positive FX rate can panic the validation request.**  
   [The unchecked `gross / rate`](https://github.com/wealthfolio/wealthfolio/blob/1b8beb6d7237299be9428aa70b7d05c469870fbb/crates/core/src/activities/activities_service.rs#L1446-L1448) can overflow `rust_decimal`; filtering out zero does not prevent this. I reproduced `Division overflowed` with `gross = 10`, `amount = 2`, and `fxRate = 1e-28`, which is a valid positive decimal accepted by the import schema. One malformed row can consequently abort validation for the entire batch. Use `checked_mul`/`checked_div` and treat unrepresentable candidates as non-matches.

rwgs:
Thanks @afadil - both reproduce, and the second one had a sibling worth fixing at the same time. Pushed as 5fb3d26 and 3cbc689.

**1. Contract multiplier.** Confirmed. `gross_trade_amount` applies `asset_info.contract_multiplier`, and `Asset::contract_multiplier()` reads `contractMultiplier` from metadata for *any* asset — the 100 fallback is options-only. A future has no instrument type of its own, so it is an equity-like asset carrying its multiplier in metadata, which is exactly what `BrokerSyncService` writes for a non-option position. Your row warned verbatim: `Quantity times unit price (100 CAD) does not match the stated total (5000 CAD)`.

`check_trade_total_consistency` now takes the resolved asset's multiplier and computes `gross` as `quantity * unit_price * multiplier`, mirroring `gross_trade_amount`. Bonds and options stay exempt, but I corrected the doc comment's rationale — "scales the price by the contract multiplier for options" is no longer why options are out. The honest reason to keep them out is that an option's multiplier is only *assumed* to be the standard 100 when the asset carries no metadata saying otherwise, so the booked figure is itself uncertain there.

**2. FX-rate overflow.** Confirmed — `Division overflowed`, `rust_decimal-1.42.0/src/arithmetic_impls.rs:218`, from `gross = 10`, `amount = 2`, `fxRate = 1e-28`. Both candidates now go through `checked_mul`/`checked_div`, and a candidate that cannot be represented is treated as a non-match, which is what it is.

While in there: `quantity * unit_price` has the same exposure and is reached earlier. Large-but-representable operands panic with `Multiplication overflowed` before any FX rate is consulted, and the multiplier above adds a third factor to it. `gross` is now computed with checked arithmetic too, and a product we cannot represent returns without warning rather than claiming something about a number we could not compute.

**3.** One consequence of (1) worth naming: the warning opened `Quantity times unit price (5000 CAD)`, which stops being true the moment the figure carries a multiplier — a reader with the quantity and price columns in front of them cannot get 5000 out of 1 and 100. It now names what the figure is made of, and names the multiplier when there is one.

Four tests, each written to fail first: a multiplier row that reconciles; a multiplier row quoted in another currency, which still warns (the multiplier corrects the check rather than switching it off); the 1e-28 rate; and a product too large to represent. 303 `activities::` tests pass, and `cargo fmt` plus `cargo clippy -p wealthfolio-core --all-targets -- -D warnings` are clean.

afadil:
superseded by #1571.

Issue #894, opened 2026-04-26
Market Data providers rate limiting required? Especially for AlphaVantage?


richard-lee-863:
I set up a free API with Alpha Vantage. They offer 25 requests per day. 

[https://www.alphavantage.co/support/#api-key](https://www.alphavantage.co/support/#api-key)

How often is Wealthfolio requesting to refresh quote data?

Looks like the [TIME_SERIES_DAILY](https://www.alphavantage.co/documentation/#dailyadj) api only lets you request historical data for one ticker at a time, and the free version only supports the last 100 values (days)

So I suppose users with > 25 holdings will need to be aware they'll need other providers to kick in after that.

rwgs:
@afadil Answering the "is rate limiting required?" part concretely, because the code currently claims a limit it cannot enforce.

**`RateLimit` has no per-day concept.** `crates/market-data/src/provider/capabilities.rs` defines it as exactly three fields — `requests_per_minute`, `max_concurrency`, `min_delay` — and the limiter behind it (`crates/market-data/src/registry/rate_limiter.rs`) is a per-minute token bucket. `grep -rn "per_day\|daily\|per_hour"` over the whole crate returns only Alpha Vantage's daily-*quote* parsing. So a 25-requests-per-day cap is not merely unenforced, it is unrepresentable.

**Alpha Vantage states three different limits in one file:**

| Location | Claim |
| --- | --- |
| `provider/alpha_vantage/mod.rs:9` (module doc) | "free tier is limited to 5 API calls per minute" |
| `provider/alpha_vantage/mod.rs:1126` (the field) | `requests_per_minute: 25` |
| same line, the comment on it | `// Free tier: 25 requests/day` |

The field is what the bucket reads, and it says *per minute*. So the limiter will happily issue the entire daily allowance inside the first 60 seconds, after which every call fails for the rest of the day. `min_delay: Duration::from_secs(3)` carries the comment "~25 requests per minute", which is a fourth reading of the same number.

**On your arithmetic** — yes, and it is worse than per-refresh. Enrichment is per-asset too, so on a 77-asset portfolio a first sync needs four days of daily allowance to fetch profiles alone, and nothing in the app paces it across days or tells the user that is what is happening. The 429/`Note` handling at `mod.rs:609-621` detects the rejection after the fact but cannot prevent it.

**What a fix probably needs:**

[...]

Issue #827, opened 2026-04-08
Feature Request: ETF geographic look-through — real country exposure for portfolio insights


marcoscale98:
## Problem

The current **Country Allocation** in Insights reflects the *domicile* of each asset, not the *underlying geographic exposure*. This means a portfolio holding only VWCE or MSCI World ETFs appears as ~100% Irish (where the fund is domiciled), while the real exposure is ~65% USA, ~6% Japan, ~4% UK, etc.

This has been partially raised in #551 and #558, but the root issue is that **no current market data provider supports ETF geographic look-through**.

---

## What we'd want

Given a holding like:

- €10,000 in VWCE (Vanguard FTSE All-World) → ~62% USA, ~6% Japan, ~4% UK, ...
- €5,000 in a pure US stock → 100% USA

The Insights Country Allocation should show the **aggregated real exposure** across all positions, drilling through ETF wrappers to their underlying geographic weights.

---

## Analysis of current providers

| Provider | Geographic ETF data? | Notes |
|---|---|---|
| **Yahoo Finance** | ❌ No | `topHoldings` only exposes `sectorWeightings`. `country` field = domicile only |
| **Alpha Vantage** | ❌ No | `ETF_PROFILE` explicitly excludes country (confirmed in codebase) |
| **Finnhub** | ⚠️ Paid only | Has `/api/v1/etf/country-exposure` returning `[{country, exposure%}]` — but requires ETF subscription tier (paid) |
| **OpenFIGI** | ❌ Not applicable | Identifier mapping only |
| **Börse Frankfurt** | ❌ Not applicable | Pricing only |

**Recommendation: avoid paid-only solutions.** Locking this feature behind a Finnhub ETF subscription would make it inaccessible to most users.

---

## Possible free approaches (looking for input)

- **JustETF** — excellent free coverage for European-listed ETFs, but no official API (scraping only, fragile)
- **Manual override** — allow users to manually define the geographic breakdown per ETF ticker, stored locally
- **Addon** — an addon that fetches/caches ETF composition from a free source and injects it into the allocation calculation
- **ETF provider static data** — some ETF issuers (iShares, Vanguard) publish JSON/CSV holdings files; could be fetched directly

---

## Questions for the community

1. Is anyone aware of a **free, reliable API** (or open dataset) that provides ETF country/geographic exposure weights — especially for non-US ETFs (UCITS)?
2. Would a **manual per-ticker override** (similar to how country can be corrected today) be a good interim solution?
[...]

PR #1503, opened 2026-08-19
fix(classification): resolve a country to its own region node, not its continent

rwgs:
## The problem

The seeded `regions` taxonomy is three levels deep — continent → sub-region →
country — and carries **227 `country_*` categories** keyed on ISO 3166-1
alpha-2 (`crates/storage-sqlite/migrations/2026-01-01-000002_taxonomies/up.sql`).

`map_country_to_region` in `crates/core/src/assets/auto_classification.rs`
hand-coded roughly fifty country names against that, and returned a `country_*`
id for **six** of them — US, CA, JP, CN, HK, AU. Everything else landed on a
continent or sub-region:

```rust
"united kingdom" | "uk" | ... | "germany" | ... | "russia" => {
    Some("R10")
} // Europe
```

`country_DE` is seeded, under `R1020` (Western Europe). So is `country_GB`,
and 40 more under Europe alone. The classifier walked past all of them.

**A German holding cannot be told apart from a Portuguese one.** Both are
`R10`. A US holding, in the same taxonomy on the same enrichment pass,
resolves to `country_US`. The granularity of the answer depends on which
country the provider happened to name.

**Europe has no drill-down at all.** `aggregate_by_taxonomy` in
`crates/core/src/portfolio/allocation/allocation_service.rs` builds a child row
only where the assigned category differs from its top-level ancestor:

```rust
// Only add as child if different from top-level (i.e., it was rolled up)
if cat_id != top_level_id && *value > Decimal::ZERO {
```

An asset assigned directly to `R10` therefore contributes nothing to Europe's
children, and `DrillableDonutChart` opens the holdings sheet instead of
drilling because `category.children` is empty. Click Americas and it opens into
United States; click Europe and there is nothing beneath it, with 42 seeded
country nodes sitting there unused.

And the coverage half: a country name outside the fifty produced **no region
assignment at all**, so the holding counted as unclassified in the regions
breakdown rather than being placed.

## The change

`resolve_country_category` matches the provider country against the regions
taxonomy **itself**, in three steps:

1. **A bare ISO 3166-1 alpha-2 code**, which is what several providers send
   instead of a name, and what the seeded ids are keyed on.
2. **An alias**, for the spellings where provider and taxonomy disagree. The
   taxonomy uses UN names — "Russian Federation", "Czechia", "Viet Nam",
   "Korea (South)" — and providers use colloquial ones. Only the disagreements
   are listed; a country whose provider spelling already matches needs no
   entry, which is why the table is ~35 lines rather than 227.
3. **The category name**, case- and whitespace-insensitive.

Matching the loaded categories rather than a hard-coded table has two
properties worth stating:

- **An id is only ever assigned if it exists.** An unrecognised country cannot
  mint a `country_ZZ` assignment that no category renders — `category_display`
  would fall back to printing the raw id.
- **A country the user adds by hand resolves without a code change.**

`map_country_to_region` stays as the fallback for spellings the taxonomy does
not carry ("Holland", "England", "Deutschland"), so nothing that resolved
before stops resolving. Its `country_*` arms are now shadowed by the taxonomy
lookup, which returns the same ids.

Existing assignments are unaffected until an asset is re-enriched;
`replace_auto_taxonomy_assignments` then swaps the `AUTO` row for the
country-level one, and still abandons the taxonomy entirely if any non-`AUTO`
assignment exists, so hand classification is not touched.

## Tests

Four in `auto_classification.rs`, against a trimmed regions seed that keeps the
continent → sub-region → country shape:

- Germany and Portugal resolve to `country_DE` and `country_PT` — the two the
  continent-level mapping made identical — and United States still resolves to
  `country_US`
- "Russia" resolves to `country_RU` against the taxonomy's "Russian
  Federation"; a bare `DE` resolves; case and surrounding whitespace do not
  matter
- a country the taxonomy does not carry falls back to the coarse region rather
  than inventing one, an unknown name resolves to nothing, and `ZZ` does not
  mint `country_ZZ`
- `classify_asset` end to end writes a single `country_DE` row at 10000 bps
  with source `AUTO`

`cargo test -p wealthfolio-core --lib` passes; the three `addons::tests`
failures on this machine reproduce on a clean checkout of `main` and are
unrelated to this change. `cargo clippy -p wealthfolio-core --all-targets` and
`cargo fmt --check` are clean.

## Checklist


afadil:
@rwgs, Could we normalize country names before they reach core? AssetProfile.country already says it should contain an ISO alpha-2 code. Yahoo and Alpha Vantage could normalize their values there, while Finnhub already returns codes. Core would then only validate country_XX against the taxonomy, with a name fallback for legacy data. I wouldn’t put taxonomy IDs in the Yahoo provider.
WDYT?

rwgs:
@afadil Agreed, and pushed as `c53e999d5`.

You're right that the contract was already broken rather than merely unstated. `AssetProfile::country` documents alpha-2, and two of the three providers that fill it send something else — Yahoo passes `summaryProfile.country` straight through as `"United States"`, Alpha Vantage's `OVERVIEW` sends `"USA"`. Only Finnhub was honest, so I've left it alone.

**On the last point directly: no taxonomy IDs go near the providers.** The original commit touched exactly one file (`crates/core/src/assets/auto_classification.rs`), and no `country_*` string exists outside core in either commit. The new helper returns `"DE"`, never `country_DE`.

## What the commit does

`models::country::to_iso_alpha2` resolves a bare code, a taxonomy name, or a colloquial spelling or endonym, insensitive to case and whitespace. Yahoo and Alpha Vantage call it on the way out.

One design decision worth your attention: **the name table is generated from the seeded regions taxonomy** rather than hand-written. So the two agree by construction — a name market-data resolves is a country the app can also place — and they cannot drift apart in a later migration. 227 names, plus 61 aliases for the spellings the taxonomy carries under another name (`"Russia"` for `"Russian Federation"`, `"Czech Republic"` for `"Czechia"`, plus `"Holland"`, `"Scotland"`, and the endonyms a localised provider can return).

It also **removes a duplicate table rather than moving one.** Yahoo says `"United States"` and Alpha Vantage says `"USA"`, so a table per provider would have duplicated immediately. Since `core` already depends on `wealthfolio-market-data`, `resolve_country_category` now calls the same function and its own 49-line alias list is gone.

The classifier's fallbacks stay, and they are still load-bearing, because normalisation only reaches what a provider sends on this enrichment. Profiles already on disk hold names; `convert_profile` synthesises the stored `countries` array from the scalar; hand-entered spellings never pass a provider at all. The taxonomy-name match after the code lookup earns its keep for one case specifically: categories the user added by hand.

## Two calls I'd rather you made than me

1. **An unresolvable name is carried through unchanged rather than becoming `None`.** That keeps anything that classified before classifying, but it means the field is honestly "alpha-2, or a name we could not resolve" — which is what the doc comment now says. If you would rather it be strictly alpha-2 and drop the rest, that is a one-line change and I will make it.
2. **The commit touches core as well as the providers.** Deleting the duplicate in the same change seemed right, but if you would prefer this PR's core diff left as it was for review, say so and I will split it in two.

## Tests

Nine in `models/country.rs`: the three provider spellings of the United States, case and whitespace, the taxonomy-vs-colloquial pairs, endonyms, and the negatives — unknown name, empty string, `ZZ`, and a country the seed omits. Three of those hold the generated tables to what the lookups assume: sorted, no alias shadowing a name, every code real. One in `provider/yahoo` maps a `quoteSummary` fixture end to end. The Alpha Vantage overview test now asserts `"US"` where it asserted `"USA"` — the one deliberate behaviour change.

`cargo test -p wealthfolio-market-data --lib` 280 pass. `-p wealthfolio-core --lib` 1815 pass, with the same three Windows-only `addons::tests` failures that reproduce on a clean `main` (#1542). `cargo clippy --all-targets` and `cargo fmt --check` clean.

## Filed separately

Three things surfaced while doing this, none of which belongs in this PR:

- **#1606** — the regions taxonomy omits 22 ISO 3166-1 countries, including Slovenia, Jersey, Guernsey and the Isle of Man. Relevant here only because it is the reason `"Jersey"` still resolves to nothing: this PR matches the taxonomy's categories, so seeding those 22 makes them work with no further code change.
- **#894** — `RateLimit` has no per-day field, so Alpha Vantage's 25/day free tier is declared as `requests_per_minute: 25` and spent inside the first minute. Added the code evidence to the existing issue.
- **#827** — `wealthfolio/asset-profiles` answers that issue's open question about a free dataset for ETF country weights, via `country_weights[]` with alpha-2 codes.

That last one is also worth raising against your framing here, because it suggests a better shape than what I have implemented. The asset-profiles stock schema carries **both** `country` and `country_code` rather than choosing between them. Applied here that would mean adding `country_code` to `AssetProfile` alongside `country` — providers fill the code, core resolves on it, the display name survives intact — instead of normalising in place as this commit does. I did not do that here because it is a wider change than your comment asked for and it wants your call on the model, but if you prefer it I am happy to redo it that way.

afadil:
Thank you @rwgs, seems good. merging

PR #1512, opened 2026-08-20
feat: add custom date range to spending dashboard

markitosgv:
## Description

Adds a custom date range option to the main Spending dashboard so users can inspect a complete calendar year or any other arbitrary period instead of being limited to named intervals or a single month.

The selected range:

- is stored in `spendingFrom` and `spendingTo` URL parameters;
- survives reloads and can be shared as part of the dashboard URL;
- uses daily, weekly, or monthly chart granularity depending on its duration;
- is compared with the immediately preceding period of the same length;
- is available through the existing responsive date range selector on desktop and mobile.

This also adds strict parsing for the date range query parameters and unit coverage for valid, incomplete, invalid, and reversed ranges.

## Validation

- `pnpm check`
- `NODE_OPTIONS=--no-experimental-webstorage pnpm --filter frontend test --run` (1,454 tests passed)
- `pnpm build`
- Manual verification in the web app with a full-year 2025 range and a shorter custom range

## Checklist


afadil:
@markitosgv thanks for this Pr. Pushed a few adjustments to reuse the existing date picker, improve the mobile layout, and keep custom dates and comparison labels consistent across Insights.

Issue #1459, opened 2026-08-16
Net Worth → Cash drawer fills up with phantom `$0` account rows


kryofin:
## Summary

The Cash drill-down lists one row per account holding cash. On portfolios with
any meaningful history it fills up with rows reading `$0` / `0.0%`, typically
long-closed accounts that were fully liquidated years ago. The accounts hold
nothing; they are listed because their stored cash balance is a residual on the
order of `1e-15` rather than exactly zero.

This spurious balance usually happens when accounts are imported from other
tools (Actual Budget, etc.) and this bug is an annoyance for folks migrating
over to Wealthfolio.

## Reproduce

1. Dashboard → **Net Worth** tab → click the **Cash** breakdown row.
2. The drawer's ITEMS list shows every account with a non-zero stored cash
   balance.

**Expected:** only accounts that actually hold cash.

**Actual:** every account that has ever held cash, most reading `$0`.

Illustrative shape of a portfolio with a dozen accounts, several closed:

| Account (synthetic) | State                    | Stored cash balance  | Drawer row           |
| ------------------- | ------------------------ | -------------------- | -------------------- |
| Checking A          | active                   | `4250.00`            | `$4.3K` ✅           |
| Brokerage B         | active                   | `1.6774119e-12`      | `$0` ❌              |
| Brokerage C         | closed, fully liquidated | `-2.39977309999e-13` | `$0` ❌              |
| Retirement D        | closed, hidden           | `9.2751693e-12`      | `$0` ❌              |
| Retirement E        | closed                   | `0.0`                | _(absent — correct)_ |

Note the last row: accounts whose stored balance is **exactly** `0.0` are
correctly absent. Only the contaminated ones appear. Both active and hidden
accounts are affected identically — **this is not a hidden-account bug** (see
[Hidden vs. archived](#hidden-vs-archived-accounts) below).

---

## Cause 1 — the drawer rounds after the zero check, not before

`NetWorthService::get_net_worth`
(`crates/core/src/portfolio/net_worth/net_worth_service.rs`) skips a cash balance
only when it is _exactly_ zero, and applies `DECIMAL_PRECISION` rounding
**after** that test:

```rust
for (currency, &amount) in &snapshot.cash_balances {
    if amount.is_zero() { continue; }              // <-- exact-zero only
    let cash_base = self.convert_cash_balance_to_base(...);
    let (asset_id, name, market_value_base, category) = (
        format!("CASH:{}:{}", account.id, currency),
[...]

Issue #889, opened 2026-04-25
Suggestion: Use "Amount" instead of calculation to prevent cash balance drift due to rounding errors


lhewett946:
When I input the quantity / price figures provided by my broker, they are pre-rounded to two decimal places. This results in the "Amount" calculated by Weathfolio having some error compared with what is in my brokerage account. Over time, the error accumulates until my cash balance is noticeably out of sync with what is actually in my brokerage account.

I tried to work around this by updating the "Amount" column for my transactions to the amount of cash that was actually added/removed from my cash balance in my brokerage account but it appears that manually editing this column does not have any effect on the calculation of the cash balance.

It would be useful if it did, as otherwise I have to recalculate the figures for my transactions from my broker to avoid creating rounding errors in Wealthfolio.

sylvank:
Hi, I was literally going to post the same thing! I'm new to Wealthfolio (loving it so far) and was just doing imports of old transactions previously saved in Quicken. The stock prices are rounded to the nearest cent, while in reality the actual transaction may have been in the fractional cents: $0.00xx . 

As you say, Wealthfolio seems to ignore the "Amount" field, which is accurate in my imports, going only by Quantity X Price = Cash difference. From the "How to Import CSV" help:

> Amount field: For cash activities (DIVIDEND, DEPOSIT, WITHDRAWAL, TAX, FEE, INTEREST, TRANSFER_IN, TRANSFER OUT), amount is preferred when provided, **otherwise calculated from quantity x unitPrice.**

I think it would make sense to auto-fill Amount if it is not entered or imported; but treat it as the canonical source of truth if it has been imported or manually entered.

This below example isn't ideal, I was just trying to prove that it's the price, not amount, that affects the cash balance:

A transaction with an (example, modified) fractional price:
<img width="760" height="42" alt="Image" src="https://github.com/user-attachments/assets/fe0b1042-ce16-45b6-bd2e-19b1ce9322cb" />

The cash balance:
<img width="588" height="96" alt="Image" src="https://github.com/user-attachments/assets/1e5b86b9-dc00-4e32-8acb-d7e767b17b67" />

Rounding the price, now amount & price are correct. 
<img width="758" height="39" alt="Image" src="https://github.com/user-attachments/assets/156b00f7-59df-4b53-b2c7-0d5e6d72187c" />

Cash balance has been updated by 72 x 0.005 = $0.36. 
[...]

PR #1520, opened 2026-08-20
Skip dust cash balances when building holdings

### LINKED ISSUE #1519: Cash drill-down in Insights lists many $0.00 rows that once held cash
## Summary

On **Insights → Breakdown → Allocation → Cash**, the drawer lists one cash row per account, and most of them render `$0.00`. The accounts are long since liquidated — every dollar was swept out years ago — but they still emit a full cash holding.

This is the same class of defect as #1459, in a different code path. #1459 was fixed in `net_worth_service.rs` by rounding cash balances *before* the zero check; the holdings service, which backs the Insights allocation drill-down, still uses an exact zero comparison.

## Steps to reproduce

1. Have one or more accounts that once held cash and were later fully liquidated.
2. Let the portfolio recalculate a few times (the incremental recalc modes are what accumulate the noise — see *Root cause*).
3. Go to `Insights → Breakdown`, select the **Allocation** lens, click **Cash**.
4. The drawer's holdings list shows a row per account, most of them `$0.00`.

**Expected:** only accounts with an actual cash balance appear.

**Actual:** every account that has ever held cash appears; the settled ones show `$0.00`.

## Minimal reproduction

Three brokerage accounts, snapshot cash balances (illustrative values):

| Account   | `cash_balances["USD"]`      |
| --------- | --------------------------- |
| Account A | `2500.00`                   |
| Account B | `0.000000000000001`         |
| Account C | `-0.0000000000000023679074` |

Expected Cash drill-down: **one** row (Account A, `$2,500.00`). Actual: **three** rows — Account A `$2,500.00`, Account B `$0.00`, Account C `$0.00`.

Querying `POST /api/v1/allocations/holdings/query` with `{"taxonomyId":"asset_classes","categoryId":"CASH"}` on  real portfolio returned many such rows with extremely small carried balances. Two thirds of the list was noise. Category totals and percentages were correct throughout, which is why nothing else looks wrong.

## Root cause

Snapshot `cash_balances` persist as serde-float JSON, and the incremental recalc modes re-seed from that `f64` round-trip, so residual balances on the order of `1e-15` accumulate on settled accounts and compound across runs.

Cash holdings are built from those balances at two sites in `crates/core/src/portfolio/holdings/holdings_service.rs`, both guarded by an exact zero comparison:

- `build_live_holdings_from_snapshot` — `if amount == Decimal::ZERO { continue; }`
- `holdings_from_snapshot` — same guard

`1e-15 != 0`, so a full `Holding` is emitted. Nothing downstream drops it: `compute_holdings_by_allocation_from_holdings` skips a row only when its *matched share* is `<= 0` (the share is 1 for cash — the value is never tested), and `merge_non_cash_detail_rows` deliberately keeps cash rows un-merged so each account gets its own line. The value only collapses to `$0.00` at render time in `allocation-detail-sheet.tsx`.

Negative dust is the same story with the sign flipped, so anywhere a negative cash balance is treated as a credit line, it can emit a phantom `$0` liability row too.

## Suggested fix

Round to `DECIMAL_PRECISION` (8, well below a cent) before the zero test at both sites, matching what #1459 did for net worth. Genuine sub-cent balances survive; no real balance changes.

## Environment

- Wealthfolio 3.7.0, web/server mode

kryofin:
Fixes #1519

Related to #1459 and #889

## Problem

The **Insights → Breakdown → Allocation → Cash** drawer lists a row for every account that has ever held cash, most of them rendering `$0.00`. Accounts fully liquidated years ago keep residual dust balances, which are non-zero and so emit a full cash `Holding` that only collapses to zero at render time.

This is the same defect #1459 fixed for the Net Worth breakdown, in a different code path. `b715b204` rounded before the zero check in `net_worth_service.rs`; the holdings service — which backs the Insights allocation drill-down — was never touched and still compares against exact zero.

## Change

`holdings_service.rs` builds cash holdings from `snapshot.cash_balances` at two sites, both guarded by `if amount == Decimal::ZERO { continue; }`:

- `build_live_holdings_from_snapshot` (the live `get_holdings` path)
- `holdings_from_snapshot` (the historical snapshot viewer path)

Both now call a shared `is_cash_dust` helper that rounds to `DECIMAL_PRECISION` before testing for zero. `DECIMAL_PRECISION` is 8 — well below a cent — so genuine sub-cent balances survive; only float noise is dropped.

Emitted balances are unchanged: the rounding is applied to the *test*, not to the value written into the holding. This is the same invariant `b715b204` held, reached differently — those sites already called `.round_dp` after the check, so the commit could round the emitted value while moving the call earlier. These sites have no pre-existing rounding, so rounding the emitted amount would be a new behavior change.

Cash totals, category values, and allocation percentages are identical before and after — the dust contributed nothing to them. What changes is the row count in the drill-down, and `holdingsCount` in the value strip, which was inflated by the same phantom rows.

Nothing downstream filtered these out: `compute_holdings_by_allocation_from_holdings` skips a row only when its matched *share* is `<= 0` (the share is 1 for cash — the value is never tested), and `merge_non_cash_detail_rows` deliberately keeps cash rows un-merged so each account gets its own line.

The dust itself originates in the storage layer — snapshot `cash_balances` persist as serde-float JSON and the incremental recalc modes re-seed from that `f64` round-trip, so the noise compounds across runs. That remains tracked separately, as noted in #1459.

## Tests

Six new tests in `crates/core/src/portfolio/holdings/holdings_service.rs`:

- `cash_dust_is_detected_below_decimal_precision` — unit-tests the predicate in both signs, and asserts `0.00000001` (exactly `DECIMAL_PRECISION`) is **not** dust.
- `cash_dust_boundary_sits_just_above_half_of_decimal_precision` — pins the actual discard threshold. `round_dp` uses banker's rounding, so the exact midpoint `5e-9` goes to even and is dropped while `5.1e-9` survives; both signs asserted.
- `get_holdings_skips_dust_cash_balances` — positive and negative dust yields no holdings at all.
- `get_holdings_keeps_real_cash_balances_alongside_dust` — a real balance and a genuine sub-cent balance both survive while dust in a third currency is dropped.
- `get_holdings_does_not_round_the_cash_balance_it_emits` — feeds a balance with more than 8 decimal places and asserts `quantity`, `market_value.local`, and `cost_basis.local` all keep full precision, so the unrounded-emit invariant cannot be tidied away unnoticed.
- `holdings_from_snapshot_skips_dust_cash_balances` — same for the snapshot path.

Each was mutation-tested rather than merely observed green. Reverting the fix to `amount == Decimal::ZERO` fails the three behavioral tests (the two predicate unit tests correctly still pass); additionally rounding the emitted amount fails `get_holdings_does_not_round_the_cash_balance_it_emits` with `left: 1234.56789012`, `right: 1234.567890123456789`.
## Verification

Apart from test gates, verified end-to-end against a rebuilt server image and real data

## Related site not changed

`build_cash_currency_split` in `crates/core/src/portfolio/valuation/current_account_valuation.rs` uses the same `!value.is_zero()` pattern. It aggregates per currency across all accounts, so dust only surfaces there if an entire currency consists of nothing but dust. Left alone to keep this diff scoped; worth a follow-up.

### DIFF
--- a/apps/tauri/src/commands/portfolio.rs
+++ b/apps/tauri/src/commands/portfolio.rs
@@ -10,7 +10,7 @@ use crate::{
 };
 
 use chrono::NaiveDate;
-use log::{debug, info, warn};
+use log::{debug, info};
 use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use tauri::{AppHandle, State};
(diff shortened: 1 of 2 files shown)


kryofin:
@afadil Any feedback on this one? It's a small change and mostly test code.

afadil:
@kryofin Thanks! I made a small adjustment to align the desktop and web paths. Merging.

PR #1526, opened 2026-08-21
feat(accounts): add annual interest rate field

DamienKCorp:
Allows users to enter and display an annual interest rate on cash accounts (e.g. French Livret A / LDDS at 1.7%).

### Database
- Migration `2026-08-20-000001`: adds `interest_rate TEXT` column to `accounts` table
- Diesel schema updated accordingly

### Backend (Rust)
- `Account`, `NewAccount`, `AccountUpdate` in `wealthfolio-core` carry `interest_rate: Option<String>`
- `AccountDB` (storage-sqlite) maps the field through all three `From` conversions
- `wealthfolio-server` REST models (`NewAccount`, `AccountUpdate`) and their `From` impls updated
- `wealthfolio-connect` broker sync: `interest_rate: None` added to `NewAccount` initializer

### Frontend (TypeScript / React)
- `Account` type and `newAccountSchema` (Zod) include `interestRate?: string`
- Account form: new "Interest Rate" field (`type="text" inputMode="decimal"`) with `%` suffix
- Account edit modal: `interestRate` included in `defaultValues` so the saved value is restored on re-open
- Account detail page header: badge displays `Interest Rate: X.XX%` when set

### i18n
- `settings.json` key `accounts_form_interest_rate_label` — 7 locales (en/fr/de/es/ja/ko/zh)
- `account.json` key `interest_rate` — 7 locales

## Test plan
- [X] Create a Cash account, enter `1.7` as interest rate → save → badge `Interest Rate: 1.70%` visible on account page
- [X] Re-open the account form → field pre-filled with `1.7`
- [X] Save without touching the field → value preserved
- [X] Clear the field → badge no longer shown on account page
- [X] Accounts without an interest rate unaffected

## Description

image

## Checklist

- [X] I have read and agree to the [Contributor License Agreement](https://github.com/wealthfolio/wealthfolio/blob/main/CLA.md).

### DIFF
--- a/apps/frontend/src/i18n/locales/de/account.json
+++ b/apps/frontend/src/i18n/locales/de/account.json
@@ -153,5 +153,6 @@
   "to_this_account_in": "auf dieses Konto in",
   "your_total_is": "Ihr Gesamtbetrag beträgt",
   "more_count_one": "+{{count}} weitere",
-  "more_count_other": "+{{count}} weitere"
+  "more_count_other": "+{{count}} weitere",
+  "interest_rate": "Zinssatz"
 }
--- a/apps/frontend/src/i18n/locales/de/settings.json
+++ b/apps/frontend/src/i18n/locales/de/settings.json
@@ -82,6 +82,7 @@
   "accounts_form_name_label": "Kontoname",
   "accounts_form_name_placeholder": "Anzeigename des Kontos",
   "accounts_form_group_label": "Kontogruppe",
+  "accounts_form_interest_rate_label": "Zinssatz",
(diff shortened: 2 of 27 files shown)


Jonjon-prog:
Hey @DamienKCorp , if we want to go this way we should be able to set a different rate each year and save them (as for the livret A for example) but currently the best way it's to just put dividend into the cash account at the end of the year

DamienKCorp:
Hi @Jonjon-prog. Fully agree with that. Let's stay on this way by putting dividend into the cash account at the end of the year.

PR #1527, opened 2026-08-21
fix(activity): let a synced needs-review activity be reclassified

rwgs:
## Description

> **Updated 2026-09-01, after #1614.** That PR built the `CREDIT` and
> `ADJUSTMENT` editors this branch had grown, so the branch was re-merged onto it
> and cut back to the part still missing: the `UNKNOWN` reclassification path.
> The Problem section below has been corrected to match current `main`; see
> [What #1614 already fixed](#what-1614-already-fixed) for what was dropped.

## Problem

Broker sync can persist an activity that the UI then refuses to edit, so the
review it was flagged for can never be resolved.

`needs_review()` flags an activity whose provider type is `UNKNOWN`, absent,
low-confidence, or carrying warning reasons, and an absent type is defaulted to
`UNKNOWN` outright:

- `crates/connect/src/broker/mapping.rs:433` —
  `.unwrap_or_else(|| "UNKNOWN".to_string())`
- `crates/connect/src/broker/mapping.rs:48-80` — `needs_review()`
- `crates/connect/src/broker/mapping.rs:624-625` — flagged rows are stored as
  `ActivityStatus::Draft`

Reclassifying such a row is the whole point of the Draft state, but the form
cannot do it. `ACTIVITY_FORM_CONFIG` has entries for twelve types — BUY, SELL,
DEPOSIT, WITHDRAWAL, DIVIDEND, TRANSFER, SPLIT, FEE, INTEREST, CREDIT,
ADJUSTMENT, TAX — and `UNKNOWN` is not among them, so the renderer falls through
to `activity:form_renderer.not_editable` ("This activity type can't be edited
here.",
`apps/frontend/src/pages/activity/components/activity-form-renderer.tsx:41-47`).

The type cannot be changed to escape that, because while editing it is pinned to
the activity's own value and the picker is not rendered at all
(`activity-form.tsx:67` and `:133` on current `main`):

```tsx
const effectiveSelectedType = isEditing ? initialType : selectedType;
...
{!isEditing && !hidePicker && <ActivityTypePicker ... />}
```

So the one field that has to change is the one field that is immutable, and its
current value is what blocks the form. `activityTypeOverride` would be the
obvious escape hatch — it is read by `getEffectiveType()`
(`apps/frontend/src/lib/types.ts:179-180`) — but no UI, adapter, or command
writes it, so it is unreachable from the app.

The consequence is not only a stuck row. `DefaultActivityCompiler` skips
anything not `POSTED` (`crates/core/src/activities/compiler.rs:38-41`), so an
unresolved Draft with cash impact is silently absent from the account's computed
balance. On a `TRANSACTIONS`-mode cash account, a draft that should have been a
withdrawal leaves the balance overstated by its amount, with no way to correct
it.

<a id="what-1614-already-fixed"></a>

### What #1614 already fixed

This PR originally also argued that `CREDIT` and `ADJUSTMENT` rows could not be
edited at all, and carried forms for both. #1614 added those editors, and its
adjustment form is better than the one here — it has the cash/securities mode
toggle this branch was working towards. So the following were dropped in the
re-merge rather than offered again:

- `credit-form.tsx` and `adjustment-form.tsx`, and their `ACTIVITY_FORM_CONFIG`
  entries.
- An `EditableActivityType` alias. #1614 widened `PickerActivityType` to include
  both types, making the alias an exact synonym; callers went back to it.
- The `requires_asset_identity` option-expiry guard in `activities_service.rs`,
  which #1614 implemented identically, and its two create-path tests.
- `CREDIT` as a reclassification-only picker entry — #1614 put it in
  `SECONDARY_ACTIVITY_TYPES`, so it is offered for ordinary creation now. Only
  `ADJUSTMENT` still needs surfacing, and only while reclassifying.

The diff went from 21 files and 1273 insertions to 14 and 573.

## Fix

When editing an activity whose stored type has no form, show the type picker
instead of the dead-end message. Choosing a type renders that form, seeded with
the row's existing values via the config's `getDefaults`, and submitting takes
the existing update path — which already sets `is_user_modified = 1`,
`needs_review = 0`, and defaults status to `POSTED`
(`crates/storage-sqlite/src/activities/model.rs:629-682`). Re-sync then skips
the row (`crates/storage-sqlite/src/activities/repository.rs:2604-2614`), so the
correction sticks.

Behaviour for a type that does have a form is unchanged: it stays pinned and the
picker stays hidden.

The pieces:

- `hasActivityForm()` in `activity-form-config.ts` — asks whether a stored type
  has an editor. Homed next to the config it asks about, and unit-tested.
- `activity-form.tsx` — uses it to decide whether to pin the stored type or
  offer the picker. The type badge now always reports the type the row is stored
  under, so it is clear what is being reclassified. The picked type is scoped to
  the activity's id, so opening a different row starts unpicked rather than
  inheriting the previous row's choice.
- `activity-type-picker.tsx` — an opt-in `includeReclassificationTypes` that
  also offers `ADJUSTMENT`, which has an editor but is deliberately not a way to
  record a new activity.
- The mobile sheet had the same hole (raised in the 2026-08-24 review) and now
  walks the type step before details for such a row, instead of jumping to a
  details form with no type and no way back. Mobile still needs `CREDIT` behind
  the same flag: #1614 did not touch the mobile type step.
- `schemas.ts` — `ADJUSTMENT`'s `assetId` becomes optional, with a `superRefine`
  requiring either a security or a positive amount, so a reclassified cash
  adjustment validates.

No new i18n keys: the badge reuses `activity:activity_type`, and the reclassify
entries reuse existing type labels.

## Testing

- `activity-form-config.test.ts` covers the predicate, including that
  `TRANSFER_IN`/`TRANSFER_OUT` only have a form via their picker alias.
- `activity-form.test.tsx`, `activity-type-picker.test.tsx` and
  `mobile-forms/__tests__/mobile-activity-form.test.tsx` cover the
  reclassification path on both sheets.
- `activities_service_tests.rs` keeps one test,
  `test_update_cash_adjustment_without_asset` — #1614 covers creating a cash
  adjustment but not editing one, which is the flow this PR creates.
- `pnpm test` — 1769 passed (218 files). `pnpm type-check`, `cargo test`,
  `cargo fmt --check`, `eslint` — clean.

## Follow-up, not in this PR

Even with drafts resolvable, an unresolved one still skews a `TRANSACTIONS`-mode
account's balance with nothing saying so where the number is shown.
`cash-audit.ts` only detects balances crossing _negative_, so a balance
overstated by a missing outflow is invisible by construction, and drafts are
surfaced on the Connect sync summary and the activity grid rather than on the
account. Happy to open that separately if it's wanted.

## Checklist


afadil:
@rwgs, Thanks for digging into this. I verified the underlying `UNKNOWN` draft issue is real, and the desktop fix is heading in the right direction.

Could you please address the mobile flow too? Mobile editing still skips type selection for `UNKNOWN`, so the activity remains stuck there.
- `CREDIT` and `ADJUSTMENT` are valid types. The current change forces users to select a different type just to edit the amount or date. They should remain editable without changing their classification.
- Please also reset the selected replacement type when the activity changes. Otherwise, the type chosen for one unsupported row can carry into the next one.
- A couple of component tests covering desktop and mobile reclassification would help lock the behavior down.

Thanks again. Happy to take another look once these are addressed.

rwgs:
@afadil and FYI - the body now says the two new editors mirror "the field sets mobile already shows for those types". That's accurate but it does mean the ADJUSTMENT editor inherits mobile's constraint that a symbol is required (adjustmentActivitySchema enforces it), so a synced ADJUSTMENT with no symbol - a pure cash correction, which classify_import_activity does allow would need a symbol added before it could be saved. 
Pre-existing on mobile, not something this PR makes worse, but the honest answer is that it wants fixing in both places at once rather than only on desktop.

rwgs:
Rebuilt on current `main` — #1614 had built much of what this branch carried, so
rather than resolve the conflicts mechanically I took your version of everything
that overlapped and cut this back to the part that is still missing.

Dropped, because #1614 does them (and in the adjustment form's case, does them
better — the cash/securities mode toggle is what this branch was working
towards):

- `credit-form.tsx`, `adjustment-form.tsx`, and their `ACTIVITY_FORM_CONFIG`
  entries
- an `EditableActivityType` alias — widening `PickerActivityType` to include
  `CREDIT`/`ADJUSTMENT` made it an exact synonym, so callers went back to yours
- the `requires_asset_identity` option-expiry guard, which you implemented
  identically, plus its two create-path tests
- `CREDIT` as a reclassification-only picker entry, now that it is in
  `SECONDARY_ACTIVITY_TYPES`

What is left is the original defect: a synced needs-review row is stored as
`UNKNOWN`, `UNKNOWN` has no entry in `ACTIVITY_FORM_CONFIG`, and the type is
pinned while editing — so the one field that has to change is the one field that
cannot. This adds `hasActivityForm()`, offers the picker when the stored type has
no editor, and does the same on the mobile sheet. Mobile still needs `CREDIT`
behind that flag, since #1614 did not touch the mobile type step.

One Rust test survived the cull: `test_update_cash_adjustment_without_asset`.
#1614 covers *creating* a cash adjustment but not *editing* one, which is
precisely the flow this PR opens up.

21 files and 1273 insertions down to 14 and 573. The description has been updated
to match — the old one claimed `CREDIT` and `ADJUSTMENT` had no forms, which is
no longer true. All checks green.

PR #1528, opened 2026-08-21
fix(import): make amount optional for investment CSVs

### LINKED ISSUE #1523: [Bug] CSV import requires an `amount` column for investment (BUY/SELL) imports, contradicting the docs and the app's own sample CSV
## Summary

In the **Import Activities** wizard, step 2 (Mapping), `amount` is an unconditionally required column mapping for the default (investment) import profile. A CSV of pure BUY/SELL trades that already has `date`, `activityType`, `symbol`, `quantity` and `unitPrice` mapped cannot advance past the Mapping step unless some column is also mapped to `amount` — and because the field is flagged required, the "Ignore" option is suppressed for it, so the requirement cannot be dismissed. The "Review Assets" button stays disabled.

The [CSV import guide](https://wealthfolio.app/docs/guide/csv-import/) says the opposite: `amount` is _"auto-calculated from `quantity × unitPrice` when left blank; **required** for cash activities"_. Every other layer of the codebase agrees with the docs — only the mapping gate does not.

Version: 3.7.0 (built from `18b33c9f5`, all line references below are against that commit)

## Steps to reproduce

1. Select a regular (non-cash) investment account.
2. Import a CSV containing only BUY/SELL rows, with columns for trade date, trade type, symbol, quantity and price — and **no** amount column.
3. On the Mapping step, map all five fields.

**Actual:** the "Fields" status card stays red, `amount` carries a required asterisk, and "Review Assets" is disabled. The `amount` dropdown offers no "Ignore" entry, so there is no way forward other than pointing it at an unrelated column.

**Expected:** per the docs, the import proceeds and `amount` is derived from `quantity × unitPrice`.

## Root cause

`IMPORT_REQUIRED_FIELDS` is a flat list with no per-activity-type conditionality:

```ts
// apps/frontend/src/lib/constants.ts:267-274
export const IMPORT_REQUIRED_FIELDS = [
  ImportFormat.DATE,
  ImportFormat.ACTIVITY_TYPE,
  ImportFormat.SYMBOL,
  ImportFormat.QUANTITY,
  ImportFormat.UNIT_PRICE,
  ImportFormat.AMOUNT,
] as const;
```

`DEFAULT_ACTIVITY_IMPORT_PROFILE` passes it straight through as `requiredMappingFields` (`apps/frontend/src/pages/activity/import/utils/activity-import-profile.ts:116-124`), and `useStepValidation` hard-gates the wizard on it:

```ts
// apps/frontend/src/pages/activity/import/activity-import-page.tsx:155-158
const requiredFieldsMapped = importProfile.requiredMappingFields.every(
  (field) => isFieldMapped(mapping.fieldMappings[field], headers),
);
if (!requiredFieldsMapped) return false;
```

That boolean becomes `canGoNext` → `disabled={!canGoNext || isNextLoading}` (`components/step-navigation.tsx:81`), with a second guard in `handleNext` (`activity-import-page.tsx:455`). Because `isRequired` is true, the skip `SelectItem` is not rendered (`components/mapping-table-cells.tsx:90-100`) — hence no "Ignore" escape hatch.

Worth noting the gate is on **column mapping**, not row values: `isFieldMapped` only tests header presence (`utils/draft-utils.ts:51-55`). A CSV whose `amount` column is entirely blank passes fine; a CSV that omits the column is blocked.

## Everything downstream already treats `amount` as optional for BUY/SELL

- **Row validation** — `utils/draft-utils.ts:490-500`: BUY/SELL requires `symbol`, `quantity`, `unitPrice`; `amount` is never checked. `utils/draft-utils.test.ts` builds BUY and SELL drafts with no amount mapping at all and asserts `status === "valid"`.
- **Portfolio engine** — `crates/core/src/portfolio/snapshot/holdings_calculator/economics.rs:49-80`: `should_use_activity_amount` returns `false` for a non-bond BUY/SELL that has both quantity and unit price, so `gross_trade_amount` computes `qty × price` and ignores the stored `amount` entirely.
- **Shipped sample CSV** — `apps/frontend/src/pages/activity/import/utils/sample-csv.ts` leaves `amount` blank on every BUY/SELL row and populates it only on the cash rows. The app's own starter file would not satisfy the app's own stated rule.
- **In-repo docs** — `docs/activities/activity-types.md:56-57` (BUY) and `:78-79` (SELL) list `amount` under **Optional Fields**; DEPOSIT (`:120`) and WITHDRAWAL (`:138`) list it as required.
- **The in-app help panel contradicts itself** — `apps/frontend/src/i18n/locales/en/activity.json:1239` lists `amount` among the required fields, `:1245` says _"Use amount for cash activities and cash transfers"_, and the example CSV rendered a few lines below it (`import-help.tsx:122-131`) leaves `amount` blank on its BUY rows.

## Relation to #249

[#249](https://github.com/wealthfolio/wealthfolio/issues/249) reported this on v1.1.1 and was closed with:

> Yes, the amount is now required for cash activities to simply and prevent confusion about using quantity and unit price. Doc updated in v1.1.2

That intent is right, but the implementation applies the requirement to **all** activity imports rather than cash ones. The cash and credit-card profiles already declare their own list — `[DATE, ACTIVITY_TYPE, AMOUNT]` at `activity-import-profile.ts:126-145` — which is exactly where a cash-only requirement belongs. The investment default separately inherits `AMOUNT` from the shared constant, which is the part that misfires.

For history: the required list was five fields with no `AMOUNT` from `779166089` (2024-10-23, _"optional import fields"_, which explicitly made amount skippable). `AMOUNT` was added by `327a6fc02` (_"Improve CSV import Experience"_, 2025-04-02) — a 54-file wizard rewrite with an empty commit body — and shipped in v1.1.0. The list has been unchanged since.

Possibly related: [#1165](https://github.com/wealthfolio/wealthfolio/issues/1165) (blank `unitPrice`/`quantity`/`amount` columns rejected in the Schwab template).

## Suggested fix

Drop `ImportFormat.AMOUNT` from `IMPORT_REQUIRED_FIELDS` (`constants.ts:267`). Since `CASH_TRANSACTION_IMPORT_PROFILE` and `CREDIT_CARD_TRANSACTION_IMPORT_PROFILE` already declare `[DATE, ACTIVITY_TYPE, AMOUNT]` independently, `amount` stays required precisely where it is genuinely needed, and investment imports fall back to the documented `quantity × unitPrice` derivation.

The `activitiesRequiredFields` help string would need the same correction across the seven locales (`i18n/locales/*/activity.json`).

jay-tau:
## Description

Investment CSVs with valid BUY/SELL rows are blocked at Mapping when they have no Amount column, even if Quantity and Unit Price are supplied.

The [CSV import guide's amount rules](https://wealthfolio.app/docs/guide/csv-import/#how-amounts-are-imported) describe calculating missing BUY/SELL totals from the trade details, and the [BUY/SELL activity reference](https://github.com/wealthfolio/wealthfolio/blob/main/docs/activities/activity-types.md#trading-activities) also permits an omitted amount to be calculated. That documented approach is correct: these trades already provide the information needed to calculate their value. The unconditional mapping requirement prevents files without an Amount column from reaching the existing validation and calculation logic.

The CSV guide's required-column table still lists Amount; this PR aligns the mapping gate with its activity-specific trade rules so those files can use the documented behavior.

- make `amount` optional at the mapping stage for investment imports, where quantity and unit price are already required
- keep `amount` required for cash and credit-card transaction imports
- leave activity-specific requirements to the existing row validation in Review

Fixes #1523

## Why this works

Import mapping requirements are profile-specific. The investment profile already requires `date`, `activityType`, `symbol`, `quantity`, and `unitPrice`, so removing `amount` only stops the mapping step from demanding a redundant column for BUY and SELL rows. Those rows are validated from positive quantity and unit price values and do not require `amount`.

This does not make Amount universally optional. Review still validates every parsed row according to its activity type, so rows such as regular dividends or splits fail with an Amount error when their required value is missing. Cash and credit-card profiles also keep their separate required fields of `date`, `activityType`, and `amount`, so their mapping behavior is unchanged.

### Automated checks

- `pnpm --filter frontend test --run src/pages/activity/import/utils/activity-import-profile.test.ts src/pages/activity/import/utils/draft-utils.test.ts src/pages/activity/import/components/mapping-table.test.tsx`
- `pnpm type-check`
- targeted ESLint
- `pnpm --filter frontend build`
- `pnpm --filter frontend build:tauri`

### CSV fixture validation

Tested six synthetic CSV files in an isolated web instance with a fresh database and separate Securities, Cash, and Credit Card accounts. All six files were also parsed successfully through the web import API.

- BUY and SELL rows with quantity and unit price but no Amount column passed Mapping and Review.
- A mixed investment file without Amount passed Mapping, while Review correctly rejected a regular dividend and a split that require Amount.
- Regular dividend and split rows with explicit Amount values passed Review.
- Cash and credit-card files kept Amount as a required mapping and accepted their profile-specific activity aliases.
- A transaction file without an Amount column was blocked at Mapping for Cash and Credit Card accounts.

## Checklist

### DIFF
--- a/apps/frontend/src/lib/constants.ts
+++ b/apps/frontend/src/lib/constants.ts
@@ -270,10 +270,7 @@ export const IMPORT_REQUIRED_FIELDS = [
   ImportFormat.SYMBOL,
   ImportFormat.QUANTITY,
   ImportFormat.UNIT_PRICE,
-  ImportFormat.AMOUNT,
 ] as const;
-
-export type ImportRequiredField = (typeof IMPORT_REQUIRED_FIELDS)[number];
 
 export const ExportDataType = {
   ACCOUNTS: "accounts",
--- a/apps/frontend/src/lib/types.ts
+++ b/apps/frontend/src/lib/types.ts
@@ -48,7 +48,7 @@ export {
 
 export type { HoldingCategoryFilterId } from "./constants";
 
-export type { ActivitySubtype, ImportRequiredField } from "./constants";
+export type { ActivitySubtype } from "./constants";
 
 export interface PortfolioWithAccounts {
   id: string;
--- a/apps/frontend/src/pages/activity/import/components/mapping-table-cells.tsx
+++ b/apps/frontend/src/pages/activity/import/components/mapping-table-cells.tsx
@@ -8,7 +8,6 @@ import {
   CsvRowData,
   ImportFormat,
   ImportMappingData,
-  ImportRequiredField,
   type SymbolSearchResult,
 } from "@/lib/types";
 import { cn } from "@/lib/utils";
@@ -56,7 +55,7 @@ export function MappingHeaderCell({
   const displayHeader = Array.isArray(mappedHeader) ? mappedHeader[0] : mappedHeader;
   const isMapped = displayHeader ? headers.includes(displayHeader) : false;
   const isEditing = editingHeader === field || !isMapped;
-  const isRequired = requiredFields.includes(field as ImportRequiredField);
+  const isRequired = requiredFields.includes(field);
 
   return (
     <div>
--- a/apps/frontend/src/pages/activity/import/utils/activity-import-profile.test.ts
+++ b/apps/frontend/src/pages/activity/import/utils/activity-import-profile.test.ts
@@ -18,6 +18,7 @@ describe("activity import profiles", () => {
     expect(profile).toBe(DEFAULT_ACTIVITY_IMPORT_PROFILE);
     expect(profile.assetResolutionEnabled).toBe(true);
     expect(profile.requiredMappingFields).toContain(ImportFormat.SYMBOL);
+    expect(profile.requiredMappingFields).not.toContain(ImportFormat.AMOUNT);
   });
 
   it("uses a transaction profile for cash and credit card accounts", () => {


jay-tau:
@afadil I was hoping to stack more PRs on top of this change. Even if you do not merge it immediately, I would really appreciate it if you could review it, so I can start working on the rest of the PRs. :)

PR #1539, opened 2026-08-23
feat(goals): let a pension fund be drawn down until it runs out

rwgs:
## Description

A "Fund" (defined-contribution) income stream cannot run out.

- `step_plan_pension_funds` (`crates/core/src/planning/retirement/engine.rs`)
  grows the balance while `age < start_age`, then sets it to `0.0` at and after
  the start age.
- The projection loop's `pension_assets` sum filters to funds with
  `age < stream.start_age`, so the pot also leaves reported assets at that age.
  Asserted today by `pension_assets_stop_showing_once_dc_payout_starts`.
- `resolve_plan_dc_payouts` computes one monthly figure — the projected balance
  at `start_age` times the payout rate — and `plan_income_at_age` pays it every
  year to the horizon. `RetirementIncomeStream` has a `start_age` and no end
  age, and nothing caps cumulative payments against a balance.

So the payout rate is an annuity conversion factor, not an interest draw. That
is correct for a bought annuity: the pot genuinely is gone, and a level rate
above the fund's own return is funded by capital consumption and mortality
credits. It is wrong for self-managed drawdown, where an 8%/yr draw on a fund
earning 5% is currently paid in full forever. The same 8% describes two
different plans and nothing records which was meant.

The portfolio is the only depletable pot in the plan today — the withdrawal
engine draws from `buckets`, and `failure_age` and `spending_shortfall_age`
(`crates/core/src/planning/retirement/dto.rs`) report the consequence. That draw
is need-based, never a percentage of a named pot, so "rate-based" and
"depletable" are currently mutually exclusive.

### What this adds

An optional `payoutMode` per income stream: `annuity` or `drawdown`.

**`annuity`** is the default and is today's behaviour, unchanged. The field is
optional and an absent value deserialises to `annuity`, so saved plans project
bit-identically. Two tests guard that: one deserialises a stream serialised
without the field, and one compares a plan with the field absent against the
same plan set explicitly to `annuity`, asserting equal income and portfolio
paths year by year.

**`drawdown`** keeps the balance through the payout phase:

1. The first withdrawal is the rate times the balance at the start age, as now.
2. Each later year escalates by the stream's existing growth setting —
   `annualGrowthRate`, else inflation when `adjustForInflation`, else flat.
3. The remaining balance earns a payout-phase return, withdrawal first:
   `balance = (balance - withdrawal) * (1 + postPayoutReturn)`.
4. A withdrawal is capped at the remaining balance. Once the balance is gone the
   income stops instead of being paid out of nothing.

`pension_assets` keeps counting a drawdown fund while it pays, since it is still
an asset. The existing assertion stays true for `annuity` and now has a
`drawdown` counterpart.

A fixed escalating withdrawal is what depletes, and depletion is the point. A
percent-of-remaining-balance draw never runs out — the income falls instead — so
it is a separate idea and is deliberately not included here.

### Payout-phase return

There was no such assumption to reuse. `accumulationReturn` is labelled "Fund
return before payout" in the sidebar and stops applying at the start age, and
the plan's retirement return belongs to the portfolio. This defaults to the
plan's retirement return, net of fees, with an optional per-stream
`postPayoutReturn` override. Reusing `accumulationReturn` would contradict its
own label.

### Where the depletion state lives

Depletion depends on what earlier years already withdrew, so it lives in a
`DrawdownLedger` belonging to a single projected path, threaded through all four
of them: the deterministic projection, each Monte Carlo simulation, each SORR
scenario, and each required-capital sizing trial. Without the last three a
drawdown fund would still pay forever inside `run_monte_carlo`, `run_sorr` and
`retirement_feasible_from_capital` — overstating success rates and undersizing
the target against income the fund cannot actually pay that long.

### Reporting

A per-stream exhausted age is reported on the projection and on
`RetirementOverview`, beside `failureAge` and `spendingShortfallAge`, and the
dashboard readiness line names the earliest fund to run out. It sits behind
portfolio depletion and the spending gap, which are worse news. A fund that
empties at 79 should be visible on the dashboard, not only in the year-by-year
table.

No special handling is needed for the shortfall itself: expenses continue, the
income line disappears, and the portfolio either covers the gap or trips
`spending_shortfall_age`. The existing machinery reports the consequence.

### Validation

`validate_retirement_plan` now rejects a `drawdown` fund with neither a balance
nor contributions — it would be a stream that pays nothing — and bounds
`postPayoutReturn` by the existing `RETIREMENT_MIN/MAX_ANNUAL_RETURN` limits.

### Incidental tidying

`plan_income_at_age` and `plan_income_at_age_stochastic` had duplicated the
per-stream growth maths; they now share it. The fund's future-value formula is
extracted from `resolve_plan_dc_payouts` so the payout and the opening drawdown
balance are computed from one place and cannot drift apart.

The sidebar preview needs no drawdown maths of its own: the first monthly payout
is the rate times the balance at the start age in both modes, and the dashboard
coverage chart already prefers the engine's `annualIncome` for retirement years,
so the depleting years come from the engine rather than from a copy of it.

### Testing

`cargo test -p wealthfolio-core` (1783 pass; 3 pre-existing `addons` packaging
failures unrelated to this change, failing identically on a clean checkout),
`cargo test -p wealthfolio-server -p wealthfolio-app`, `cargo clippy -p
wealthfolio-core --all-targets` (clean), `cargo fmt --check`, plus
`pnpm test` (1501), `pnpm type-check` and `pnpm lint:quiet`.

New coverage: a drawdown at a rate above its payout-phase return empties at a
stated age with the final withdrawal capped at what is left and zero income
after; a drawdown below that return never empties; an indexed drawdown empties
strictly earlier than a fixed one at the same rate; an annuity at the same rate
still pays at the horizon; `pension_assets` tracks a drawdown fund through its
payout phase; the readiness line names the earliest fund and yields to portfolio
depletion.

## Checklist


afadil:
@rwgs thanks for this PR, found a few correctness issues that I think we should address before merging:

- **Drawdown starts too late:** The ledger is only initialized when the portfolio enters retirement. If a fund’s `startAge` is 65 and retirement begins at 70, the fund disappears from `pensionAssets` at 65, makes no withdrawals for five years, then reappears with its full balance at 70. Please start and advance the ledger from the stream’s `startAge`, or prevent this configuration.

- **The payout return changes when the control is touched:** The UI displays the gross retirement return, but the backend default uses the return after fees. For example, the UI can show 3.37% while the engine uses 2.77%. Saving the displayed 3.37% as an override then silently increases the projection. Please make the UI and backend use the same value.

- **Pension assets can jump at payout age:** Before `startAge`, `pensionAssets` uses a hard-coded 4% fallback and annual contribution timing. At `startAge`, it switches to the drawdown ledger, which uses the configured plan return and monthly contribution timing. This can create a large unexplained jump in the chart. Please use one consistent balance calculation across both phases.

- **Tests:** Please add regression coverage for a drawdown starting before retirement, the net-of-fees return fallback, and balance continuity at `startAge`.

rwgs:
Thanks — all four are addressed in `fix(goals): open a drawdown fund at its own start age, from one balance`. I also merged `main` in: #1541 landed while this was in review and overlaps with the third point, and the branch was conflicting on it.

**Drawdown starts too late.** The ledger is now opened per fund, the first year that fund is due to pay, and drawn from every year after — retired or not. A withdrawal taken before retirement funds no modelled spending, because the plan models none yet; that is already true of every income stream whose `startAge` precedes retirement. But it leaves the fund, which is the point, and it is reported as that year's `annualIncome` so the chart accounts for where the money went.

The two paths that model only retirement years — SORR and required-capital sizing — catch their ledger up to the retirement age first, so a fund that has been paying since 65 arrives part-drawn. Sizing at 70 against a 100k fund drawing 12%/yr from 65 now needs 590k, against 551k if the pot were still whole at 70 and 625k with no fund at all.

I did not take the "prevent this configuration" option: validation can only see `targetRetirementAge`, and in FIRE mode actual retirement can land years later than that or never arrive, so the configuration it forbade would still occur.

**The payout return changes when the control is touched.** The control now reads `payoutPhaseReturn`, which mirrors the engine's fallback the way #1541's `planAccumulationReturn` mirrors the accumulation one. With no override it displays the retirement return net of fees — 2.77%, not 3.37%, on the plan in `phase_returns_subtract_fee` — so saving what is displayed changes nothing.

**Pension assets can jump at payout age.** #1541 settled the 4% fallback half of this. The half left was contribution timing: `step_plan_pension_funds` treated a year of monthly contributions as a year-end lump sum, while `projected_dc_balance_at_start` gives those same contributions their in-year growth. The stepper is now one year of that formula, so the balance a fund reaches at its start age is the balance its payout is a percentage of.

As with #1541, this changes the reported pre-payout pension-asset line for existing plans, annuity ones included — it is the last input to that line that still disagreed with the payout basis. Income and portfolio paths are untouched: `pension_balances` only ever fed that one reported figure.

**Tests.** Three in the engine, one in the sidebar's maths:

- `a_drawdown_starting_before_retirement_is_drawn_from_its_start_age` — starts at 65, retires at 70: holds 100k at 65, 88k at 66 and 40k at 70, pays 12k at 66, and empties at 73 instead of 78.
- `the_default_payout_phase_return_is_net_of_fees` — an absent `postPayoutReturn` projects identically to `Some(0.0337 - 0.006)` and differently from `Some(0.0337)`.
- `a_fund_balance_is_continuous_at_its_start_age` — the balance at the start age is the year before's grown on and contributed to, and the first payout is the payout rate applied to that same balance.
- `dashboard-math.test.ts` pins the sidebar's fallback to that same net figure.

`cargo test -p wealthfolio-core` (1787 pass; the same 3 `addons` packaging failures as a clean checkout), `cargo clippy -p wealthfolio-core --all-targets`, `cargo fmt --check`, plus `pnpm test` (1507), `pnpm type-check` and `pnpm lint:quiet`.

PR #1548, opened 2026-08-24
feat(addon-sdk): expose spend categorization to addons

blastik:
## Description

Adds `ctx.api.spending` (`SpendingAPI`) to the addon SDK with `isEnabled()`, `getCategories()`, `getRules()`, `saveRule()`, `deleteRule()`, and `rerunRules()`. Addons can use Wealthfolio's existing spend-category taxonomy and reusable categorization-rule engine instead of assigning one-off labels to individual activities.

The motivating use case is an importer addon that already recognizes recurring transaction patterns and lets users associate each pattern with an expense, income, or savings category. Saving the association as a rule keeps future imports categorized consistently.

### What's new

- SDK types for spend-category kinds/categories and categorization rule inputs/results.
- A medium-risk `spending` addon permission covering all six public methods, including permission-dialog labels and backend manifest detection.
- Shared desktop/web adapters for spending settings, taxonomies, categorization rules, and reruns.
- A guarded `ctx.api.spending` bridge that maps expense, income, and savings taxonomies into one addon-facing shape.
- Deterministic addon-scoped rule IDs derived from `(addonId, ruleKey)`, so repeated saves update the same rule and separate addons cannot collide.
- An atomic `upsert_categorization_rule` path across Tauri, Axum, the spending service, and SQLite, avoiding list-then-create races.
- Rule listing remains available while Spending is disabled, allowing addons to manage stored rules before the user opts an account into Spending. `rerunRules()` remains a no-op until Spending is enabled.
- Runtime validation rejects unsupported category kinds before a malformed rule can reach storage.
- API reference documentation and an unreleased changelog entry.

### Idempotent rule saves

`saveRule()` derives a stable ID from the addon ID and the addon's `ruleKey`, then calls the atomic backend upsert. The SQLite writer performs the create-or-update decision within its serialized transaction, so concurrent saves for the same rule cannot create duplicates.

### Safety

- `rerunRules()` defaults to `onlyUncategorized: true`, preserving existing manual or AI assignments unless an addon explicitly requests otherwise.
- Addons can only address their own deterministic rule IDs through `saveRule()` and `deleteRule()`.
- The bridge validates category kind values at runtime in addition to the TypeScript type contract.

### Testing

- `pnpm --filter @wealthfolio/addon-sdk type-check`
- `pnpm build:types`
- `pnpm --filter frontend exec vitest run src/addons/type-bridge.test.ts src/addons/spending-rule-key.test.ts src/addons/iframe/addon-iframe-manager.test.ts src/adapters/adapter-command-parity.test.ts` — 77 tests passed
- `cargo test -p wealthfolio-core test_detect_addon_permissions_spending`
- `cargo check -p wealthfolio-server -p wealthfolio-app`
- `pnpm exec prettier --check apps/frontend/src/addons/type-bridge.ts apps/frontend/src/addons/type-bridge.test.ts`

A live Tauri/web session was not run.

I'm the author of the trade-republic-importer-addon (community addon) that motivated this — happy to adjust the API shape if you'd rather see something different.

## Checklist


afadil:
@blastik, Thanks for this useful addition. A few things before merge; two are blockers, and I'd settle the API shape first since it's a public contract once an addon ships against it.

### Blockers

**1. `spending.*` isn't in the sandbox allowlist, so no addon can call this.** `addon-iframe-manager.ts:313` gates brokered calls against `ALLOWED_API_METHODS`; `exchangeRates.*` is at 345–348, `spending.*` is missing, so every call throws `Unknown addon host API method`. The new tests hit `createSDKHostAPIBridge` directly and bypass the broker, so CI stays green. Worth a drift test that walks the bridge and asserts each `namespace.fn` is allowlisted — nothing enforces that today.

**2. The create-vs-update probe is wrong on a default install.** `list_categorization_rules` returns `Ok(Vec::new())` when Spending is disabled (`apps/tauri/src/commands/spending.rs:276`) and `spending.enabled` defaults to false (`crates/spending/src/settings/service.rs:34`), but `create`/`delete` aren't gated. So with Spending off: first upsert creates; second with the same `ruleKey` takes the create branch again and hits a PK collision on the plain `insert_into` (`categorization_rules.rs:300`); delete silently no-ops forever, so the addon can never remove its own rules. So the "harmless duplicate rule" note is off on both counts — default state rather than a race, error rather than a duplicate.

Cleanest fix is an `upsert_categorization_rule` command — cheap here because all writes go through one serialized writer actor (`write_actor.rs:28`), so get-then-insert-or-update inside a single `exec_tx` is atomic for free, killing the list dependency and the race together. I know that breaks the "no `crates/*`" framing; I think it's worth it. The no-Rust alternative is update-first / create-on-NotFound, but the error strings differ between Tauri and Axum so discriminating them is fragile. Either way `deleteCategorizationRule` should just call through — the repo delete already no-ops on a missing row (`categorization_rules.rs:584`), and the probe is the only reason it breaks.

### Bugs

**3. `rerunCategorizationRules()` with no argument throws on desktop.** The Tauri command takes a bare `only_uncategorized: bool` with no serde default (`spending.rs:331`) and `JSON.stringify` drops the undefined key; the Axum handler has `#[serde(default)]` (`api/spending.rs:322`), so web works and desktop fails. Fix at the bridge — adding `#[serde(default)]` to the command would silently select `false`, the mode that overwrites existing rule/ai/import assignments. Default to `true`.

**4. Two of three activity taxonomies are unreachable.** The migration seeds `spending_categories`, `income_sources`, `savings_categories` (`up.sql:305-315`) and Wealthfolio's own rule form offers all three (`rule-form.tsx:146`). This API defaults to the first and documents only it, so an importer handling deposits and savings transfers can realistically only categorize expenses.

**5. `taxonomyId` is broader than its consent label.** Any id passes straight to `get_taxonomy`, so granting `spending` also grants "read every asset-classification taxonomy" behind a dialog saying "View your spend categories." Fixed by #4's shape change.

**6. `adapters/shared/spending.ts` duplicates `features/spending/adapters/rules.ts`.** Same five wrappers and signatures; only log strings differ, and `shared/platform.ts` is a bare re-export of `#platform` so even `invoke`/`logger` are the same binding. Suggest keeping the new file canonical and collapsing the five in `rules.ts` to `export { ... } from "@/adapters/shared/spending"` (presets stay). Its one consumer needs no change.

### Shape

Against existing SDK convention: every method repeats its namespace noun (it's `accounts.getAll`, never `contributionLimits.getContributionLimits`); `save` is the house word for create-or-update, not `upsert`; `upsertCategorizationRule` returns `void` and discards the `CategorizationRule` the command already returns, unlike every other create/update; and `activityType?: string` throws away `ActivityType` from `data-types.ts:24`. The surface is also write-only — no way to list rules, and `CategorizationRule` isn't in the SDK, so an addon can't reconcile after a user deletes a rule in the UI.

```ts
export type SpendCategoryKind = 'expense' | 'income' | 'saving';

export interface SpendCategory {
  kind: SpendCategoryKind;
  taxonomyId: string;
  categoryId: string;
  key: string;
  name: string;
  /** Display breadcrumb, e.g. "Food & Dining / Groceries" */
  path: string;
}

export interface CategorizationRuleInput {
  /** Stable key you choose and reuse; scoped to your addon id. */
  ruleKey: string;
  name: string;
  pattern: string;
  matchType?: CategorizationRuleMatchType;   // default "contains"
  kind: SpendCategoryKind;
  categoryId: string;
  activityType?: ActivityType;
  accountId?: string;
  priority?: number;
}

export interface SpendingAPI {
  isEnabled(): Promise<boolean>;
  getCategories(kind?: SpendCategoryKind): Promise<SpendCategory[]>;   // omit → all three
  getRules(): Promise<CategorizationRule[]>;                            // this addon's rules
  saveRule(rule: CategorizationRuleInput): Promise<CategorizationRule>;
  deleteRule(ruleKey: string): Promise<void>;
  rerunRules(onlyUncategorized?: boolean): Promise<number>;             // default true
}
```

`kind` as a closed union instead of a raw `taxonomyId` fixes 4 and 5 together, all three taxonomies discoverable, asset taxonomies unreachable, honest permission label, and a fourth kind stays non-breaking. It matches the app's own vocabulary too (`quick-categorize-popover.tsx` uses `"expense" | "saving" | "income"`). `isEnabled()` needs no backend work (`get_spending_settings` is on both targets) and closes the silent-degradation gap, since today `rerunRules` returning `0` is indistinguishable from "nothing matched."

Renaming also touches `functions` in `permissions.ts` and the labels in `addon-function-names.ts`.

Thanks again.

blastik:
@afadil thanks for the thorough read — pushed a round addressing all of it.

Both blockers: `spending.*` is now in the iframe allowlist (plus a test that walks the bridge and fails if a namespace method isn't allowlisted, so this can't silently drift again). And `saveRule` now goes through a real `upsert_categorization_rule` command that does the get-then-insert-or-update inside one transaction on the writer actor, so it's race-free instead of relying on a `list()` that's empty when Spending's off. `deleteRule` just calls through now too, since the repo already no-ops on a missing id.

Took your API shape wholesale — `isEnabled`/`getCategories`/`getRules`/`saveRule`/`deleteRule`/`rerunRules`, `kind` instead of `taxonomyId`, `saveRule` returns the rule. That fixes the other two taxonomies and the permission-label mismatch in one move. `rerunRules()` now defaults to `true` and always sends an explicit bool (was the desktop crash). And `adapters/shared/spending.ts` is canonical now — `features/spending/adapters/rules.ts` just re-exports the five it duplicated.

Let me know if anything's off.

PR #1587, opened 2026-08-29
Add presets for some Romanian spendings

steprp:
## Description
Adds a Romanian(ro) categorization rule preset

## Checklist

- [X] I have read and agree to the


afadil (changes requested):
@steprp Thanks for adding the Romanian preset! A few things need fixing before merge:

- Change the language from `sv-SE` to `ro-RO`.
- Add `ro` to the bundled preset validation test.
- Update the Mega Image regex. It does not match `MEGA IMAGE` with a space.
- Review the airline rule. It still contains Swedish airlines such as `BRA Sverigeflyg`.
- Double-check categories like Altex and Apple billing.
- Please add a few real, anonymized Romanian transaction examples as tests.

It looks like this started from the Swedish preset, but some Swedish values were left behind. Please review the full file for similar leftovers.

PR #1533, opened 2026-08-22
feat(i18n): add Brazilian Portuguese (pt) language support

DamienKCorp:
fixes https://github.com/wealthfolio/wealthfolio/issues/1522

Adds Portuguese (pt-BR) as the 8th supported language in Wealthfolio.

### Frontend
- `locales.ts`: registered `{ code: "pt", label: "Português" }` after Español
- `i18next.config.ts`: added `"pt"` to the locales array
- `locales/pt/`: 20 new translation files (account, activity, ai, allocation, asset, auth, common, connect, dashboard, goals, health, holdings, income, insights, onboarding, performance, settings, spending, sync, ui)

### Backend
- `settings_service.rs`: added `"pt"` to `SUPPORTED_UI_LANGUAGES` so the backend accepts and persists the language preference

## Validation
- `pnpm format:check` ✅ — Prettier clean across all 20 JSON files
- `pnpm type-check` ✅ — `LocaleCode` union type updated automatically via `as const`
- `cargo test -p wealthfolio-core -- settings` ✅ — all settings validation tests pass
- `cargo clippy -p wealthfolio-core` ✅ — no warnings

## Checklist

- [X] I have read and agree to the [Contributor License Agreement](https://github.com/wealthfolio/wealthfolio/blob/main/CLA.md).

By submitting this PR, I agree to the [CLA](https://github.com/wealthfolio/wealthfolio/blob/main/CLA.md).

PR #1605, opened 2026-08-31
fix(i18n): complete and correct the `many` plural forms

afadil:
Follow-up to #1533. Two fixes to the CLDR `many` plural category, touching all four locales that have one.

`fr`, `es`, `pt` and `it` all have a `many` category that fires at exact non-zero millions. It is a narrow slot, not a magnitude: 999,999 and 1,500,000 both take `other`.

## 1. `fr` and `es` had no `_many` forms at all

Not for any of their 294 plural stems. Where a form is missing, i18next resolves nothing and renders the raw key:

```
fr  count=1000000  ->  "globalEvents.priceUpdateFailed"     <- raw key
es  count=1000000  ->  "globalEvents.priceUpdateFailed"     <- raw key
pt  count=1000000  ->  "Falha na atualização do preço de 1000000 ativos"
it  count=1000000  ->  "Aggiornamento del prezzo non riuscito per 1000000 asset"
```

Filled in from `_other`, as `pt` and `it` already do.

## 2. Seven keys per locale had a `_many` that should never have existed

Caught in review on this PR. Some keys end in `_other` because their value is the word "Other", not because they are plurals:

```
account.json  type_other                   = "Other"
common.json   component.account_type_other = "Other Accounts"
goals.json    sidebar.spending.add_other   = "+ Other"
```

None has a `_one` sibling and nothing passes them a count, so the twins were unreachable keys that do not exist in English. The same seven were already on main in `pt` and `it` from the two earlier locale PRs, so all four are cleaned here rather than leaving half of it for a third PR.

| locale | CLDR categories | `_many` before | after |
| --- | --- | --- | --- |
| en, de | one, other | n/a | n/a |
| zh, ja, ko | other | n/a | n/a |
| pt, it | one, **many**, other | 301, of which 7 bogus | **294** |
| fr, es | one, **many**, other | **0** | **294** |

294 `_many` forms for exactly 294 real plural stems, in all four.

## The rule

A stem is a plural **only when English has both `_one` and `_other`**. Matching either half alone is wrong in opposite directions:

- `_other` alone catches the seven "Other" enum labels above
- `_one` alone catches 11 keys like `form.err_amount_gt_zero` ("must be greater than 0")

The test used this rule; the script that generated the forms did not. That mismatch is what the review found.

## Why it went unnoticed

The old guard enumerated locales by hand and checked a single key. When Italian shipped, `it` and `pt` were added to the list and `fr`/`es` were not. Its millions assertion was `.not.toBe("")`, but the failure mode is the raw key rather than an empty string, so `fr`/`es` passed anyway. It also only looked for **missing** forms, which is why it stayed green while the generator invented 28 keys.

## Tests

Two checks over every locale folder, globbed so a new locale is covered the moment it lands:

- **coverage** — each locale defines every category its own `Intl.PluralRules` can select
- **no invented forms** — no locale defines a plural-category key that English lacks and whose stem is not an English plural stem

Both verified to fail when the condition they guard is reintroduced, including a `_few` key on `de`, which has no `few` category at all.

The existing `CLDR many category` block is kept: it exercises i18next resolution at runtime, while these two check key presence statically.

## Verification

```
prettier      clean
tsc --noEmit  0 errors
eslint        0 errors, 0 warnings
vitest        205 files, 1647 passed
```

Per-locale accounting against main. The `changed` column confirms no existing translation was touched:

```
locale   added   removed   changed
fr         294         0         0
es         294         0         0
pt           0         7         0
it           0         7         0
others       0         0         0
```

## Not included

The preposition these languages use before a noun at round millions ("1 000 000 **d'**actifs", "1.000.000 **de** activos", "1.000.000 **de** ativos"). With the count rendered in digits the shorter form is idiomatic enough, and `pt`/`it` shipped that way, so all four stay consistent. Adding it is a separate pass over all four locales.


afadil:
Good catch, thanks. Fixed in d52e713.

You were right that `type_other` is an enum label, not a plural. My generator matched on the `_other` suffix alone, while the test I wrote in the same PR used the correct rule (a stem is a plural only when English has both `_one` and `_other`). I applied that rule in one place and not the other.

Removed the 7 per locale. I also cleaned the same 7 in `pt` and `it`, which were already on main from the two earlier locale PRs, so all four are consistent rather than leaving half of it for another PR. Every `many` language now has 294 `_many` forms for exactly 294 real plural stems.

Also added the mirror assertion, which is the actual gap here: the existing check only looked for *missing* forms, so it stayed green while the run invented 28 keys. The new one fails on any plural-category key a locale defines that English doesn't have and whose stem isn't an English plural stem. I checked both directions fail when the thing they guard is reintroduced.

One note on the rule, since it bites both ways: matching `_one` alone is wrong too, because `en` has 11 keys like `form.err_amount_gt_zero" ("must be greater than 0") that end in a category word by coincidence. Requiring both suffixes is what separates the real plurals from the lookalikes.

PR #1612, opened 2026-09-01
feat(allocation): calculated rebalancing worksheet (M1)

Jonjon-prog:
Opening this early, as a draft. The design in #1565 is settled, but two things surfaced while building the calculation that I would rather have your eye on now than after the UI is wired on top of them. The second one in particular is a gap in the design rather than in the code.

This is M1 from `docs/features/allocations/self-directed-rebalancing-design.md` §12: the complete calculated worksheet. It ships as one PR, kept reviewable through the commit order agreed in the thread — core calculation and limits first, then prefill and account allocation, then copy and locales, then export. Only the first of those four is here so far.

## What is in this draft

The arithmetic that turns a target into a set of adjustments, as a pure module. No repositories, no clock, no currency conversion: the service resolves the drift report, the taxonomy contributions, prices and constraints, and hands them over as plain structs. That keeps the §12 verification criteria as unit tests rather than mock-heavy orchestration tests.

| Commit                               | Design           | What it does                                                                                                                                                   |
| ------------------------------------ | ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `spread category gaps`               | §4.2, §4.4, §4.5 | Gaps against the planning total, the allocate-by-current-holding-proportions split, and the combination of a multi-category security's intents into one amount |
| `apply the worksheet limits`         | §4.6             | Held quantity, turnover cap, available proceeds, funding, rounding, minimum line size, remaining cash                                                          |
| `run the cash-first pass sequence`   | §4.5             | One pass for Invest cash, a fixed two for Rebalance                                                                                                            |
| `place calculated lines in accounts` | §6               | Reductions keep their source accounts, increases auto-assign only when unambiguous, funding validated per account                                              |

46 tests on the calculator, 1957 on the crate.

### 1. Which increases absorb a funding shortfall

I raised this on #1565 and applied my reading, so here it is against real code.

§4.6 step 4 scales increases down to the funding available. My reading is that the shortfall belongs to the increases that depend on reduction proceeds, and that increases already covered by cash deployed in the first pass are left alone. Scaling those uniformly would leave selected cash undeployed whenever the turnover cap bites.

`a_shortfall_never_scales_increases_the_first_pass_already_funded` pins the behaviour: 500 of cash-funded increases and 500 expecting proceeds, with a cap that leaves only 200 of proceeds. The cash-funded line stays at 500 and the other drops to 200. If you meant the uniform version, that single test is what changes.

### 2. External cash has no account, and §6 wants funding checked per account

This one is a real gap and I do not think the code can resolve it on its own.

The worksheet has two cash inputs. Tracked cash is recorded in specific accounts. The external contribution is a single global field for cash the user says they could add, attached to no account at all.

§6 says an increase in one account cannot be funded by cash recorded in another, and that no transfer between accounts is assumed. So when the calculation puts 3000 of increases in an account holding no cash, the per-account check has to decide what the external contribution means there.

What this draft does is count external cash as available everywhere, on the grounds that it is not recorded anywhere and the user entering it is saying they would move it wherever it is needed.

The flaw is that it gets double-counted. Two accounts each needing 3000 both pass individually against a 5000 external contribution, though together they need 6000. It does not corrupt the result — the global scaling in §4.6 step 4 already caps total increases at tracked plus external plus proceeds before assignment runs, so the aggregate is never overspent — but the per-account check is weaker than it reads.

The shape that would fix it, and that I would rather agree before building: let the user allocate the external contribution per account, but only when several accounts are in scope. With a single account in scope, assign it there by default and show nothing extra. That keeps the common case as simple as it is today and only asks for the split when the split is genuinely ambiguous. It does mean a new input in the cash control, which is why I would like your view before it goes in.

Happy to take a different shape entirely if you see one.

## One implementation choice worth flagging

A reduction is drawn from the accounts that actually hold the security, split in proportion to what each holds. Under whole-unit policy every share is floored and the remainder is simply not drawn.

The alternative was to hand the last unit to one account so the parts add back up. I did not, because the account a reduction leaves decides which tax wrapper it leaves, and §3 rules that out as a non-goal. The cost is that the reduction comes out slightly smaller than the security-level line, so the remaining cash is recomputed after assignment and can go negative — a funding gap to report, never something to redistribute.

## Still to come on this branch

The orchestration service, then prefill and account allocation in the UI, then copy across all nine locales, then export. `optimizer.rs` and `rebalance_service.rs` are untouched for now and come out with the UI switch, so nothing here changes current behaviour.

---

## Checklist

### DIFF
--- a/crates/core/src/portfolio/allocation_targets/mod.rs
+++ b/crates/core/src/portfolio/allocation_targets/mod.rs
@@ -5,6 +5,7 @@ mod optimizer;
 mod rebalance_service;
 mod target_service;
 mod validation;
+mod worksheet_calculator;
 
 pub use drift_service::*;
 pub use model::*;
@@ -12,3 +13,4 @@ pub use optimizer::*;
 pub use rebalance_service::*;
 pub use target_service::*;
 pub use validation::*;
+pub use worksheet_calculator::*;
--- a/crates/core/src/portfolio/allocation_targets/model.rs
+++ b/crates/core/src/portfolio/allocation_targets/model.rs
@@ -473,6 +473,118 @@ pub struct RebalancePlan {
     pub after_bps_by_category: std::collections::HashMap<String, i32>,
 }
 
+// ── Calculated worksheet types ───────────────────────────────────────────────
+//
+// Shapes for the calculated rebalancing worksheet described in
+// `docs/features/allocations/self-directed-rebalancing-design.md`. The
+// arithmetic that produces them lives in `worksheet_calculator`.
+
+/// Which directions the calculation is allowed to produce (design §4).
+///
+/// Replaces the three-way `ScenarioMode`: cash-flow-only and sell-to-rebalance
+/// were the two ends of the same cash-first sequence, and hybrid was that
+/// sequence with both inputs present.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum WorksheetMode {
+    /// Cash increases positions and nothing is reduced.
+    #[default]
+    InvestCash,
+    /// Cash is used first, then reductions cover what is left. Offered only
+    /// when the target allows sells.
+    Rebalance,
+}
+
+/// How a category's gap is spread across the securities that carry it
+/// (design §4.2). One rule ships first, and the user selects it explicitly.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AllocationRule {
+    #[default]
+    CurrentHoldingProportions,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum WorksheetDirection {
+    Increase,
+    Reduce,
+}
+
+/// A category amount that no eligible security can carry (design §4.4).
+///
+/// Nothing is recorded, everything is excluded, or no price is usable. The
+/// amount is surfaced against its category rather than forced onto an
+/// unrelated security, and is excluded from the projected figures.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UnresolvedCategoryAmount {
+    pub category_id: String,
+    pub category_name: String,
+    pub amount: Decimal,
+    pub reason: UnresolvedReason,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum UnresolvedReason {
+    /// No security in the category is recorded in scope.
+    NoRecordedSecurity,
+    /// Every security carrying the category was left out of the eligible set.
+    NoEligibleSecurity,
+    /// Securities exist and are eligible, but none has a usable price.
+    NoUsablePrice,
+}
+
+/// Proportional scaling applied before the worksheet is prefilled (§4.6).
+///
+/// Both factors are reported so a scaled figure is never mistaken for the full
+/// amount. `None` means that limit did not bind.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AdjustmentScaling {
+    /// Step 2 — reductions scaled to fit the target's turnover cap.
+    pub reduction_factor: Option<Decimal>,
+    /// Step 4 — increases that depend on reduction proceeds, scaled to the
+    /// funding those proceeds actually raised.
+    pub increase_factor: Option<Decimal>,
+}
+
+/// One prefilled worksheet line.
+///
+/// Amounts are primary and quantities are estimates (§7).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CalculatedAdjustment {
+    pub line_id: String,
+    pub direction: WorksheetDirection,
+    pub asset_id: String,
+    pub symbol: String,
+    /// `None` when several accounts are eligible and the user has to place the
+    /// line (§6). Reductions always carry the account they are drawn from.
+    pub account_id: Option<String>,
+    /// Signed, in base currency. Negative for a reduction.
+    pub amount: Decimal,
+    pub quantity: Decimal,
+    pub unit_price: Decimal,
+    /// Below the target's minimum line size — reported, never dropped (§4.6).
+    pub is_below_minimum: bool,
+}
+
+/// The calculated adjustments the worksheet is prefilled with (§4, §5).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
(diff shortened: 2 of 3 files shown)


afadil:
Hi @Jonjon-prog, still working on this? need some help or want me to take over?
Thanks

Jonjon-prog:
Hi @afadil, yes, still on it, and sorry for the little delay, I have been busy at work.
I have just pushed the orchestration service, the second part of M1 in core: `09bf14efb` to `1e05ead5b`.

**What is in it**

External cash is now keyed by the account it would arrive in, as you suggested.
The calculator still spends the total, and only the account funding check tells the accounts apart, so two accounts can no longer both pass against the same contribution.
A single account scope has one entry and asks the user for nothing extra.

`AllocationWorksheetService` has two methods.
`generate_adjustments` resolves the drift report, taxonomy contributions, holdings, quotes, FX and constraints, then runs the calculator.
`calculate_worksheet` is the preview ported from the worksheet branch: it validates the lines as the user has them, projects them, and never calls the calculator, so an edit cannot regenerate anything.

Eligibility now gates increases in both modes, and an empty selection gives unresolved category amounts instead of the validation error `main` raises in cash flow only.
Blocking constraints from #1177 are applied before the calculation, since the preview rejects any line they cover; `Avoid` still only warns in the preview.
Both paths share one implementation of constraint matching.
Prices in the prefill come from quotes and FX like the preview, so a prefilled line is not restated as soon as it is checked.

51 tests on the calculator, 23 on the service, 1994 on the crate.
Nothing is wired to Tauri or the server yet, and `optimizer.rs` and `rebalance_service.rs` are untouched.

**Three calls I made on my own**

1. `planning_total`, which §4.2 uses without defining.
With a cash sleeve it is the drift total plus external cash, since tracked cash is already inside it; without one it also adds the tracked cash selected, which is what the old optimizer did.
The prefill and the preview use the same function, so the projection cannot contradict the prefill that produced it, and the turnover cap is measured against it too.

2. Spending past the available funding is a warning in the preview, not an error.
The reference branch refused the calculation, but §5 says an edit that breaks a limit is reported and left alone, so the projected cash sleeve may also go negative.
Same for an account whose increases exceed what it can fund.

3. The account funding check counts the cash recorded in each account, not a share of the tracked cash selected.
The global funding scale already caps the total, so this check only answers whether the cash sits in the right account, and prorating would raise false shortfalls on valid worksheets.

**One open question**

The preview accepts at most 50 lines, a limit that comes from the reference branch.
The prefill has no such bound, so a portfolio with more than 50 eligible securities could produce a worksheet its own preview rejects.
Dropping lines would mean choosing between securities, so I left it: would you rather raise the cap, remove it, or keep it and handle it in the UI?

**What is left**

Prefill and account allocation in the UI, then copy across the nine locales, then export.
Happy to keep carrying it.
If you want to take a piece, the locales are the most separable one.

afadil:
Thanks @Jonjon-prog, few points:

- **Planning total:** Agree with your proposed definition. Add tests for cash left undeployed and taxonomies without a cash category.
- **Funding shortfalls:** Keep the edits and preview visible. Show “Funding needed”; acknowledging the warning should not make the result appear funded.
- **Account funding:** Agree with checking each account’s available cash rather than prorating the selected amount. But use the same eligibility rules globally and per account. Cash excluded globally must also be excluded from the account check.
- **Line limit:** Remove the old 50-line cap and allow zero adjustments. Generation and preview should handle the same states, including additional rows from account splits. Don’t drop adjustments to fit a limit.
- **CI:** The Rust check currently fails because the test fixture is missing `exchange_mic` and `instrument_type`.
- **Tests:** Please add generation-to-preview tests for these cases. The current service tests cover helpers rather than the complete flow.

The separation between generation and manual editing looks right. Please keep that when connecting the UI.

Jonjon-prog:
Hi @afadil, the UI is wired up and the worksheet now works end to end: choosing which accounts a worksheet may change, cash not yet recorded entered per account, per-account funding shown where the change is placed, and eligible securities listed with the accounts that hold them.
Everything from `8995ede3d` to `02e5e7659` is on the branch.

I am still working through how the screen itself should be organised — it carries a lot at once and that is not settled, so expect the layout to move again.

Five calls I made on my own, all of them from running it against a real portfolio.

**1. Cash to deploy is bounded by the accounts the worksheet changes.**
The control took its ceiling from the drift report, which spans the whole scope and counts cash accounts.
The core counts neither, so the worksheet could offer 2912.66 of cash and then refuse the generation it had just invited, with `Tracked cash selected (2912.66) exceeds observed deployable cash (67.60)`.
The ceiling now comes from the core's own figure for the selected accounts, which is the same rule you asked for on the account funding check.
I also stopped refusing an amount above it: the worksheet deploys the cash that exists, and the preview reports the difference as a `CashUnavailable` warning.
This is the one I would most like your read on, because §5 says an edit that breaks a limit is reported rather than corrected, and this corrects.
My reasoning is that the selected cash is not a limit being crossed but an input that is wrong: it feeds `planning_total`, so accepting cash that does not exist would not produce a visible shortfall, it would quietly resize every category target.
Cash that is not recorded yet already has a home in the contribution input.

**2. A line short of one whole unit is reported, not refused.**
`resolved_quantity_and_amount` returned an error, so a single unaffordable line failed the whole preview and took the rest of the worksheet with it.
That sat badly with "allow zero adjustments" and with §4.6.6, where a line below the minimum is flagged rather than dropped.
It now resolves to nothing and says so, through a `BelowOneUnit` warning.

**3. An increase is placed in the only eligible account that already records the security.**
§6 auto-assigns only when exactly one account is eligible, and I have extended that to when exactly one eligible account already holds the security.
The argument is that this is the same fact §6 already draws reductions from — where a security sits is a fact rather than an assignment — and splitting every increase by hand was the slowest part of the screen for changes that were never a choice to begin with.
A second holder hands the decision straight back, and any amount typed by hand takes over immediately.
I rewrote `an_increase_with_several_eligible_accounts_stays_unallocated` to use two holders, since its security sat in one account and the test was passing for the wrong reason.
Say the word and I will put it back to the letter of §6.

**4. Prices carry 32-bit float residue, and the worksheet was acting on it.**
Yahoo sends floats, so 39.85 is stored as 39.84999847412109375 and 9.173 as 9.173000335693359375.
An amount meant to buy one unit then divides out a fraction of a billionth short of it, the whole-unit floor turns that into zero, and the user is left correcting decimals by hand over a difference no market recognises.
Quantities are now rounded to six decimals before flooring, well below any fraction anyone trades.
The residue itself belongs to the market data layer: a 32-bit float carries about seven significant digits, so rounding to seven at ingestion would restore 39.85 exactly without harming an asset priced at 0.00000123.
That is every price in the app rather than this PR, so I left it alone — happy to open it separately if you want it fixed.

**5. One thing I tried and reverted.**
I moved entry to units for whole-share targets, then took it back out.
It did remove the rounding, but it put a second unit on a screen denominated in currency, and §4.6.5 keeps amounts primary for the same reason.
It is in the history as `bdec13f69` and `9c74caf11`; the fixes underneath it stayed.

Fixed along the way: the worksheet header was asking for a copy key that was never written, and the calculated summary — the only place the leftover cash is explained — rendered in the position view alone, so anyone reviewing the result saw funding left over with nothing accounting for it.

Still to come: copy across the ten locales, then export.

PR #1635, opened 2026-09-04
 feat(spending): exclude categories from spending

### LINKED ISSUE #1559: Need ability to filter categories from spending analysis
We need the ability to filter categories out of the spending analysis. This is a very common use case as sometimes some categories are not true spending (eg, reimbursable transactions) or some large transactions skew the data or swamp meaningful insights (eg, large tax payments for a previous year that wouldn't map to this year).

kryofin:
Fixes #1559

Lets a user mark spending categories as **excluded from spending**. Excluded categories (and their subcategories) drop out of every spending surface — headline totals, category breakdowns, daily/monthly series, pace, budgets, event summaries, and the row-derived charts — while their transactions stay visible in the ledgers. Typical use: taxes, especially for the previous year, reimbursable expenses, a "Transfers to partner" bucket, or a one-off that would otherwise distort the month.

## What changes

**Setting** — a new `spending.excluded_category_ids` settings key (JSON array of category ids, sorted/deduped for stable sync payloads). Ids are resolved to the effective set (raw ids + descendants) by a small `ExclusionIndex`; raw ids are kept even if the category no longer exists so stale assignments still filter.

**Backend filtering** — every aggregation surface derives included/excluded portions through one allocator helper (`split_spending_allocations`), so the reconciliation invariants keep holding:

- headline spent == Σ per-category rows + uncategorized
- Σ by_day.spent == Σ by_month.spent == headline.spent

Covered: insight (headline, by_day, by_day_by_category, by_month, pace, category rows, budget fanout), monthly report (summarize, daily series, breakdown), event summaries, and budget actuals (filtered *before* the top-category roll-up so an excluded subcategory can't leak into an included parent; excluded top-level rows leave the budget snapshot while their targets persist). Uncategorized spend is never excluded, and a fully-excluded activity is not re-bucketed as uncategorized. Income and savings paths are untouched.

**Rows carry the answer** — `CashActivity` gains `visible_spending_amount`, computed once in `CashActivityService` with the same classifier, allocator, and `ExclusionIndex` as the reports (the taxonomy is only read when something is excluded). The frontend's row-derived visuals (rhythm heatmap, day-of-week chart, when-you-spend grid, events timeline/card, baseline pace, event chart data) and the heatmap-cell drill-in sheet read that field, so they agree with the server headline without re-deriving allocation precedence or descendant exclusion in TypeScript. Rows from an older backend fall back to the unfiltered amount. The category drill-down drawer (server-driven since #1667) needs no row math: its stats come from `by_day_by_category`, which is already exclusion-filtered, and it keeps excluded subtrees out of its transaction query so the list matches the numbers.

**UI** — the category edit dialog (spending taxonomy, edit mode) gets an "Exclude from spending" switch. Excluded rows render dimmed with a crossed-eye badge; subcategories inherit a parent's exclusion ("Excluded with {parent}", own switch disabled) — the tree passes inherited state to the dialog, so no ancestor walk lives in the frontend. Saving sequences the settings write before the category write and reports once; if either fails the dialog stays open with its values and a retry skips the half that already landed. The dashboard eyebrow and the where-stage breakdown footer show an "N categories excluded" link to the settings page.

**Sync** — the new key is gated in the app-settings outbox, the inbound allow-list, and the bootstrap row filter; an older peer logs and skips it.

## Commits

1. `feat(spending): add excluded-category settings key and exclusion index`
2. `feat(spending): filter excluded categories from the spending insight`
3. `feat(spending): filter excluded categories from reports, events, and budget`
4. `feat(spending): exclude-from-spending toggle in the category edit dialog`
5. `feat(spending): excluded-category hints and server-filtered row amounts`

## Tests

- Rust: `ExclusionIndex` (descendants, stale ids, cycle guard), insight reconciliation with a mixed set (excluded parent, included, split with one excluded line, uncategorized), analytics breakdown/summarize exclusion, budget roll-up guard, `CashActivityService.visible_spending_amount` (excluded-via-parent, partial split, uncategorized, income row, no-exclusions passthrough).
- Frontend: `getVisibleSpendingAmount` field/fallback, events card, categories settings page.

## Notes for review

- Exclusion is stored on settings rather than as a column on the category. That keeps the taxonomy model and its sync bundle untouched, at the cost of a two-write save (sequenced, see above) and no pruning of a deleted category's id from the list (harmless: the index ignores unknown ids and the UI counts only live ones). A category-level flag would be the deeper design if exclusion grows more semantics.
- `spending_summary` (multi-period rollup) is not wired to any command or route and was left unfiltered.

## Checklist


afadil:
@kryofin thanks. Added a small fix for the heatmap weekly subtotals to exclude filtered categories

PR #1719, opened 2026-09-13
fix(quotes): stop a degraded provider response from wiping quote history

raphaelcoeffic:
## Problem

`BackfillHistory` purges every non-MANUAL quote for an asset — across **all** sources — before writing the provider response. The only gate was that the response was non-empty:

```rust
if quotes_count > 0 {
    if plan.purge_provider_quotes { delete_provider_quotes_for_asset(&asset_id).await }
    upsert_quotes(&quotes).await
}
```

One bar is enough to delete months of history and replace it with itself. A provider that degrades at the wrong moment truncates an asset to whatever few days it happened to return, and because the purge is not scoped to the winning provider it takes the other source's history with it.

This is reachable from every "rebuild history" entry point: Settings → Market Data → Rebuild History, the command palette, and the per-asset *Refetch price history* / *Refresh Quotes* / *Refetch Prices* actions.

#1718 fixes one concrete way a response ends up short. This guards the class.

## What is *not* a bug

Dropping the previous provider's rows is the **point** of rebuilding after switching providers — if you have chosen a new provider you do not want the old one's data lingering, even if the new one has less history. An earlier version of this patch blocked that, which was wrong.

So the guard distinguishes a deliberate choice from a degraded response, via two conditions:

- a provider returning **less than it already gave us** for this asset — degradation, whoever it is;
- a provider **the user did not choose** shrinking overall coverage — a failover about to delete a deeper source's history.

`response_is_preferred_provider` is read from `asset.preferred_provider()`, the user's explicit choice, rather than `quote_sync_state.data_source`, which lags behind a switch because it is written *after* a successful sync.

When the purge is refused the response is still merged, so fresh prices land and only the destructive replace is skipped. A failed bounds lookup also skips the purge — an unreadable store must not licence a destructive replace.

## Note on the diff size

+543 lines, but only ~40 are production code. The rest is an in-memory `QuoteStore` for the tests: the purge is the only operation in quote sync that deletes stored data and it had **no coverage at all**. The mock mirrors the two SQL behaviours that matter — the unique index on `(asset_id, day, source)` and `source != 'MANUAL'` in the delete.

`apply_fetched_quotes` is extracted so the persistence step can be driven without mocking `ActivityRepositoryTrait`'s 36 methods.

## Tests

| test | asserts |
|---|---|
| `backfill_keeps_history_when_the_response_is_short` | 6-bar response does not replace 200 days; prices still merge |
| `backfill_keeps_history_when_an_unchosen_provider_wins_short` | failover cannot delete the preferred provider's deeper history |
| `backfill_purges_the_previous_provider_on_a_deliberate_switch` | a chosen provider **does** drop the old source, even if shallower |
| `backfill_still_purges_when_the_response_covers_the_history` | healthy rebuild still clears superseded rows; MANUAL preserved |
| `incremental_sync_without_purge_preserves_history` | control — no purge requested |
| `purge_is_refused_only_when_it_would_destroy_history` | the decision function's cases |

The third and fourth are the ones showing this guards the bug rather than disabling the feature. Neutering `purge_would_destroy_history` fails exactly the three that should fail.

`cargo test -p wealthfolio-core`: 2038 pass. clippy and fmt clean.

## Related

- #1718 — the null-deserialisation bug that produced the short responses in practice. Worth landing first; this becomes defence-in-depth once it is in.
- Still open: `fetch_quotes_with_diagnostics` accepts the first provider returning ≥1 quote with no completeness check, and a provider error is only logged once every provider has failed — so a preferred provider losing is invisible.

## CLA

 - [x] I confirm I have read and agree to [CLA.md](https://github.com/wealthfolio/wealthfolio/CLA.md).

### DIFF
--- a/crates/core/src/quotes/sync.rs
+++ b/crates/core/src/quotes/sync.rs
@@ -303,6 +303,111 @@ fn should_allow_inactive_asset(
             ))
 }
 
+fn response_starts_later(
+    existing_earliest: Option<NaiveDate>,
+    response_earliest: Option<NaiveDate>,
+) -> bool {
+    match (existing_earliest, response_earliest) {
+        (Some(existing), Some(response)) => response > existing,
+        // An empty response would delete everything and put nothing back.
+        (Some(_), None) => true,
+        // Nothing stored for this comparison, so there is no coverage to lose.
+        (None, _) => false,
+    }
+}
+
+/// Whether the backfill purge would destroy history rather than replace it.
+///
+/// Deleting the previous provider's rows is the *point* of a rebuild after
+/// switching providers, so a shallower history is an acceptable outcome of a
+/// deliberate choice. What must not happen is a transiently degraded response
+/// silently replacing good data. The two are told apart by:
+///
+/// - a provider returning less than it already gave us for this asset, which is
+///   degradation whoever it is; and
+/// - a provider the user did not choose shrinking overall coverage, which means
+///   a failover deleting a deeper source's history.
+fn purge_would_destroy_history(
+    same_source_earliest: Option<NaiveDate>,
+    any_source_earliest: Option<NaiveDate>,
+    response_earliest: Option<NaiveDate>,
+    response_is_preferred_provider: bool,
+) -> bool {
+    if response_starts_later(same_source_earliest, response_earliest) {
+        return true;
+    }
+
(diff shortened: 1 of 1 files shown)


afadil:
Superseded by #1761, now merged. Refresh preserves existing history, while deliberate replacement is available through the separate confirmed **Reset provider history** action

PR #1720, opened 2026-09-13
fix(valuation): refresh the latest saved valuation on incremental recalc

raphaelcoeffic:
Incremental valuation recalculation discarded the latest saved day as its cash-flow anchor. Once today had a saved valuation, subsequent refreshes wrote no rows; resuming the following day also left the latest saved day at its previous price.

Reuse `since_date_calculation_window(last_saved)` so the preceding day is the discarded anchor and the latest saved day through today is persisted. Keep the existing save path, which already upserts existing dates. This preserves cash-flow calculation without adding a five-day overlap or changing persistence behavior.

Scope: refresh the latest saved valuation and extend history. Quote syncing is unchanged; corrections to dates older than the latest saved valuation remain outside this fix, including delayed closes when the user's local date has already advanced.

Validation:
- Two service-level regression tests use the existing in-memory fixtures, change quotes, invoke `IncrementalFromLast`, and inspect saved rows. They cover repeated same-day refreshes, next-day resume, preservation of a deposit on the first rewritten day, no duplicate dates, and unchanged earlier rows.
- Both regression tests fail with the original window selection and pass with the fix.
- `cargo test --locked -p wealthfolio-core`: 2,035 passed; one doc-test ignored.
- `cargo fmt --all -- --check` and core Clippy with `--all-targets --all-features -- -D warnings`: passed.

- `cargo check --locked -p wealthfolio-app -p wealthfolio-server`: passed, using the CI-style placeholder frontend assets and `CONNECT_API_URL=http://test.local`.

## CLA

- [x] I confirm I have read and agree to [CLA.md](https://github.com/wealthfolio/wealthfolio/CLA.md).

### DIFF
--- a/crates/core/src/portfolio/valuation/valuation_service.rs
+++ b/crates/core/src/portfolio/valuation/valuation_service.rs
@@ -25,7 +25,7 @@ use crate::portfolio::valuation::valuation_model::{
     DailyAccountValuation, ExternalFlowSource, NegativeBalanceInfo, ValuationStatus,
 };
 use crate::portfolio::valuation::ValuationRepositoryTrait;
-use crate::quotes::{Quote, QuoteServiceTrait};
+use crate::quotes::{Quote, QuoteServiceTrait, OVERLAP_DAYS};
 use crate::utils::time_utils;
 use async_trait::async_trait;
 use chrono::{DateTime, Duration, NaiveDate, Utc};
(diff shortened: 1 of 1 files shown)


afadil:
Thanks for the fix! I simplified the anchor handling a bit and added regression tests

PR #1745, opened 2026-09-15
fix(accounts): clean up account data and configuration on deletion

afadil:
## Description

Deleting an account leaves holdings snapshots, daily valuations, and configuration references behind. Remove these within the account deletion transaction, including accepted account deletions from device sync.

- Keep existing foreign-key cascades for normal account-owned rows. Explicitly delete snapshots and valuations in the existing storage helper, including when an accepted remote deletion has no account row to remove. Snapshot positions cascade from their snapshots.
- Use one storage helper for logical account deletions to remove import-template links, account allocation targets, and account references in shared allocation constraints. Remove the exact account ID from contribution-limit CSV lists and spending-account JSON lists while retaining surrounding configuration and reusable templates.
- Run the helper only for logical account deletions. Bulk snapshot replacement bypasses it, preserving local broker snapshots and configuration tables omitted by older snapshots. Direct SQL account deletion does not perform this extra cleanup. Cleanup follows the account deletion without additional sync events; handling later independent events that reintroduce references is outside this change.
- Repair existing orphan calculated snapshots and valuations. Repair source snapshots, import-template links, and allocation rows only when the account is absent and has a recorded deletion tombstone, preserving data that may be waiting for its account to sync. Migration cleanup uses direct deletes and explicitly removes dependent rows while foreign keys are disabled. Historical spending and contribution-limit lists are left unchanged; the runtime helper cleans them on future deletions.

Remove the manually maintained trigger allowlist from portable-backup validation. Selected backups are treated as trusted inputs; imported triggers may run during staged migrations. Integrity checks, final schema comparison, private staging, and pre-restore backups remain. Document this trust policy and pin the backup-test fixture to its two intended migrations.

One data-only cleanup migration; no new account-deletion trigger, table rebuild, or vacuum. Supersedes #839 without additional AccountService repository dependencies or best-effort cleanup calls.

## Validation

- Regression coverage includes cleanup failure rollback, repeated deletion, local and remote deletion (including an absent account), snapshot replacement preserving broker history and omitted configuration, tombstone-gated historical repair, and runtime list cleanup while the migration preserves historical lists.
- `cargo test -p wealthfolio-storage-sqlite --lib --offline`: 368 passed.
- `cargo test --locked -p wealthfolio-server --test database_secret_file --offline`: 3 passed.
- `cargo check -p wealthfolio-server -p wealthfolio-app --offline`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- Independent agent reviewed the implementation and found no actionable issues.


afadil:
@codex review

Issue #1749, opened 2026-09-15
feat(addon-sdk): could budgets, spending accounts, and activity categories be exposed


Czmr-108:
Hi! Thanks for Wealthfolio — I've been using it to track my own finances for a while now.

I ended up building a small addon for myself. Wealthfolio keeps track of what I've already spent, but I kept wanting to know what's coming: my rent, my paychecks, my subscriptions, the money I put aside each month. So the addon holds all my upcoming stuff and projects it forward — how much I'll have left at the end of the month, whether I'll run short before payday, roughly where my net worth lands in a few years. When something actually happens I post it into Wealthfolio so the app still has the real history.

It's just for me, nothing fancy, but it's become how I plan my month.

The thing is, there are a few settings I've already filled in inside Wealthfolio that my addon can't see, so I've had to set them up a second time in the addon — and then they slowly go out of sync whenever I change one and forget the other. I went looking through the SDK to see if I'd missed something, and I don't think they're exposed. Would any of these be possible?

**1. My budget groups.** Wealthfolio has budgets — I can see `get_budget`, `create_budget_group`, `assign_category_to_group` and the `BudgetGroup` / `BudgetTarget` types in the app — but there's nothing budget-related in `packages/addon-sdk` at all. So my addon keeps its own copy of Needs / Wants / Savings and its own category assignments, which is the same setup twice. If an addon could just read the budget, the groups could come from Wealthfolio the way my categories now do.

**2. Writing the amounts back.** Less important, but: my addon works out what each group comes to from my upcoming bills, which is usually a decent starting point for the next month's targets. At the moment I've got a little copy button next to each number so I can paste it into Wealthfolio by hand. Being able to write them with `upsert_budget_target` would save me doing that every month.

**3. Which accounts count as spending accounts.** There's a Spending accounts screen in Settings where I've already picked this, but addons can't see it — `Account` doesn't have a field for it, and `accountType` doesn't help because my everyday account, my emergency fund and my small business account are all `CASH`. They look identical from the addon's side. I need to tell them apart to work out what I can actually spend before payday, so right now I'm re-tagging them inside the addon.

[...]

PR #1753, opened 2026-09-17
feat(addon-sdk): expose spending reads

PAKalucki:
## Description

Expose two existing, read-only Spending capabilities to addons:

- `ctx.api.spending.getReport(request)` returns aggregate spending, income, and
  saving totals, category breakdowns, and daily series.
- `ctx.api.spending.searchCashActivities(request)` returns paginated cash
  activities enriched with cash-flow buckets, category assignments, splits,
  and effective spending amounts.

This lets addons analyze Wealthfolio's existing Spending classification rather
than maintaining a second category model. Personal inflation and budgeting
addons are examples of consumers that need this data. Related to #1749.

The bridge reuses the existing `get_spending_report` and
`search_cash_activities` commands. It adds no backend endpoint, schema, or
storage changes.

### Permission model

- `getReport` uses the existing medium-risk `spending` permission.
- Transaction-level `searchCashActivities` uses the existing high-risk
  `activities` category as `activities.searchCashActivities`.

Runtime guards and static addon permission detection enforce the split. This
keeps rule-only addons from gaining transaction-history access through their
existing Spending permission.

### Included

- Public SDK request, response, activity-assignment, split, and report types.
- Host bridge wiring for both desktop and web runtimes.
- Iframe RPC allowlisting and user-facing permission labels.
- Static permission detection and regression coverage.
- API reference and unreleased changelog entries.

### Testing

- `pnpm --filter @wealthfolio/addon-sdk type-check`
- `pnpm --filter @wealthfolio/addon-sdk lint`
- `pnpm --filter @wealthfolio/addon-sdk build`
- `pnpm build:types`
- `pnpm --filter frontend exec vitest run src/addons/type-bridge.test.ts src/addons/iframe/addon-iframe-manager.test.ts src/adapters/adapter-command-parity.test.ts` — 80 tests passed
- `pnpm --filter frontend build`
- Targeted Prettier and frontend ESLint checks passed (one pre-existing `require-await` warning remains in `addons-runtime-context.ts`).

## Checklist


Czmr-108:
This is great — thank you for picking this up so quickly, I really appreciate it.

`searchCashActivities` returning the category assignment solves the part that was bothering me most: anything I add directly in Wealthfolio currently lands in my addon as a single "Uncategorized" row, because the amount and account come through but the category doesn't. Being able to read the app's own classification instead of keeping a second one is exactly what I was hoping for.

The cash-flow buckets and effective spending amounts look useful too — I've been working those out myself from activity types, so having the host's own view of what counts as spending versus a transfer should be more reliable than my guess.

The permission split makes sense to me as well. Transaction-level history is a bigger ask than reading spending rules, and keeping `searchCashActivities` under `activities` rather than letting it come along with the `spending` permission seems right.

Happy to test it against my addon once it lands if that's useful — just let me know.

Thanks again.

afadil:
Thanks for this PR. Pushed a small follow-up: reports now include `baseCurrency`, and I added tests for permission denials and iframe RPC calls.

